PR #31, opened 2015-03-15
[-] MO: 1-click upgrade fails miserably on sockets

dkarvounaris:
1-click upgrade will cause a mysql error and will fail miserably to even start any upgrades, if something else than host:port is being used in settings.inc.php. However, other strings are also valid connection strings, as example: localhost:/var/run/mysql/mysqld.sock. This fix supports them.


gRoussac:
Hi,

Thank you for your PR. What is strange here is that it should not work miserably neither in the solution.

https://github.com/PrestaShop/PrestaShop/blob/1.6/classes/db/DbMySQLi.php

Can you please add line feeds on if statements and replacing the spaces by a tabulation on line 37.

Thank you in advance. Regards

dkarvounaris:
Sorry, that happens I guess when copy-pasting from the console instead of using the IDE and being up all night :)

What do you mean it should not work neither? Not sure I understand.

gRoussac:
Hi,

Don't be, that's just a detail. How can you use localhost:/var/run/mysql/mysqld.sock with the core DbMySQLi.php if it is the same code ?

Regards

dkarvounaris:
Ok this is the last time I'm using Github for editing... what a mess! :D

dkarvounaris:
Well, localhost:/path/to/socket has been working with the core perfectly. Just "/path/to/socket" isn't however. I think the core used DbPDO though and not DbMysqli on the server I had that issue.

gRoussac:
Hi,

No don't close it, i'll try to check that and core too.

Rgerads

dkarvounaris:
I wasn't aware I did close that? Oh, maybe it was the one syncing I did that delete and recreated the patch done it automatically. Don't worry, I will resubmit, I was checking anyway the core and wanted to change both.

PR #42, opened 2015-10-05
[-] MO : Fix for "You already have the X version." + #PSCSX-6286 Class 'MCached' not found in classes/cache/Cache.php

gRoussac:


gRoussac:
thank you !

PR #46, opened 2015-11-12
[*] MO : Warn when an opcache configuration can cause upgrade failures

dkarvounaris:
See http://forge.prestashop.com/browse/PSCSX-7272


Quetzacoalt91:
It seems this issue it still valid today, I got an issue (#82) which disappeared as soon as I disabled Opcache.

I can't merge this PR because of the differences, but I agree this requires something to be done on the check list or directly during the process.

Keeping this PR opened until a solution is found.

Quetzacoalt91:
As said in https://github.com/PrestaShop/autoupgrade/pull/82, the issue of loaded classes seems to be fixed with the future release of the autoupgrade v4.8.0.

Except if someone else encounter this issue even with the latest version, I think it's safe to close this one.

Regards

PR #54, opened 2017-03-20
Fix for non-linux system

aleeks:
ping @maximebiloe


maximebiloe:
Muchas gracias @aleeks

PR #57, opened 2017-03-27
Fix bug when using php7.1

SebBareyre:
MO: fix this bug when using php7.1 :
PHP Fatal error:  Uncaught Error: [] operator not supported for strings in /modules/autoupgrade/db/DbQuery.php:70


maximebiloe:
Thank you @SebBareyre

PR #64, opened 2017-04-07
Fix JS trans escape

maximebiloe:


aleeks:
Thank you @maximebiloe

PR #65, opened 2017-04-28
Do not suggest automatic rollback after 1.7 upgrade

maximebiloe:


aleeks:
Thank you @maximebiloe

PR #73, opened 2018-01-08
Code cleanup & refacto

Quetzacoalt91:
The objective of this PR is to clean the deprecated code, and make the module easier to maintain. 
However, the existing interface & features remain the same, don't expect to see changes during the upgrade process.

### Information
* Compatibility range: PS 1.6 & 1.7

## Documentation
https://docs.google.com/document/d/1DQMP6bg5_fgbJQScdidCQCEmAi4MzRgYNhdGVcYOqm4/edit?usp=sharing

### General
- [x] Clean deprecated code for backward compatibility with unsupported versions
- [x] Add composer and its autoloader
- [x] Tests
- [x] Add entrypoint for whole upgrade process

### Templating
- [x] 1 - Export all content to display outside the controller
- [x] 2 - Use another templating system for the exported content (Twig)

### Classes (features)
- [ ] Release entity
- [ ] Release manager (download, comparison ...), currently `Upgrader.php`
- [x] Upgrade channel manager
- [x] Logger (existing one seems to be unused)
- [x] Module configuration
- [x] JSON Response generator
- [x] Destination version of PS checker (allowing the module to use the proper adapter)
- [x] Error handler
- [ ] Guzzle for API addons

### Improvements
- [ ] Check all server requirements are ok before starting upgrades
- [x] In case of error, output all log waiting to be displayed
- [ ] Install new modules

### Upgrades
* [x] From 1.7 to 1.7
* [x] From 1.6 to 1.7
* [x] From 1.6 to 1.6


kpodemski:
Closer and closer... let me know when it would be ready to do some tests @Quetzacoalt91 :)

Quetzacoalt91:
You may go for it! The latest commits are precisely fixing some bugs detected during the tests. :)

kpodemski:
Cool! I'll try to run some tests on few stores today, starting from default 1.6 version, i'll let you know about results

kpodemski:
Hey @Quetzacoalt91 

https://github.com/Quetzacoalt91/autoupgrade/blob/87a227b1fd950ceaf1dc8ad8e3879884b14c35c1/classes/Parameters/UpgradeConfiguration.php#L93

you should have -1 here, there's no 3 in array and i get:
modules\autoupgrade\classes\Parameters\UpgradeConfiguration.php line 61 - Undefined offset: 3

kpodemski:
Even after this change the process went completely wrong, in few 30 seconds everything stopped because ajax returned maintenance page on response, what is much weirder i cannot access back-office right now because it was erased by upgrader for some reason

Result: entire store broken

I was testing 1.6.1.18, clean installation.

jocel1:
@kpodemski Sounds like you should do the same change for 'loopFiles', 'loopTime', 'maxBackupFileSize' and 'maxWrittenAllowed'

jocel1:
Ha yes, I though you were suggesting to do the -1 in each getter which was using getPerformanceLevel :)

Quetzacoalt91:
Thanks for your report @kpodemski, I'll review this.
I'll reproduce your upgrade. You said you starred a fresh shop on 1.6.18 and upgraded it on 1.7?
(Weird because this what we are supposed to test on Travis)

kpodemski:
@Quetzacoalt91 yeah, to be honest at first i wanted to upgrade "real" store but then i thought that i should have start from the easiest scenario possible, clean installation, right after installation went to the upgrade process... 

first try, set performance to "High", first error, solved, trying again, maintanance page because of deletion my /back-office/ administration directory

Quetzacoalt91:
I tried and it worked, there is probably another test case to find.

I wonder if there is an issue when then admin folder is different from `admin` and `admin-dev`?

kpodemski:
@Quetzacoalt91 yeah well, it might be, but admin folder name is always randomized, in my case that was /back-office/ but sometimes is /123123sadasd/ etc.

I can try with admin-dev

kpodemski:
@Quetzacoalt91 i'm happy to write that everything went perfectly using /admin-dev/ as a name of admin folder, next step for me is to test it on some bigger project instead of clean installation

i'll keep you posted :)

kpodemski:
One suggestion:
- update translations after the upgrade, most of the back-office was completely outdated when it comes to translations, of course i did that manually but it would be cool to have it as an option or by default

Quetzacoalt91:
The previous version is already supposed to update the languages, as exported in the new CoreUpgrader classes:

* https://github.com/Quetzacoalt91/autoupgrade/blob/8253c85e7f5e67bd19277c914fc39381c2a0df7e/classes/UpgradeTools/CoreUpgrader/CoreUpgrader.php#L470-L495
* https://github.com/Quetzacoalt91/autoupgrade/blob/8253c85e7f5e67bd19277c914fc39381c2a0df7e/classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php#L65-L92

On my side, the menu entries weren't translated, but the other parts of the back-office seemed fine.

.gitignore
```diff
@@ -0,0 +1,4 @@
+nbproject
+
+/tests/E2E/
+/vendor/
```
PierreRambaud: why `nbproject`? Should we add something like https://github.com/github/gitignore/blob/master/Prestashop.gitignore? or at least:

- .project
- .idea
- build
- coverage
- .env

.travis.yml
```diff
@@ -0,0 +1,33 @@
+language: php
+services: docker
+
+php:
+  - '5.6'
```
PierreRambaud: Only 5.6?

ajax-upgradetabconfig.php
```diff
@@ -0,0 +1,91 @@
...
+ *  @copyright  2007-2018 PrestaShop SA
+ *  @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *  International Registered Trademark & Property of PrestaShop SA
+ */
+
+use PrestaShop\Module\AutoUpgrade\Tools14;
+
+if (function_exists('date_default_timezone_set'))
+{
+	// date_default_timezone_get calls date_default_timezone_set, which can provide warning
```
PierreRambaud: Remove tabs, use spaces please :)

ajax-upgradetabconfig.php
```diff
@@ -0,0 +1,91 @@
...
+ * @param string $callerFilePath Path to the caller file. Needed as the two files are not in the same folder
+ */
+function autoupgrade_init_container($callerFilePath)
+{
+    if (php_sapi_name() === 'cli') {
+        $_POST['dir'] = getopt('', array('dir:'))['dir'];
+    }
+
+    define('_PS_ADMIN_DIR_', realpath($callerFilePath.'/../'));
+//    require_once(realpath($callerFilePath.'/../../config/config.inc.php'));
```
PierreRambaud: useless comment

ajax-upgradetabconfig.php
```diff
@@ -0,0 +1,91 @@
...
+
+    if (!defined('_PS_MODULE_DIR_'))
```
PierreRambaud: use `{}` please

ajax-upgradetab.php
```diff
@@ -24,85 +24,30 @@
...
+ */
+
+require_once(realpath(dirname(__FILE__).'/../../modules/autoupgrade').'/ajax-upgradetabconfig.php');
+$container = autoupgrade_init_container(dirname(__FILE__));
+
+(new \PrestaShop\Module\AutoUpgrade\ErrorHandler($container->getLogger()))->enable();
+
+if (!$container->getCookie()->check($_COOKIE)) {
+    // If this is an XSS attempt, then we should only display a simple, secure page
+    if (ob_get_level() && ob_get_length() > 0)
```
PierreRambaud: use `{}` please

ajax-upgradetabconfig.php
```diff
@@ -0,0 +1,91 @@
...
+        define('_PS_MODULE_DIR_', realpath($callerFilePath.'/../../').'/modules/');
+
+    define('AUTOUPGRADE_MODULE_DIR', _PS_MODULE_DIR_.'autoupgrade/');
```
PierreRambaud: You test for _PS_MODULE_DIR_ but not for AUTOUPGRADE_MODULE_DIR, any reason?

ajax-upgradetabconfig.php
```diff
@@ -0,0 +1,91 @@
...
+
+    // the following test confirm the directory exists
+    if (!isset($_POST['dir'])) {
+        echo 'no directory';
+        exit(1);
+    }
+
+    // defines.inc.php can not exists (1.3.0.1 for example)
+    // but we need _PS_ROOT_DIR_
+    if (!defined('_PS_ROOT_DIR_'))
```
PierreRambaud: use {} please

ajax-upgradetabconfig.php
```diff
@@ -0,0 +1,91 @@
...
+//        eval('class Tools extends Tools14{}');
+
+    $dir = Tools14::safeOutput(Tools14::getValue('dir'));
+
+    if (realpath($callerFilePath.'/../../').DIRECTORY_SEPARATOR.$dir!== realpath(realpath($callerFilePath.'/../../').DIRECTORY_SEPARATOR.$dir)) {
+        echo 'wrong directory :'.(isset($_POST['dir']) ? $dir : '');
+        exit(1);
+    }
+
+//    if (!defined('_MYSQL_ENGINE_'))
```
PierreRambaud: Remove useless comments or add comment to define why they are useful please

autoupgrade.php
```diff
@@ -24,6 +24,8 @@
 *  International Registered Trademark & Property of PrestaShop SA
 */
 
+require __DIR__.'/vendor/autoload.php';
```
PierreRambaud: is it really useful?

classes/AjaxResponse.php
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * @var array Params to send (upgrade conf, details on the work to do ...)
+     */
+    private $nextParams = array();
+
+    /**
+     * Request format of the data to return.
+     * Seems to be never modified. Converted as const.
+     */
+    const nextResponseType = 'json';
```
PierreRambaud: const in uppercase please

classes/BackupFinder.php
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @return array
+     */
+    private function getBackupFilesAvailable($backupPath)
+    {
+        $array = array();
+        $files = scandir($backupPath);
+
+        foreach ($files as $file) {
+            if ($file[0] != '.' && substr($file, 0, 16) == 'auto-backupfiles') {
```
PierreRambaud: No need to check for `$file[0]` if you use substr just after

classes/Cookie.php
```diff
@@ -0,0 +1,146 @@
...
+        return ($cookie['autoupgrade'] == $this->encrypt($cookie['id_employee']));
+    }
+
+    /**
+     * @param string $string
+     * @return string MD5 hashed string
+     */
+    private function encrypt($string)
+    {
+        return md5($this->readKey().$string);
```
PierreRambaud: Hum md5 isn't a crypt command and isn't secure to store data :/

classes/Log/StreamedLogger.php
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * {@inherit}
+     */
+    public function log($level, $message, array $context = array())
+    {
+        if (empty($message)) {
+            return;
+        }
+
+        $log = self::$levels[$level] .' - '. $message."\n";
```
PierreRambaud: use PHP_EOL instead of \n

classes/Log/StreamedLogger.php
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Set the verbosity of the logger
+     * 
+     * @param int $filter
+     * @return $this
+     */
+    public function setFilter($filter)
+    {
+        if (!array_key_exists($filter, self::$levels)) {
```
PierreRambaud: using `isset` is at least 100x faster than `array_key_exists`, maybe you can change if you don't need to check if value isn't null

classes/LoggedEvent.php
```diff
@@ -0,0 +1,51 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to http://www.prestashop.com for more information.
+ * 
+ *  @author PrestaShop SA <contact@prestashop.com>
+ *  @copyright  2007-2018 PrestaShop SA
+ *  @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *  International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace Composer\Script {
```
PierreRambaud: Add comment to explain why you add this part of code please

classes/AjaxResponse.php
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    public function getNext()
+    {
+        return $this->next;
+    }
+
+    public function getStatus()
+    {
+        return ($this->getNext() == 'error' ? 'error' : 'ok');
```
PierreRambaud: instead of checking getNext, check for 
`$this->error`

classes/BackupFinder.php
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @return array
+     */
+    private function getBackupDbAvailable($backupPath)
+    {
+        $array = array();
+
+        $files = scandir($backupPath);
+
+        foreach ($files as $file) {
```
PierreRambaud: You can use Symfony FileSytem otherwise.

classes/ChannelInfo.php
```diff
@@ -0,0 +1,119 @@
+<?php
+/**
+ * 2007-2017 PrestaShop
```
PierreRambaud: 2017 :)

classes/ErrorHandler.php
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Create a json encoded
+     * @param string $log
+     * @return string
+     */
+    public function generateJsonLog($log)
+    {
+        return '{"nextErrors":["'.addcslashes($log, '"\\').'"],"error":true,"next":"error"}';
```
PierreRambaud: Hum, use json_encode instead can have a surprise with only addcslashes :)

classes/ErrorHandler.php
```diff
@@ -0,0 +1,164 @@
...
+     * @param string $file
+     * @param int $line
+     * @param int $type Level of criticity
+     * @param string $message
+     * @param bool $display
+     */
+    protected function report($file, $line, $type, $message, $trace = null, $display = false)
+    {
+        if ($type >= Logger::CRITICAL) {
+            http_response_code(500);
```
PierreRambaud: Can use Symfony const  https://api.symfony.com/3.4/Symfony/Component/HttpFoundation/Response.html instead of 500

classes/ErrorHandler.php
```diff
@@ -0,0 +1,164 @@
...
+        }
+        $log = "[INTERNAL] $file line $line - $message";
+        if (!empty($trace)) {
+            $log .= "\n".$trace;
```
PierreRambaud: PHP_EOL instead \n

classes/Log/LegacyLogger.php
```diff
@@ -0,0 +1,110 @@
...
+    protected $lastInfo = '';
+
+    /**
+     * @var Resource File descriptor of the log file
+     */
+    protected $fd = null;
+
+    public function __constuct($fileName = null)
+    {
+        if (!is_null($fileName)) {
```
PierreRambaud: !== is faster than is_null 
FYI: https://stackoverflow.com/questions/8228837/is-nullx-vs-x-null-in-php
There are many bench about why is_null is slow =)

classes/Parameters/FileConfigurationStorage.php
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * @param string $fileName File name to load
+     * @return mixed or array() as default value
+     */
+    public function load($fileName = '')
+    {
+        $configFilePath = $this->configPath . $fileName;
+        $config = array();
+
+        if (!empty($configFilePath) && file_exists($configFilePath)) {
```
PierreRambaud: !empty useless

classes/Parameters/FileConfigurationStorage.php
```diff
@@ -0,0 +1,137 @@
...
+        }
+        return $files;
+    }
+
+    /**
+     * Delete all temporary files in the config folder
+     */
+    public function cleanAll()
+    {
+        $filesystem = new Filesystem;
```
PierreRambaud: Maybe you can call `new FileSystem` just once?

classes/Parameters/UpgradeFileNames.php
```diff
@@ -0,0 +1,151 @@
...
+ */
+class UpgradeFileNames
+{
+    /**
+     * configFilename contains all configuration specific to the autoupgrade module
+     *
+     * @var string
+     * @access public
+     */
+    const configFilename = 'config.var';
```
PierreRambaud: Const in uppercase please

classes/PrestashopConfiguration.php
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * @return array Details of the filesystem permission check
+     */
+    protected function getRootWritableDetails()
+    {
+        $result = array();
+        // Root directory permissions cannot be checked recursively anymore, it takes too much time
+        $result['root_writable'] =  ConfigurationTest::test_dir('/', false, $report);
+        $result['root_writable_report'] = $report ? $report : true; // Avoid null in the array as it makes the shop non-compliant
```
PierreRambaud: If report can be a boolean or null, maybe cast it in boolean instead using ternary?

classes/State.php
```diff
@@ -0,0 +1,319 @@
...
+     */
+    private $warning_exists = false;
+
+    /**
+     * @param array $savedState from another request
+     */
+    public function importFromArray(array $savedState)
+    {
+        foreach($savedState as $name => $value) {
+            if (!empty($value) && property_exists(__CLASS__, $name)) {
```
PierreRambaud: You can use `property_exists($this, $name);` to be more explicit

classes/State.php
```diff
@@ -0,0 +1,319 @@
...
+        }
+        return $this;
+    }
+
+    /**
+     * @param string $encodedData
+     */
+    public function importFromEncodedData($encodedData)
+    {
+        $decodedData = json_decode(base64_decode($encodedData), true);
```
PierreRambaud: Maybe test for nextParams key availability?

classes/State.php
```diff
@@ -0,0 +1,319 @@
...
+     * @return array of class properties for export
+     */
+    public function export()
+    {
+        return get_object_vars($this);
+    }
+
+    public function initDefault(Upgrader $upgrader, $prodRootDir, $version)
+    {
+        $postData = 'version='.$this->getInstallVersion().'&method=listing&action=native&iso_code=all';
```
PierreRambaud: use http_build_query instead (http://php.net/manual/fr/function.http-build-query.php)

classes/State.php
```diff
@@ -0,0 +1,319 @@
...
+        $this->setModulesAddons($modules_addons);
+
+        // installedLanguagesIso is used to merge translations files
+        $installedLanguagesIso = array_map(
+            function($v) { return $v['iso_code']; },
+            \Language::getIsoIds(false)
+        );
+        $this->setInstalledLanguagesIso($installedLanguagesIso);
+
+        $rand = dechex(mt_rand(0, min(0xffffffff, mt_getrandmax())));
```
PierreRambaud: Can you explain why you use this function?

classes/TaskRunner/AbstractTask.php
```diff
@@ -0,0 +1,122 @@
...
+        $this->translator = $this->container->getTranslator();
+        $this->checkTaskMayRun();
+    }
+
+    /**
+     * @return sting base64 encoded data from AjaxResponse
+     */
+    public function getEncodedResponse()
+    {
+        return base64_encode($this->getResponse()->getJson());
```
PierreRambaud: `base64_encode($this->getJsonReponse());` to remove duplicate

classes/TaskRunner/AllUpgradeTasks.php
```diff
@@ -0,0 +1,147 @@
...
+{
+    const initialTask = 'upgradeNow';
+
+    private $step = self::initialTask;
+
+    public function run()
+    {
+        $requireRestart = false;
+        while ($this->canContinue() && !$requireRestart) {
+            echo "\n=== Step ".$this->step."\n";
```
PierreRambaud: PHP_EOL instead of \n

classes/TaskRunner/AllUpgradeTasks.php
```diff
@@ -0,0 +1,147 @@
...
+    {
+        if (empty($this->step)) {
+            return false;
+        }
+
+        if ($this->error) {
+            return false;
+        }
+
+        return ! in_array($this->step, array('error'));
```
PierreRambaud: Could be `return $this->step !== 'error';`

classes/TaskRunner/Rollback/RestoreDb.php
```diff
@@ -0,0 +1,244 @@
...
+    {
+        $databaseTools = new Database($this->container->getDb());
+        $ignore_stats_table = array(
+            _DB_PREFIX_.'connections',
+            _DB_PREFIX_.'connections_page',
+            _DB_PREFIX_.'connections_source',
+            _DB_PREFIX_.'guest',
+            _DB_PREFIX_.'statssearch'
+        );
+        $start_time = time();
```
PierreRambaud: CamelCase please :D

classes/TaskRunner/Upgrade/BackupDb.php
```diff
@@ -0,0 +1,300 @@
...
+                                            }
+                                            break;
+                                        }
+                                    }
+                                }
+                            }
+                            $s = rtrim($s, ',');
+
+                            if ($i < $sizeof) {
+                                $s .= "),\n";
```
PierreRambaud: PHP_EOL again please

classes/TaskRunner/Upgrade/RemoveSamples.php
```diff
@@ -0,0 +1,118 @@
...
+        $this->stepDone = false;
+        $this->next = 'removeSamples';
+
+        $removeList = $this->container->getState()->getRemoveList();
+        $latestPath = $this->container->getProperty(UpgradeContainer::LATEST_PATH);
+        // remove all sample pics in img subdir
+        // This part runs at the first call of this step
+        if (null === $removeList) {
+            $removeList = $this->container->getFilesystemAdapter()->listSampleFiles(array(
+                $latestPath.'/prestashop/img/c', '.jpg',
```
PierreRambaud: This array look weird

classes/TaskRunner/Upgrade/RemoveSamples.php
```diff
@@ -0,0 +1,118 @@
...
+                $latestPath.'/prestashop/img', 'logo_stores.gif',
+                $latestPath.'/prestashop/modules/editorial', 'homepage_logo.jpg',
+                // remove all override present in the archive
+                $latestPath.'/prestashop/override', '.php',
+            ));
+
+            $this->container->getState()->setRemoveList($removeList);
+
+            if (count($removeList)) {
+                $this->logger->debug($this->translator->trans('Starting to remove %s sample files', array(count($removeList)), 'Modules.Autoupgrade.Admin'));
```
PierreRambaud: Don't use more than 130 characters per line, can't see anything in github :D

classes/TaskRunner/Upgrade/UpgradeFiles.php
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /**
+     * First call of this task needs a warmup, where we load the files list to be upgraded.
+     * 
+     * @return boolean
+     */
+    protected function warmUp()
+    {
+        $admin_dir = str_replace($this->container->getProperty(UpgradeContainer::PS_ROOT_PATH).DIRECTORY_SEPARATOR, '', $this->container->getProperty(UpgradeContainer::PS_ADMIN_PATH));
```
PierreRambaud: You can maybe set a variable containing `$this->container->getProperty(UpgradeContainer::LATEST_PATH).DIRECTORY_SEPARATOR` to make this page more readable.

classes/TaskRunner/Upgrade/UpgradeFiles.php
```diff
@@ -0,0 +1,284 @@
...
+            DIRECTORY_SEPARATOR.'vendor'.DIRECTORY_SEPARATOR.'composer'.DIRECTORY_SEPARATOR.'autoload_static.php',
+            DIRECTORY_SEPARATOR.'vendor'.DIRECTORY_SEPARATOR.'composer'.DIRECTORY_SEPARATOR.'include_paths.php',
+            DIRECTORY_SEPARATOR.'vendor'.DIRECTORY_SEPARATOR.'composer',
+            DIRECTORY_SEPARATOR.'vendor',
+        );
+
+        foreach ($filesToMoveToTheBeginning as $file) {
+            if ($key = array_search($file, $list_files_to_upgrade)) {
+                unset($list_files_to_upgrade[$key]);
+                $list_files_to_upgrade = array_merge(array($file), $list_files_to_upgrade);
```
PierreRambaud: Instead of using array_merge, you can use `$listFilesToUpgrade[] = $file;`

classes/Tools14.php
```diff
@@ -24,2244 +24,2241 @@
...
  */
 
-if (!defined('_PS_DISPLAY_COMPATIBILITY_WARNING_'))
-    define('_PS_DISPLAY_COMPATIBILITY_WARNING_', false);
+namespace PrestaShop\Module\AutoUpgrade;
 
-if(!class_exists('Tools14',false))
+use Tab;
+
+class Tools14
```
PierreRambaud: Why Tools14?

classes/Tools14.php
```diff
@@ -24,2244 +24,2241 @@
...
-                    $url .= '?'.$explode[1];
-            }
+    /**
+     * getProtocol return the set protocol according to configuration (http[s])
+     * @param Boolean true if require ssl
+     * @return String (http|https)
+     */
+    public static function getProtocol($use_ssl = null)
+    {
+        return (!is_null($use_ssl) && $use_ssl ? 'https://' : 'http://');
```
PierreRambaud: just use `return $use_ssl ? 'https://' : 'http://';`, no need to check if variable isn't null :)

PierreRambaud: Is it hte same function as `getCurrentUrlProtocolPrefix`?

classes/Tools14.php
```diff
@@ -24,2244 +24,2241 @@
...
+     * Get a value from $_POST / $_GET
+     * if unavailable, take a default value
+     *
+     * @param string $key Value key
+     * @param mixed $defaultValue (optional)
+     * @return mixed Value
+     */
+    public static function getValue($key, $defaultValue = false)
+    {
+        if (!isset($key) OR empty($key) OR !is_string($key))
```
PierreRambaud: !isset and empty are recurrent, only empty is needed

classes/Tools14.php
```diff
@@ -24,2244 +24,2241 @@
...
-         * @return mixed Value
-         */
-        public static function getValue($key, $defaultValue = false)
-        {
-            if (!isset($key) OR empty($key) OR !is_string($key))
-                return false;
-            $ret = (isset($_POST[$key]) ? $_POST[$key] : (isset($_GET[$key]) ? $_GET[$key] : $defaultValue));
+    public static function getIsset($key)
+    {
+        if (!isset($key) OR empty($key) OR !is_string($key))
```
PierreRambaud: same as above

classes/Tools14.php
```diff
@@ -24,2244 +24,2241 @@
...
+        if (is_int($currency))
+            $currency = Currency::getCurrencyInstance((int)($currency));
+        $c_char = (is_array($currency) ? $currency['sign'] : $currency->sign);
+        $c_format = (is_array($currency) ? $currency['format'] : $currency->format);
+        $c_decimals = (is_array($currency) ? (int)($currency['decimals']) : (int)($currency->decimals)) * _PS_PRICE_DISPLAY_PRECISION_;
+        $c_blank = (is_array($currency) ? $currency['blank'] : $currency->blank);
+        $blank = ($c_blank ? ' ' : '');
+        $ret = 0;
+        if (($isNegative = ($price < 0)))
+            $price *= -1;
```
PierreRambaud: abs?

classes/Tools14.php
```diff
@@ -24,2244 +24,2241 @@
...
-                if ($to_currency)
-                    $price *= $c_rate;
-                else
-                    $price /= $c_rate;
-            }
+        $c_id = (is_array($currency) ? $currency['id_currency'] : $currency->id);
+        $c_rate = (is_array($currency) ? $currency['conversion_rate'] : $currency->conversion_rate);
 
-            return $price;
+        if ($c_id != (int)(Configuration::get('PS_CURRENCY_DEFAULT')))
```
PierreRambaud: $c_id look useless.

classes/Tools14.php
```diff
@@ -24,2244 +24,2241 @@
...
+     * @param string $dirname Directory name
+     */
+    public static function deleteDirectory($dirname, $delete_self = true)
+    {
+        $dirname = rtrim($dirname, DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR;
+        if (file_exists($dirname))
+            if ($files = scandir($dirname))
+            {
+                foreach ($files as $file)
+                    if ($file != '.' && $file != '..' && $file != '.svn')
```
PierreRambaud: This function must be cleaned or you can use Symfony FileSystem as above to delete directories :)

classes/Tools14.php
```diff
@@ -24,2244 +24,2241 @@
...
-            return (self::dieObject($object, false));
-        }
+    /**
+     * ALIAS OF dieObject() - Display an error with detailed object
+     *
+     * @param object $object Object to display
+     */
+    public static function d($object, $kill = true)
+    {
+        return (self::dieObject($object, $kill = true));
```
PierreRambaud: change $kill = true by $kill :)

classes/Tools14.php
```diff
@@ -24,2244 +24,2241 @@
...
+     * Get token to prevent CSRF
+     *
+     * @param string $token token to encrypt
+     */
+    public static function getToken($page = true)
+    {
+        global $cookie;
+        if ($page === true)
+            return (self::encrypt($cookie->id_customer.$cookie->passwd.$_SERVER['SCRIPT_NAME']));
+        else
```
PierreRambaud: useless else

classes/Tools14.php
```diff
@@ -24,2244 +24,2241 @@
...
-				ORDER BY c.level_depth ASC
-				LIMIT '.(int)$category['level_depth']);
-
-                    $n = 1;
-                    $nCategories = (int)sizeof($categories);
-                    foreach ($categories AS $category)
-                    {
-                        $fullPath .=
-                            (($n < $nCategories OR $linkOntheLastItem) ? '<a href="'.self::safeOutput($link->getCategoryLink((int)$category['id_category'], $category['link_rewrite'])).'" title="'.htmlentities($category['name'], ENT_NOQUOTES, 'UTF-8').'">' : '').
-                            htmlentities($category['
```
PierreRambaud: This function looks very complicated. Can it be a template that we copy & paste with parametres, such as twig or smarty file?

PierreRambaud: useless else there is already a return statement

PierreRambaud: useless else, already a return statement

PierreRambaud: property_exists was added in php 5.1, we can maybe only use the native function and remove this static function :)

PierreRambaud: Look like a PHP4 function.

PierreRambaud: Must be deleted later like property_exists

classes/Twig/Block/UpgradeButtonBlock.php
```diff
@@ -0,0 +1,223 @@
...
+     */
+    private $twig;
+
+    /**
+     * @var TranslatorComponent
+     */
+    private $translator;
+
+    /**
+     * @var \Upgrader
```
PierreRambaud: \Upgrader or PrestaShop\Module\AutoUpgrade\Upgrader?

classes/Twig/Block/UpgradeButtonBlock.php
```diff
@@ -0,0 +1,223 @@
...
+
+    /**
+     * @return array
+     */
+    private function getOptChannels()
+    {
+        $translator = $this->translator;
+        return array(
+            // Hey ! I'm really using a fieldset element to regroup fields ?! !
+            array("useMajor", "major", $translator->trans('Major release', array(), 'Modules.Autoupgrade.Admin')),
```
PierreRambaud: Use single quote please.

classes/UpgradeContainer.php
```diff
@@ -0,0 +1,466 @@
...
+    {
+        if (!is_null($this->twig)) {
+            return $this->twig;
+        }
+
+        // Using independant template engine for 1.6 & 1.7 compatibility
+        $loader = new \Twig_Loader_Filesystem();
+        $loader->addPath(realpath(__DIR__.'/..').'/views/templates', 'ModuleAutoUpgrade');
+        $twig = new \Twig_Environment($loader, array(
+            //'cache' => '/path/to/compilation_cache',
```
PierreRambaud: use namespace and remove useless comments

classes/UpgradeContainer.php
```diff
@@ -0,0 +1,466 @@
...
+    }
+
+    public function getUpgradeConfigurationStorage()
+    {
+        return new UpgradeConfigurationStorage($this->getProperty(self::WORKSPACE_PATH).DIRECTORY_SEPARATOR);
+    }
+
+    public function getWorkspace()
+    {
+        if (!is_null($this->workspace)) {
```
PierreRambaud: Prefer usage of !== null instead of is_null

views/templates/block/upgradeButtonBlock.twig
```diff
@@ -0,0 +1,169 @@
...
+            {% if showUpgradeButton %}
+            <a class="button button-autoupgrade btn btn-default" href="index.php?tab=AdminSelfUpgrade&amp;token={{ token }}&amp;refreshCurrentVersion=1">{{ 'Refresh the page'|trans }}</a> -
+            {% else %}
+            <a class="button button-autoupgrade btn btn-primary" href="index.php?tab=AdminSelfUpgrade&amp;token={{ token }}&amp;refreshCurrentVersion=1">{{ 'Check if a new version is available'|trans }}</a> -
+            {% endif %}
+            <span style="font-style: italic; font-size: 11px;">{% if lastVersionCheck %}
+                {{ 'Last check: %s'|trans([lastVersionCheck|date('Y-m-d H:i:s')]) }}
+                {% else %}
+                {{ 'Last check: never'|trans }}
+                {% endif %}</span>
```
PierreRambaud: Missing break line

classes/ErrorHandler.php
Quetzacoalt91: I can't, this module can be used un PrestaShop 1.6, which does not provide Symfony

classes/PrestashopConfiguration.php
Quetzacoalt91: Hehe, the variable is `null` if no error occurred, and containsa string if an error needs to be reported.

classes/State.php
Quetzacoalt91: Nope, this comes directly from the previous version to generate a random value. https://github.com/PrestaShop/autoupgrade/blob/6e863f7b2d196b1128a5a6018056a917242a612e/AdminSelfUpgrade.php#L651

classes/TaskRunner/AllUpgradeTasks.php
Quetzacoalt91: Yep, I was suspecting others steps to be listed.

classes/Tools14.php
Quetzacoalt91: Backward compatibility Tools class for PrestaShop 1.4.
I wanted to keep it to avoid relying on the core Tools.

classes/TaskRunner/Upgrade/UpgradeFiles.php
Quetzacoalt91: No, read the previous line and the name of the different variables.
We absolutely need to move these files at the *beginning* of the array.

classes/Twig/Block/UpgradeButtonBlock.php
Quetzacoalt91: PrestaShop\Module\AutoUpgrade\Upgrader. The namespace has been added after @eternoendless created these classes.

classes/UpgradeContainer.php
Quetzacoalt91: Even better, I will enable the cache from now. I couldn't before because getting access to some path was difficult.

classes/UpgradeContainer.php
```diff
@@ -311,7 +311,10 @@ public function getLogger()
             return $this->logger;
         }
 
-        $logFile = $this->getProperty(self::TMP_PATH).DIRECTORY_SEPARATOR.'log.txt';
+        $logFile = null;
+        if (is_writeable($this->getProperty(self::TMP_PATH))) {
```
PierreRambaud: We use `is_writable` in the core, even translator doesn't like `writeable` :P

Quetzacoalt91:
Error handler has been improved to flush the queued logs in case of error. It used to display only the last error.

Test case:
capture du 2018-06-08 17-03-45

kpodemski:
@Quetzacoalt91 have you tried to upgrade with a name of admin directory different than admin and admin-dev?

Quetzacoalt91:
@kpodemski yep, I tried with something like `hahaha1234`.

I your server running under Windows? I haven't tried this environment yet.

kpodemski:
@Quetzacoalt91 yeah, i've tried with an `back-office` name on Windows and `admin-dev` on OSX, `back-office` is problematic

Quetzacoalt91:
I have to test it with IIS as well, but I can already tell that it works with Apache. Interesting module errors though.

upgrade wampp

classes/ErrorHandler.php
```diff
@@ -130,7 +130,12 @@ public function fatalHandler()
      */
     public function generateJsonLog($log)
     {
-        return '{"nextErrors":["'.addcslashes($log, '"\\').'"],"error":true,"next":"error"}';
+        return json_encode(array(
```
PierreRambaud: :+1:

eternoendless:
yolo

PR #74, opened 2018-02-05
Move composer files to move on the first place

Quetzacoalt91:
During the upgrade, file update / deletion is made is several ajax calls (400 files per call).

Sometimes, depending on the number of file in the release package, all the composer files are not copied in the same call. This trigger errors about undefined class etc, and this breaks the whole process. 

The objective is to move these files at the beginning of the array instead of the end. We then make sure all these files are handled in the same call.

Fixes http://forge.prestashop.com/browse/BOOM-4742

### DIFF
--- a/AdminSelfUpgrade.php
+++ b/AdminSelfUpgrade.php
@@ -1651,7 +1651,7 @@ class AdminSelfUpgrade extends AdminSelfTab
             // also add files to remove
             $list_files_to_upgrade = array_merge($list_files_diff, $list_files_to_upgrade);
 
-            $filesToMoveToTheEnd = array(
+            $filesToMoveToTheBeginning = array(
                 DIRECTORY_SEPARATOR.'vendor'.DIRECTORY_SEPARATOR.'autoload.php',
                 DIRECTORY_SEPARATOR.'vendor'.DIRECTORY_SEPARATOR.'composer'.DIRECTORY_SEPARATOR.'ClassLoader.php',
                 DIRECTORY_SEPARATOR.'vendor'.DIRECTORY_SEPARATOR.'composer'.DIRECTORY_SEPARATOR.'autoload_classmap.php',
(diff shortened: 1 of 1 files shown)


AdminSelfUpgrade.php
```diff
@@ -1663,10 +1663,10 @@ public function ajaxProcessUpgradeFiles()
                 DIRECTORY_SEPARATOR.'vendor'.DIRECTORY_SEPARATOR.'composer'.DIRECTORY_SEPARATOR.'include_paths.php',
             );
 
-            foreach ($filesToMoveToTheEnd as $file) {
+            foreach ($filesToMoveToTheBeginning as $file) {
                 if ($key = array_search($file, $list_files_to_upgrade)) {
-                  $list_files_to_upgrade[] = $list_files_to_upgrade[$key];
-                  unset($list_files_to_upgrade[$key]);
+                    unset($list_files_to_upgrade[$key]);
+                    $list_files_to_upgrade = array_merge(array($file), $list_files_to_upgrade);
```
eternoendless: This will reindex the array every time.

Maybe it would be better to merge it outside the loop so the array is reindexed only once.

```php
if (!empty($filesToMoveToTheBeginning)) {

  foreach ($filesToMoveToTheBeginning as $file) {
    if ($key = array_search($file, $list_files_to_upgrade)) {
      unset($list_files_to_upgrade[$key]);
    }
  }

  $list_files_to_upgrade = array_merge($filesToMoveToTheBeginning, $list_files_to_upgrade);
}
```

eternoendless (changes requested):

AdminSelfUpgrade.php
Quetzacoalt91: And what if the file does not exist?

Quetzacoalt91: If you absolutely want to have only one action on the array:

```php

            foreach ($filesToMoveToTheBeginning as $index => $file) {
                if ($key = array_search($file, $list_files_to_upgrade)) {
                    unset($list_files_to_upgrade[$key]);
                } else {
                    unset($filesToMoveToTheBeginning[$index]);
                }
            }
            $list_files_to_upgrade = array_merge($filesToMoveToTheBeginning, $list_files_to_upgrade);
```

But IMHO the current is easier to read, although less performant.

eternoendless: Ok then

eternoendless (approved):

eternoendless:
Thank you @Quetzacoalt91

PR #77, opened 2018-03-07
Continue on broken symlink in directory listing

Quetzacoalt91:
When a **broken** symbolic link exists on the Filesystem, the module tries to backup it before proceeding with the upgrade.
Because the link destination does not exist, there is nothing to save. This PR asks the module to bypass this case.

Fixes http://forge.prestashop.com/browse/BOOM-4957


eternoendless:
Thank you @Quetzacoalt91

PR #84, opened 2018-06-21
Always load default conf data before saved conf files

Quetzacoalt91:
This PR covers a case where the data saved in files may change between two release. It appeared upgrading a shop under 1.6 was impossible, because of download issues.

From now, even if the data does not exist in the saved content, it will be declared in the Configuration class.

Fixes http://forge.prestashop.com/browse/BOOM-4960


Quetzacoalt91:
Merging as it fixes an issue with the download step.

PR #94, opened 2018-06-29
Fix missing opening <?php in settings file

Quetzacoalt91:
Tag `<?php` was missing in generated config/settings.php files...

We used a generated settings file as a comparison, to check files are properly created.

### DIFF
--- a/classes/UpgradeTools/SettingsFileWriter.php
+++ b/classes/UpgradeTools/SettingsFileWriter.php
@@ -65,12 +65,12 @@ class SettingsFileWriter
         $filesystem->copy($filePath, $filePath.'.bck');
 
         $fd = fopen($filePath, 'w');
+        fwrite($fd, '<?php'.PHP_EOL);
         foreach ($data as $name => $value) {
-            if (false === fwrite($fd, 'define(\''.$name.'\', \''.$this->checkString($value).'\');'."\n")) {
+            if (false === fwrite($fd, 'define(\''.$name.'\', \''.$this->checkString($value).'\');'.PHP_EOL)) {
                 throw new UpgradeException($this->translator->trans('Error when generating new settings.inc.php file.'));
             }
         }
-        fwrite($fd, '?>'."\n");
         fclose($fd);
     }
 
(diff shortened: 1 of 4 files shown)


khouloudbelguith:
Hi,
It is Ok for me, the upgrade is done successfully from 16.1.18 to 1.6.1.20.
Best regards, Khouloud

classes/UpgradeTools/SettingsFileWriter.php
```diff
@@ -65,12 +65,12 @@ public function writeSettingsFile($filePath, $data)
         $filesystem->copy($filePath, $filePath.'.bck');
 
         $fd = fopen($filePath, 'w');
+        fwrite($fd, '<?php'.PHP_EOL);
         foreach ($data as $name => $value) {
-            if (false === fwrite($fd, 'define(\''.$name.'\', \''.$this->checkString($value).'\');'."\n")) {
+            if (false === fwrite($fd, 'define(\''.$name.'\', \''.$this->checkString($value).'\');'.PHP_EOL)) {
```
PierreRambaud: :+1:

tests/SettingsFileWriterTest.php
```diff
@@ -0,0 +1,86 @@
...
+            '_PS_CACHE_ENABLED_' => '0',
+            '_COOKIE_KEY_' => 'dedeedededede',
+            '_COOKIE_IV_' => 'wololo',
+            '_PS_CREATION_DATE_' => '2018-05-16',
+            '_PS_VERSION_' => '1.6.1.18',
+            '_RIJNDAEL_KEY_' => 'zrL1GDp2oqDoXFss',
+            '_RIJNDAEL_IV_' => 'QSt/I95YtA==',
+        );
+
+        $file = sys_get_temp_dir().DIRECTORY_SEPARATOR.'temp_settings.php';
```
PierreRambaud: `tempnam(sys_get_temp_dir(), 'PSS');` is maybe better because creating a uniq file, so you don't need to touch it, and you prevent error if test are running in the same time. Otherwise : http://php.net/manual/fr/function.tmpfile.php wdyt?

Quetzacoalt91: Thanks for the tip, I used `tempname` as suggested.

However, as I'm supposed to send a file name, not a resource, I could not use the `tmpfile()` method.

PR #128, opened 2018-08-10
Check HTTP codes with curl after upgrade

Quetzacoalt91:
After an upgrade, we check:
* Front office returns an HTTP code 503 (maintenance page)
* Back office returns an HTTP code 200 (OK)


.travis.yml
```diff
@@ -23,6 +23,11 @@ install:
 before_script:
   - bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8001/index.php)" != "200" ]]; do sleep 5; done'
   - docker exec -ti prestashop_autoupgrade cp modules/autoupgrade/ -R admin-dev
+  # Front office in maintenance mode -> HTTP code 503 expected
+  - [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8001/index.php)" == "503" ]]
```
PierreRambaud: `''` are unneeded it's like http_code isn't in the quote context, can use single or even no quote (tried on debian)

Quetzacoalt91: Actually the YML can't be parsed because of the `[[`. This need a update.

PR #171, opened 2018-09-13
Prevent installation on old PHP versions

### LINKED ISSUE #170: Prevent 1-click upgrade to run on very old version of PHP (5.6 min)
**Describe the bug**

1-click upgrade crashes some store because of syntax errors.

These shops are reported to run very old versions of PHP. Although these versions are accepted by PrestaShop 1.6, they are of supported anymore by the PHP team since several years. Furthermore, latest versions of PrestaShop require newer version of PHP.

capture du 2018-09-12 16-57-00

We won't increase the compatibility range of the module, but we will handle these old versions of PHP by displaying at least an error message.

TO-DO list:
* Display an error message on the BO
* Update the README to display this pre-requisite.

Quetzacoalt91:
Fixes #170 

This PR fixes the syntax errors on very old versions of PHP, but prevent its installation on the concerned shops.

capture du 2018-09-13 10-43-38

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/autoupgrade/171)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -5,6 +5,11 @@
 Provides an automated method to upgrade your shop to the latest version of PrestaShop.
 This module is compatible with all PrestaShop 1.6 & 1.7.
 
+# Prerequisites
+
+* PrestaShop 1.6 or 1.7
+* PHP 5.6+
+
 # Installation
 
 You must have [composer][4] installed on your computer. Then, execute:
--- a/autoupgrade.php
+++ b/autoupgrade.php
@@ -24,8 +24,6 @@
  *  @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  *  International Registered Trademark & Property of PrestaShop SA
  */
-require_once _PS_ROOT_DIR_ . '/modules/autoupgrade/vendor/autoload.php';
-
 class Autoupgrade extends Module
 {
     public function __construct()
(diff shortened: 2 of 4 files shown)


doekia:
Je pense que vous avez complètement raté l'objectif d'un module d'upgrade. Il doit justement fonctionner en php5.2 pour permettre à quelqu'un de migrer son vieux shop puisqu'une vieille version ne fonctionne pas en php7 ou php5.6!

eternoendless:
Hi @doekia 

Since the official language here is English, I'm taking the liberty of translating your comment to English so that others can understand it too. I used Google Translate out of lazyness, but I think it did a pretty good job:

> I think you have completely missed the goal of an upgrade module. It must work in php5.2 to allow someone to migrate his old shop since an old version does not work in php7 or php5.6!

By "old shop" you mean lower than 1.6? This module is able to upgrade shops starting on 1.6 only, and 1.6 should be compatible with 5.6. However, it is possible (and it has been proved true) that shops can be upgraded to 1.7 all the way back from 1.4 in two steps, first upgrading to 1.6 using the latest available module for 1.4, then to 1.7 using the latest version of the upgrade module.

We had three main reasons for upping the minimum compatibility to 5.6:

1. Very old PHP versions pose a security threat for merchants, so they are strongly encouraged to upgrade.
2. Some libraries that we needed for this module are only available on PHP >=5.6
3. We need to make sure that merchants won't upgrade their shops to 1.7 if they don't meet the minimum requirements.

I hope this provides a better insight into our choices.

PR #176, opened 2018-10-09
Fix wordings without domain

eternoendless:
This change is [Reviewable](https://reviewable.io/reviews/prestashop/autoupgrade/176)


Quetzacoalt91:
Thank you @eternoendless

PR #187, opened 2018-11-05
Fix auto reload when updating the config

### LINKED ISSUE #185: The refresh button is not displayed when we use a local archive
**Describe the bug**

In the configuration page, when we use the local archive option, the button "Refresh the page" is not displayed after Clicking "Save".
We should refresh the page manually, so this button is displayed.
image

**Upgrade environment (please complete the following information):**
 - Version you started from: 1.7.4.3 / 1.6.1.22
 - Version you upgrade to: 1.7.4.4
 - Way of upgrade: Local archive

**Expected behavior**
When we click on "save" => the button "Refresh the page" should be automatically displayed.

**Screenshots**
https://drive.google.com/file/d/1mJZwWZVm2WRNWDnXJQ8fCS_K_R4Powbn/view

**Server (please complete the following information):**
 - Web Server: Apache
 - PHP Version: 1.7.4.3 / 1.6.1.22
 - Module Version: 4.4.0

Quetzacoalt91:
Fixes #185

Updating the config will trigger the page reload as before.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/autoupgrade/187)

### DIFF
--- a/js/upgrade.js
+++ b/js/upgrade.js
@@ -241,7 +241,7 @@ $(document).ready(function(){
 });
 
 function showConfigResult(msg, type) {
-  if (type === null) {
+  if (!type) {
     type = "conf";
   }
   var $configResult = $("#configResult");


js/upgrade.js
```diff
@@ -241,7 +241,7 @@ $(document).ready(function(){
 });
 
 function showConfigResult(msg, type) {
-  if (type === null) {
+  if (type === undefined) {
```
jolelievre: maybe compare null or undefined

Quetzacoalt91: I wonder if I should check `!type`

PR #235, opened 2018-12-24
Fix php docs and comments

MathiasReker:


Quetzacoalt91:
That's the last one waiting for review. Thank you @MathiasReker for all these pull-requests.

PR #237, opened 2018-12-24
Heredoc to nowdoc

MathiasReker:


MathiasReker:
Thank you @PierreRambaud. I added PHP compatibility check in my script. So I wont make this mistake in future PR's :)

PR #241, opened 2018-12-24
Pow to exponentiation

MathiasReker:


Quetzacoalt91:
Sorry @MathiasReker, but these methods are not used by the module and I consider this class as deprecated,"to be removed".
I'd rather leave this file untouched when possible.

PR #244, opened 2018-12-24
Non printable character

MathiasReker:


PierreRambaud:
Watchout with translations, they are link to crowdin, waiting for @Quetzacoalt91 feedback.

Quetzacoalt91:
As the translations are generated by PrestaShop, your changes would be lost at the next update.
If something needs to be updated, this would be on the core side.
A specific release will be done with up-to-date translations. :)

PR #248, opened 2018-12-24
Php unit strict

MathiasReker:


Quetzacoalt91:
That one was completely in conflict with #245, it looks like you forgot it. :)

PR #251, opened 2018-12-24
String line ending

MathiasReker:


Quetzacoalt91:
Like for Tools14.php, this file was copy/pasted to avoid using the core files when possible. But as the modified content of this PR is unused by the module, I would rather see it untouched, or eventually deleted.

Regards

PR #255, opened 2018-12-24
Class attributes separation

MathiasReker:


Quetzacoalt91:
I would be okay to add white-spaces between each property if they had their doc available. As this is not the case, I currently prefer to leave them as-is until further change.

PR #256, opened 2018-12-24
Increment style

MathiasReker:


Quetzacoalt91:
As this PR deals with an external lib, I would leave untouched.

PR #257, opened 2018-12-24
Static lambda

MathiasReker:


classes/State.php
```diff
@@ -142,7 +142,7 @@ public function initDefault(Upgrader $upgrader, $prodRootDir, $version)
 
         // installedLanguagesIso is used to merge translations files
         $installedLanguagesIso = array_map(
-            function ($v) { return $v['iso_code']; },
+            static function ($v) { return $v['iso_code']; },
```
PierreRambaud: Since php 5.4 and this module need to be avaiable for php 5.2 due to PrestaShop 1.6.

MathiasReker:
Thank you @PierreRambaud. I added PHP compatibility check in my script. So I wont make this mistake in future PR's :)

PR #258, opened 2018-12-24
Ordered imports

MathiasReker:


PierreRambaud:
Waiting for feedback to check if this one is really useful, (I'm the only one who check orders :roll_eyes: )

matks:
I find it useful but it will create git conflicts. Is it worth it, considering that we look very few at imported classes ?

matks:
I would be more in favor of a "scout" approach: everytime you work on a class, first you apply refactoring / lint steps such as this one, then you work on it.

MathiasReker:
I commented out this part from my script for now :)

MathiasReker:
Thank you :)

PR #260, opened 2018-12-24
Multiline whitespace before semicolons

MathiasReker:


classes/UpgradeSelfCheck.php
```diff
@@ -278,8 +277,7 @@ private function checkShopIsDeactivated()
     {
         return
             !Configuration::get('PS_SHOP_ENABLE')
-            || (isset($_SERVER['HTTP_HOST']) && in_array($_SERVER['HTTP_HOST'], array('127.0.0.1', 'localhost')))
-        ;
+            || (isset($_SERVER['HTTP_HOST']) && in_array($_SERVER['HTTP_HOST'], array('127.0.0.1', 'localhost')));
```
PierreRambaud: @Quetzacoalt91 Can we add the ipv6 notation here? `[::1]`

Quetzacoalt91: Yep, we should do that. But I guess in another PR.

Quetzacoalt91: Created #276

PR #263, opened 2018-12-24
Random api migration

MathiasReker:


Quetzacoalt91:
What is the added value of this method? In a performance point of view the change is not so visible.

```
<?php
    $START = microtime(true);
    for ($i = 1; $i < 1000000; ++$i) {
        $j = rand(100,2000);
    }
    $END = microtime(true) - $START;
    print "Short rand() took $END seconds\n";

    $START = microtime(true);
    for ($i = 1; $i < 1000000; ++$i) {
        $j = mt_rand(100,2000);
    }
    $END = microtime(true) - $START;
    print "Short mt_rand() took $END seconds\n";
```

```
Short rand() took 0.096274852752686 seconds
Short mt_rand() took 0.09580397605896 seconds
```

MathiasReker:
mt_rand uses a better randomization algorithm (Mersenne Twist), which requires more random numbers to be known before the seed can be determined and is faster. This does not mean that mt_rand is, by definition, faster than rand is, this only means that the way the numbers are generated is faster, and appears to have no real impact on the function's performance, as other answers here have demonstrated. :-)

MathiasReker:
mt_rand was introduced to fix the problems that existed

PR #265, opened 2018-12-24
Butify yml files

MathiasReker:


tests/docker-compose.yml
```diff
@@ -7,9 +7,25 @@ services:
       SELENIUM_HOST: selenium-chrome
       URL: prestashop-web
     depends_on:
-      - "prestashop"
-      - "selenium-chrome"
-    command: ["/tmp/wait-for-it.sh", "--timeout=200", "--strict", "prestashop:80", "--", "/tmp/node_modules/mocha/bin/mocha /tmp/test/campaigns/high/13_installation", "--URL=prestashop-web", "--SELENIUM=selenium-chrome", "--DIR=/home/seluser/Downloads/", "--DB_SERVER=db", "--DB_USER=root", "--DB_PASSWD=admin", "--RCTARGET=/tmp/", "--RCLINK=http://localhost/prestashop_1.7.3.0-RC1.zip"]
+      - 'prestashop'
+      - 'selenium-chrome'
+    command:
+      [
```
PierreRambaud: Not sure about this, did you run docker-compose?

MathiasReker: I don't have docker-compose installed. Should multi-line of `command ` be a problem?

PierreRambaud: Just never see this kind of way to do that, always with `-` or `|` :smile:

Quetzacoalt91:
Checked with `docker-compose config`.

PR #267, opened 2018-12-26
Remove useless return

MathiasReker:


Quetzacoalt91:
Hi @MathiasReker,

For this PR, I would rather leave these classes untouched.
The first one (Tools14) is deprecated and would accept only critical bug fixes, and the second one is a external lib which does not need to be updated manually in my opinion. If it needs some changes, it should be applied on the original project.

PR #284, opened 2019-02-27
Display a notice if PHP is below v7.1

Quetzacoalt91:
This PRs notifies when the running PHP server reached its end of life at 2 places, but the module remains compatible with PHP v5.6.

## Module configuration: Check list

capture d ecran du 2019-02-27 11-32-28

Not OK: ` The PHP version your server is running on is obsolete and needs to be upgraded. Learn more why or ignore.`
OK: `Your server is running on a supported PHP version.`

## Backoffice dashboard

capture d ecran du 2019-02-27 11-32-54

```
<head>PHP Version notice

<body>The PHP version your shop is running on is insecure.
It reached its end-of-life, which means it won't get security updates anymore and projects will stop supporting it.

Upgrading will keep your shop secured and performant.

<button>Learn more about PHP and upgrading
```

### DIFF
--- a/autoupgrade.php
+++ b/autoupgrade.php
@@ -135,7 +135,7 @@ class Autoupgrade extends Module
             return $this->_abortInstall($this->trans('Unable to create the directory "%s"', array(_PS_ROOT_DIR_ . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'xml'), 'Modules.Autoupgrade.Admin'));
         }
 
-        return parent::install();
+        return parent::install() && $this->registerHookAndSetToTop('dashboardZoneOne');
     }
 
     public function uninstall()
@@ -149,7 +149,54 @@ class Autoupgrade extends Module
         // Remove the 1-click upgrade working directory
         self::_removeDirectory(_PS_ADMIN_DIR_ . DIRECTORY_SEPARATOR . 'autoupgrade');
 
+        Configuration::deleteByName('PS_AUTOUP_IGNORE_REQS');
+        Configuration::deleteByName('PS_AUTOUP_IGNORE_PHP_UPGRADE');
+
         return parent::uninstall();
+    }
+
+    private function registerHookAndSetToTop($hookName)
+    {
+        if ($this->registerHook($hookName)) {
+            // Update module position in Dashboard
+            $query = "SELECT id_hook FROM "._DB_PREFIX_."hook WHERE name = '" . pSQL($hookName) . "'";
+            $result = Db::getInstance()->ExecuteS($query);
+            $id_hook = $result['0']['id_hook'];
+
+            $this->updatePosition((int) $id_hook, 0);
+            return true;
+        }
+        return false;
+    }
+
+    public function hookDashboardZoneOne($params)
+    {
+        // Display panel if PHP is not supported by the community
+        require_once __DIR__ . '/vendor/autoload.php';
+
+        $upgradeContainer = new \PrestaShop\Module\AutoUpgrade\UpgradeContainer(_PS_ROOT_DIR_, _PS_ADMIN_DIR_);
+        $upgrader = $upgradeContainer->getUpgrader();
+        $upgradeSelfCheck = new  \PrestaShop\Module\AutoUpgrade\UpgradeSelfCheck(
+            $upgrader,
+            _PS_ROOT_DIR_,
+            _PS_ADMIN_DIR_,
+            __DIR__
+        );
+
+        $upgradeNotice = $upgradeSelfCheck->isPhpUpgradeRequired();
+        if (false === $upgradeNotice) {
+            return false;
+        }
+
+        $this->context->controller->addCSS($this->_path.'/css/styles.css');
+        $this->context->controller->addJS($this->_path.'/js/dashboard.js');
+
+        $this->context->smarty->assign([
+            'ignore_link' => Context::getContext()->link->getAdminLink('AdminSelfUpgrade') . '&ignorePhpOutdated=1',
+            'learn_more_link' => $upgradeNotice,
+        ]);
+
+        return $this->context->smarty->fetch($this->local_path.'views/templates/hook/dashboard_zone_one.tpl');
     }
 
     public function getContent()
--- a/autoupgrade.php
+++ b/autoupgrade.php
@@ -159,13 +159,15 @@ class Autoupgrade extends Module
     {
         if ($this->registerHook($hookName)) {
             // Update module position in Dashboard
-            $query = "SELECT id_hook FROM "._DB_PREFIX_."hook WHERE name = '" . pSQL($hookName) . "'";
+            $query = 'SELECT id_hook FROM ' . _DB_PREFIX_ . "hook WHERE name = '" . pSQL($hookName) . "'";
             $result = Db::getInstance()->ExecuteS($query);
             $id_hook = $result['0']['id_hook'];
 
             $this->updatePosition((int) $id_hook, 0);
+
             return true;
         }
+
         return false;
     }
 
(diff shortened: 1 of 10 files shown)


Quetzacoalt91:
@LouiseBonnard, @eternoendless & @ttoine, you may want to review the wording before going further?

classes/UpgradeSelfCheck.php
```diff
@@ -31,6 +31,11 @@
 
 class UpgradeSelfCheck
 {
+    /**
+     * Recommended PHP Version. If below, display a notice.
+     */
+    const RECOMMENDED_PHP_VERSION = 70000;
```
PierreRambaud:
```suggestion
    const RECOMMENDED_PHP_VERSION = 71000;
```
Php 7.0 is already outdated :/ (http://php.net/supported-versions.php)

classes/UpgradeSelfCheck.php
```diff
@@ -269,6 +361,33 @@ private function checkModuleVersion()
...
 
+    /**
+     * Check current PHP version is supported.
+     * Returns an URL if unmaintained.
+     * 
+     * @return bool|string
+     */
+    private function checkPhpVersionNeedsUpgrade()
+    {
+        if (PHP_VERSION_ID >= self::RECOMMENDED_PHP_VERSION) {
```
PierreRambaud: use http://php.net/version_compare instead

views/templates/hook/dashboard_zone_one.tpl
```diff
@@ -0,0 +1,57 @@
...
+* International Registered Trademark & Property of PrestaShop SA
+*}
+
+<section id="autoupgradePhpWarn" class="panel widget">
+	<div class="panel-heading">
+    <span class="icon-stack text-danger">
+      <span class="icon icon-circle icon-stack-2x"></span>
+      <strong class="icon-stack-1x icon-stack-text">1</strong>
+    </span>
+		{l s='PHP Version notice' mod='autoupgrade'}
```
PierreRambaud: You mix tabs and spaces here

classes/UpgradeSelfCheck.php
Quetzacoalt91: As we deal with integer and not X.X.X strings, is it really useful?

PierreRambaud: Oh, my bad, read `PHP_VERSION` instead of `PHP_VERSION_ID` ><

classes/UpgradeSelfCheck.php
Quetzacoalt91: I know, but only PHP 5.6 could be shortly drop from our compatibility list. I chose to display this message for these people first.

This can be discussed with @PrestaShop/prestashop-core-developers

views/templates/hook/dashboard_zone_one.tpl
Quetzacoalt91: Will be fixed

Quetzacoalt91:
The content to be linked with this feature is under discussion and not ready yet. Closing as it can be shipped with the next version of the module.

Quetzacoalt91:
* whoooosh *

* Article from build to use as link: http://build.prestashop.com/news/announcing-end-of-support-for-obsolete-php-versions/
* New PHP version to recommend : PHP 7.1

autoupgrade.php
```diff
@@ -149,9 +149,58 @@ public function uninstall()
...
+
         return parent::uninstall();
     }
 
+    private function registerHookAndSetToTop($hookName)
+    {
+        if ($this->registerHook($hookName)) {
+            // Update module position in Dashboard
+            $query = 'SELECT id_hook FROM ' . _DB_PREFIX_ . "hook WHERE name = '" . pSQL($hookName) . "'";
+            $result = Db::getInstance()->ExecuteS($query);
```
PierreRambaud: use `getRow` instead? or better `getIdByName`?

Quetzacoalt91: What a good idea. The method is now really simpler

classes/UpgradeSelfCheck.php
```diff
@@ -32,6 +32,11 @@
 
 class UpgradeSelfCheck
 {
+    /**
+     * Recommended PHP Version. If below, display a notice.
+     */
+    const RECOMMENDED_PHP_VERSION = 70100;
```
PierreRambaud:
```suggestion
    const RECOMMENDED_PHP_VERSION = 70130;
```
Because of https://github.com/PrestaShop/PrestaShop/pull/13761/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R6

matks:
Shall we introduce something similar in PS installer ?

classes/UpgradeSelfCheck.php
Quetzacoalt91: So you mean PHP 7.1.30?

Quetzacoalt91:
That was indeed suggested.

classes/UpgradeSelfCheck.php
PierreRambaud: Yep

PierreRambaud: My bad, I mean `7.1.3`, was confused because 7.1.30 is the latest version of 7.1 :roll_eyes:

upgrade/install-4-7-0.php
```diff
@@ -0,0 +1,30 @@
...
+* versions in the future. If you wish to customize PrestaShop for your
+* needs please refer to http://www.prestashop.com for more information.
+*
+*  @author PrestaShop SA <contact@prestashop.com>
+*  @copyright  2007-2016 PrestaShop SA
+*  @license    http://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
+*  International Registered Trademark & Property of PrestaShop SA
+*/
+
+function upgrade_module_4_7_0($module)
```
Quetzacoalt91: Must be now for 4.9.0

classes/UpgradeSelfCheck.php
```diff
@@ -226,11 +300,38 @@ public function getConfigDir()
...
+    {
+        if (1 === (int) Configuration::get('PS_AUTOUP_IGNORE_PHP_UPGRADE')) {
+            return false;
+        }
+
+        if (null !== $this->phpUpgradeNoticelink) {
+            return $this->phpUpgradeNoticelink;
+        }
+
+        return $this->phpUpgradeNoticelink = $this->checkPhpVersionNeedsUpgrade();
```
jolelievre: seems like `phpUpgradeNoticelink` can only be a boolean not a string, you should update its type doc And rename it to avoid confusion

Quetzacoalt91: That's true, I forgot to update the property. Thanks

autoupgrade.php
```diff
@@ -149,9 +149,54 @@ public function uninstall()
...
+    }
+
+    public function hookDashboardZoneOne($params)
+    {
+        // Display panel if PHP is not supported by the community
+        require_once __DIR__ . '/vendor/autoload.php';
+
+        $upgradeContainer = new \PrestaShop\Module\AutoUpgrade\UpgradeContainer(_PS_ROOT_DIR_, _PS_ADMIN_DIR_);
+        $upgrader = $upgradeContainer->getUpgrader();
+        $upgradeSelfCheck = new  \PrestaShop\Module\AutoUpgrade\UpgradeSelfCheck(
```
PierreRambaud:
```suggestion
        $upgradeSelfCheck = new \PrestaShop\Module\AutoUpgrade\UpgradeSelfCheck(
```
:trollface:

autoupgrade.php
```diff
@@ -149,9 +149,54 @@ public function uninstall()
...
+            $upgrader,
+            _PS_ROOT_DIR_,
+            _PS_ADMIN_DIR_,
+            __DIR__
+        );
+
+        $upgradeNotice = $upgradeSelfCheck->isPhpUpgradeRequired();
+        if (false === $upgradeNotice) {
+            return false;
```
PierreRambaud: return empty string instead?

jolelievre: I hesitated on this one as well, but it would seem more logical indeed

autoupgrade.php
Quetzacoalt91: That should be reported by PHP CS Fixer, not you. I'll have a look at the config

PierreRambaud: I AM THE CS FIXER!

PR #288, opened 2019-03-06
Do not reset theme anymore when already active (prevent hooks to be reset)

Quetzacoalt91:
Fixes https://github.com/PrestaShop/PrestaShop/issues/11873


kpodemski:
hey @Quetzacoalt91 

even tho this is fixed we're still having problems with theme being "resetted", could you add an option to skip theme resetting during update?

PR #289, opened 2019-03-08
Remove PclZip from the project

Quetzacoalt91:
As PrestaShop 1.7 requires the zip extension, we can do the same and avoid using a lib which wasn't properly integrated to the project.

Preliminary PR for #279


classes/ZipAction.php
```diff
@@ -175,48 +122,31 @@ private function compressWithZipArchive(&$filesList, $toFile)
...
-        }
-
-        $added_to_zip = $zip->add($files_to_add, PCLZIP_OPT_REMOVE_PATH, $this->prodRootDir);
-        $zip->privCloseFd();
-        if (!$added_to_zip) {
-            (new Filesystem())->remove($toFile);
-            $this->logger->error($this->translator->trans('[ERROR] Error on backup using PclZip: %s.', array($zip->errorInfo(true)), 'Modules.Autoupgrade.Admin'));
-
-            return false;
+            chmod($to_dir, 0775);
```
PierreRambaud: Why forcing 775? 🤔

Quetzacoalt91: As I'm not the original author of this line, I guess this is for restricting access to all users (mkdir is 777 and there is no umask in the project).

PierreRambaud: I think you can remove all chmod like in this module since we remove them in the Core. Let apache / nginx user define rules :)

Quetzacoalt91: Okay, but in a dedicated PR to let the QA make a proper review. I found other occurrences of `chmod` in the code.

Quetzacoalt91: Created issue #291

PR #300, opened 2019-04-24
Improve report of errors and fix notices

Quetzacoalt91:
PR content:

* When an ajax requests fails the security checks, we set its code as 401.
* Better retrieval of the `dir` parameter
* When extracting files from a zip file, report the concerned file in case of failure
* Split error messages when upgrading languages

### DIFF
--- a/ajax-upgradetab.php
+++ b/ajax-upgradetab.php
@@ -44,6 +44,7 @@ if (!$container->getCookie()->check($_COOKIE)) {
         ob_clean();
     }
     echo '{wrong token}';
+    http_response_code(401);
     die(1);
 }
 
--- a/classes/ZipAction.php
+++ b/classes/ZipAction.php
@@ -151,11 +151,19 @@ class ZipAction
             return false;
         }
 
-        if (!$zip->extractTo($to_dir)) {
-            $zip->close();
-            $this->logger->error($this->translator->trans('zip->extractTo(): unable to use %s as extract destination.', array($to_dir), 'Modules.Autoupgrade.Admin'));
+        for($i = 0; $i < $zip->numFiles; $i++) {
+            if (!$zip->extractTo($to_dir, array($zip->getNameIndex($i)))) {
+                $this->logger->error(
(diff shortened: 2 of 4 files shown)


classes/ZipAction.php
```diff
@@ -151,11 +151,19 @@ public function extract($from_file, $to_dir)
             return false;
         }
 
-        if (!$zip->extractTo($to_dir)) {
-            $zip->close();
-            $this->logger->error($this->translator->trans('zip->extractTo(): unable to use %s as extract destination.', array($to_dir), 'Modules.Autoupgrade.Admin'));
+        for($i = 0; $i < $zip->numFiles; $i++) {
```
PierreRambaud:
```suggestion
        for ($i = 0; $i < $zip->numFiles; $i++) {
```

ajax-upgradetab.php
```diff
@@ -44,6 +44,7 @@
         ob_clean();
     }
     echo '{wrong token}';
+    http_response_code(401);
```
PierreRambaud: since php 5.4.
maybe prefer?
```
header('HTTP/1.1 401 Unauthorized');
```

classes/ZipAction.php
Quetzacoalt91: Will be automatically fixed with #303. I won't fix in this issue to avoid triggering Travis

ajax-upgradetab.php
Quetzacoalt91: The module is compatible with PHP 5.6 minimum, so we're fine.

However, I have no idea which solution is better.

Quetzacoalt91: It looks they basically have the same result. HTTP 401 will be properly set by PHP: https://github.com/php/php-src/blob/fc42ac2d78e28134bc46ef58c289586a0907dd79/main/http_status_codes.h#L46

PierreRambaud: Oh my bad, thinking it was 5.2+ because of 1.6 compliance ^^

PR #306, opened 2019-05-03
Handle folder var/cache in module

Quetzacoalt91:
Allow var/cache:
* to be cleaned during upgrade process
* to be ignored when creating backups

Also, this PR moves the cache cleanup between the steps "Files upgrade" & "DB upgrade", right before the core initialization.


Quetzacoalt91:
Approved via #308

PR #308, opened 2019-05-03
Avoid conflit with Psr\Log\LoggerInterface when loading core

Quetzacoalt91:
Fixes `[INTERNAL] /var/www/html/var/cache/prod/classes.php line 9103 - Cannot redeclare class Psr\Log\LoggerInterface` while upgrading to PS 1.7.6.0

### DIFF
--- a/classes/Log/LoggerInterface.php
+++ b/classes/Log/LoggerInterface.php
@@ -1,0 +1,111 @@
+<?php
+/**
+ * IMPORTANT NOTE.
+ *
+ * This class is taken from the composer dependancy psr/log.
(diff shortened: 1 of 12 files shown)


classes/Log/LoggerInterface.php
```diff
@@ -0,0 +1,111 @@
+<?php
+/**
+ * IMPORTANT NOTE.
+ *
+ * This class is taken from the composer dependancy psr/log.
```
tomlev: dependency

Quetzacoalt91:
Thank @tomlev for review.

Issue #196, opened 2018-12-06
Add an array of PHP versions <> PrestaShop


Quetzacoalt91:
In order to prevent compatibility issues, we should add an array in the module specifying the PHP versions compatible with PrestaShop.

If the current env is incompatible with the destination version, the upgade must be blocked.

PR #312, opened 2019-05-08
Add check for compatibility between PHP & PS

Quetzacoalt91:
Checking data given in https://devdocs.prestashop.com/1.7/basics/installation/system-requirements/

Fixes partially https://github.com/PrestaShop/autoupgrade/issues/196

Capture d’écran du 2019-05-08 17-06-48


classes/Requirement/PhpCompatibility.php
```diff
@@ -0,0 +1,87 @@
...
+            return version_compare($phpVersion, $compatibilityData['php']['min'], '>=')
+                && version_compare($phpVersion, $compatibilityData['php']['max'], '<=');
+        }
+
+        // If not given in array, guess it:
+        // - Allow older versions of PS running with older versions of PHP
+        // - Allow newer versions of PS running with newer versions of PHP
+        $data = self::PRESTASHOP_PHP_COMPATIBILITY;
+        $firstCompatibilityGiven = reset($data);
+        if (version_compare($prestashopVersion, $firstCompatibilityGiven['ps']['min'], '<=')
```
PierreRambaud: We could remove the 'min' in the first row and the 'max' in the last row. And if there are no min / no max, use these conditions

PR #315, opened 2019-05-15
Add feature for classifying new modules to be installed when upgrading

Quetzacoalt91:
Allow the the user to know what are the modules included with the PrestaShop version he upgrades to.

They are classified in separate lists, which show which one is new or deleted.

However, this feature seems unstable. It is impossible to advise to install the "new" modules, as the list may have paid modules (pspixel) or non-native modules (ps_googleanalytics).


Matt75:
Does it should use circuit breaker instead of call guzzle directly ?

Quetzacoalt91:
Yeah that's a valid question if that implementation was valid, but currently I'm not even sure we'll keep it. My main concern is finding the list of preinstalled module on both PS version.

PR #325, opened 2019-06-27
Prepend and optimize autoloader

Quetzacoalt91:
Disabling the option `prepend-autoloader` will allow the PrestaShop autoloader to go first. Although it seems dangerous as the core is unstable during the upgrade, it seems to fix some issues like https://github.com/PrestaShop/PrestaShop/issues/13633

:warning: Run `composer install` after applying this PR

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,9 @@
     "preferred-install": "dist",
     "platform": {
         "php": "5.6"
-    }
+    },
+    "prepend-autoloader": false,
+    "optimize-autoloader": true
   },
   "type": "prestashop-module",
   "autoload": {


jolelievre:
> Disabling the option `prepend-autoloader` will allow the PrestaShop autoloader to go first. Although it seems dangerous as the core is unstable during the upgrade, it seems to fix some issues like [PrestaShop/PrestaShop#13633](https://github.com/PrestaShop/PrestaShop/issues/13633)

Anyway didn't we decide that all modules should have prepend false to avoid overriding the core dependencies? (Although I get that upgrade is a more complicated state)

Quetzacoalt91:
> Anyway didn't we decide that all modules should have prepend false to avoid overriding the core dependencies? (Although I get that upgrade is a more complicated state)

Exactly! There's also a PR opened on the validator for that matter.

PR #326, opened 2019-07-02
Add PHPDoc & log when module switch the current theme

Quetzacoalt91:


classes/Upgrader.php
```diff
@@ -351,7 +351,7 @@ public function getXmlMd5File($version, $refresh = false)
      * returns an array of files which are present in PrestaShop version $version and has been modified
      * in the current filesystem.
      *
-     * @return array of string> array of filepath
+     * @return array|false
```
Quetzacoalt91:
```suggestion
     * @return string[]|false
```

PierreRambaud: So we must prefer this syntax when we know it's only composed of strings?

Quetzacoalt91: If we confirm that our analysis tools are able to understand these types, I'm in favor of using them. :)

Quetzacoalt91:
Thank you for review!

PR #327, opened 2019-07-02
Update logs around zip upgrade

Quetzacoalt91:
This PR removes the log "Extract using ZipArchive", which is now always true, and add details the path to the zipfile being modified to the logs. This provides additional details to the user about how the module works.

image


PierreRambaud:
QA needed?

Quetzacoalt91:
Thanks for review

PR #330, opened 2019-07-08
Return in php upgrade script when an error occurs

Quetzacoalt91:
When an error occurs while running a PHP upgrade method, we should return instead of running the return analysis. If we don't, we get an error then a success message. :man_shrugging:


PierreRambaud (approved):

PierreRambaud:
Thanks @Quetzacoalt91

PR #332, opened 2019-07-08
Reverse arrays and use array_pop instead of array_shift

Quetzacoalt91:
In some configuration (especially on PHP 5.6), calling `array_shift` reduce significantly the server performance. This PR removes this method when possible and replaces it with `array_reverse` / `array_pop`.

image


Progi1984 (approved):

PierreRambaud (approved):

PierreRambaud:
@Quetzacoalt91 Does it need QA?

Quetzacoalt91:
Yes, there is a high risk of issues with these changes. I already made some tests before creating this PR but I'd prefer the QA having a look at this as well. :)

Progi1984:
Thanks @Quetzacoalt91 & @sarahdib

PR #333, opened 2019-07-15
Rework module version check

Quetzacoalt91:
It appears that people used an old version of the module before upgrading their shop to 1.7.6, leading to errors during the upgrade.

We should use the information provided by the marketplace to request the user to update their module. An API endpoint allows us to have the information: `https://api.addons.prestashop.com/?method=search&query=5496&search_type=full&version=<shop_version>`

## 1.7.5.2
https://api.addons.prestashop.com/?method=search&query=5496&search_type=full&version=1.7.5.2

## 1.5.0.0
https://api.addons.prestashop.com/?method=search&query=5496&search_type=full&version=1.5.0.0


PierreRambaud:
Is it ready for review?

Quetzacoalt91:
Yep, I have to check a conflict though

classes/Client/ModuleDetailsClient.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        $circuitBreakerFactory = new SimpleCircuitBreakerFactory();
+        $circuitBreaker = $circuitBreakerFactory->create(new FactorySettings(2, 0.1, 10));
+
+        $fallbackResponse = function () {
+            return '{"results":[{"version":"0.0.0.0"}]}';
+        };
+
+        return json_decode(
+            $circuitBreaker->call('https://api.addons.prestashop.com/?method=search&query=5496&search_type=full&version=' . $this->psVersion, [], $fallbackResponse)
```
PierreRambaud: Can we store this string in a config file or at least in a const with a little comment? (What it's requesting)

classes/Client/ModuleDetailsClient.php
```diff
@@ -0,0 +1,100 @@
...
+        $response = $this->call();
+        if (!isset($response->results[0])) {
+            throw new \Exception('Module details are missing from response.');
+        }
+        $this->moduleDetails = $response->results[0];
+    }
+
+    public function call()
+    {
+        $circuitBreakerFactory = new SimpleCircuitBreakerFactory();
```
PierreRambaud: Do not create a new instance of this factory each time you call `call`

classes/Client/ModuleDetailsClient.php
PierreRambaud: something like
```php
/**
 * Search for something special
 */
const ADDONS_SEARCH_SOMETHING = 'https://api.addons.prestashop.com/?method=search&query=5496&search_type=full&version=%s';
```

Quetzacoalt91: Done!

classes/Client/ModuleDetailsClient.php
```diff
@@ -0,0 +1,113 @@
...
+        if (!isset($response->results[0])) {
+            throw new \Exception('Module details are missing from response.');
+        }
+        $this->moduleDetails = $response->results[0];
+    }
+
+    public function call()
+    {
+        $fallbackResponse = function () {
+            return '{"results":[{"version":"0.0.0.0"}]}';
```
PierreRambaud: Should we really want to say yes if request failed?

classes/Client/ModuleDetailsClient.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Requests the API and loads the data in memory
+     */
+    public function requestDetails()
+    {
+        $response = $this->call();
+        if (!isset($response->results[0])) {
+            throw new \Exception('Module details are missing from response.');
```
PierreRambaud:
```suggestion
            throw new Exception('Module details are missing from response.');
```

tests/ModuleDetailsClientTest.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public function testModuleVersionForPS1752()
+    {
+        $psVersion = '1.7.5.2';
+
+        $this->assertSame('4.9.0', $this->getAvailableModuleVersion($psVersion));
+    }
+
+    protected function getAvailableModuleVersion($psVersion)
```
PierreRambaud: Can you add tests when request failed and returns nothing and when the callback is called?

classes/Client/ModuleDetailsClient.php
Quetzacoalt91: I would rather say yes, indeed, as it prevents upgrade to be run otherwise. If the marketplace API is down, the upgrade should still be possible.

classes/Client/ModuleDetailsClient.php
Quetzacoalt91: Doing this will look for a class `PrestaShop\Module\AutoUpgrade\Client\Exception`

PierreRambaud: With a use statement of course :)

tests/ModuleDetailsClientTest.php
Quetzacoalt91: To do this properly, I need to mock the CircuitBreaker instead.

PR #340, opened 2019-09-12
Add feature to disable modules

jolelievre:
This PR fixes the issue https://github.com/PrestaShop/PrestaShop/issues/14779

The module can now disable custom modules by moving them into another folder
This feature was added inside the CoreUpgrader (bound to an option "Disable non native modules"
And a new controller was aded so that this new feature can be used via an API


classes/Module/ModuleRepository.php
```diff
@@ -0,0 +1,152 @@
...
+     * not native). The array returned contains the technical names of the modules.
+     *
+     * @param array $versions
+     *
+     * @return array
+     */
+    public function getCustomModulesOnDisk($versions)
+    {
+        $modulesOnDisk = $this->getModulesOnDisk();
+        $nativeModules = array();
```
PierreRambaud: Don't mix short array and `array`

classes/Module/ModuleRepository.php
```diff
@@ -0,0 +1,152 @@
...
+     * @param string $modulesDirectory
+     *
+     * @return array
+     */
+    private function listModulesFromDirectory($modulesDirectory)
+    {
+        $modulesInDirectory = array();
+        $modules = scandir($modulesDirectory);
+        foreach ($modules as $name) {
+            if (!in_array($name, array('.', '..', 'index.php', '.htaccess')) && @is_dir($modulesDirectory . $name) && @file_exists($modulesDirectory . $name . DIRECTORY_SEPARATOR . $name . '.php')) {
```
PierreRambaud: Could we
```suggestion
            if (in_array($name, array('.', '..', 'index.php', '.htaccess')) && @is_dir($modulesDirectory . $name) || !file_exists($modulesDirectory . $name . DIRECTORY_SEPARATOR . $name . '.php')) {
                continue;
            }
```

classes/Module/AddonsCurlClient.php
```diff
@@ -0,0 +1,159 @@
...
+ */
+
+namespace PrestaShop\Module\AutoUpgrade\Module;
+
+use PrestaShop\Module\AutoUpgrade\Tools14;
+
+/**
+ * Class AddonsCurlClient is a simple Addons client that uses Curl to perform its request.
+ */
+class AddonsCurlClient implements AddonsClientInterface
```
PierreRambaud: Is this class a copy from the Core?

classes/Module/ModuleDisabler.php
```diff
@@ -0,0 +1,156 @@
...
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * Class ModuleDisabler is used to disable/enable modules, there are two ways of
+ * disabling a module:
+ *  - move its whole folder somewhere else
+ *  - update the database to set its active column to 0
+ */
+class ModuleDisabler
```
PierreRambaud: We are already in the Module namespace, it's redundant.
```suggestion
class Disabler
```

classes/Module/AddonsCurlClient.php
```diff
@@ -0,0 +1,159 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to http://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\Module\AutoUpgrade\Module;
```
PierreRambaud: Should be in:
```suggestion
namespace PrestaShop\Module\AutoUpgrade\Addons;
```

classes/Module/AddonsClientInterface.php
```diff
@@ -0,0 +1,54 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\Module\AutoUpgrade\Module;
+
+/**
+ * Interface AddonsClientInterface is used to perform a request on the Addons API.
+ */
+interface AddonsClientInterface
```
PierreRambaud:
```suggestion
interface ClientInterface
```

classes/Module/AddonsClientInterface.php
```diff
@@ -0,0 +1,54 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to http://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\Module\AutoUpgrade\Module;
```
PierreRambaud:
```suggestion
namespace PrestaShop\Module\AutoUpgrade\Addons;
```

classes/AutoupgradeException.php
```diff
@@ -0,0 +1,33 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0  Academic Free License (AFL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\Module\AutoUpgrade;
```
PierreRambaud:
```suggestion
namespace PrestaShop\Module\AutoUpgrade\Exception;
```

classes/Module/AddonsCurlClient.php
```diff
@@ -0,0 +1,159 @@
...
+        $context = stream_context_create(array(
+            'http' => array(
+                'method' => 'POST',
+                'content' => $post_data,
+                'header' => 'Content-type: application/x-www-form-urlencoded',
+                'timeout' => 5,
+            ),
+        ));
+
+        if ($content = Tools14::file_get_contents('https://' . $end_point, false, $context)) {
```
PierreRambaud: Only http and https change here, or I miss something? :thinking:

classes/Module/ModuleRepository.php
```diff
@@ -0,0 +1,152 @@
...
+     * @param string $modulesDir
+     * @param string $disabledModulesDir
+     * @param AddonsClientInterface $addonsClient
+     */
+    public function __construct(
+        $modulesDir,
+        $disabledModulesDir,
+        AddonsClientInterface $addonsClient
+    ) {
+        $this->modulesDir = rtrim($modulesDir, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR;
```
PierreRambaud: Better to not use `DIRECTORY_SEPARATOR` constant in PHP.
If you use DIRECTORY_SEPARATOR, which resolves to \ on windows, things tend to break. For example:\ gets converted to %5C in urls.

classes/State.php
```diff
@@ -232,6 +241,18 @@ public function setInstallVersion($install_version)
...
     }
 
+    /**
+     * @param string $originVersion
+     *
+     * @return self
+     */
+    public function setOriginVersion($originVersion)
+    {
+        $this->originVersion = $originVersion;
```
PierreRambaud:
```suggestion
        $this->originVersion = (string) $originVersion;
```

classes/UpgradeContainer.php
```diff
@@ -289,8 +289,10 @@ public function getUpgrader()
             define('_PS_ROOT_DIR_', $this->getProperty(self::PS_ROOT_PATH));
         }
         // in order to not use Tools class
-        $upgrader = new Upgrader($this->getProperty(self::PS_VERSION));
-        preg_match('#([0-9]+\.[0-9]+)(?:\.[0-9]+){1,2}#', $this->getProperty(self::PS_VERSION), $matches);
+        $currentVersion = $this->getProperty(self::PS_VERSION);
+        $upgrader = new Upgrader($currentVersion);
+        $this->getState()->setOriginVersion($currentVersion);
+        preg_match('#([0-9]+\.[0-9]+)(?:\.[0-9]+){1,2}#', $currentVersion, $matches);
```
PierreRambaud: We maybe must check if $matches is not empty. something broken can came from the api :/

classes/UpgradeTools/ModuleAdapter.php
```diff
@@ -80,17 +108,32 @@ public function getModuleDataUpdater()
...
-     * @param string $pathToUpgradeScripts Path to the PHP Upgrade scripts
      */
-    public function disableNonNativeModules($pathToUpgradeScripts)
+    public function disableNonNativeModules()
     {
-        require_once $pathToUpgradeScripts . 'deactivate_custom_modules.php';
-        deactivate_custom_modules();
+        $customModules = $this->moduleRepository->getCustomModulesOnDisk([$this->originVersion, $this->upgradeVersion]);
+        foreach ($customModules as $moduleName) {
+            $this->moduleDisabler->disableModuleFromDatabase($moduleName);
```
PierreRambaud: Since this class is not final, always prefer getter function
```suggestion
            $this->getModuleDisabler()->disableModuleFromDatabase($moduleName);
```

tests/ModuleRepositoryTest.php
```diff
@@ -0,0 +1,198 @@
...
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->fileSystem = new Filesystem();
+        $this->tempDir = sys_get_temp_dir() . '/module_disabler';
+        $this->tempModulesDir = $this->tempDir . '/modules';
+        $this->tempDisabledModulesDir = $this->tempDir . '/disabled_modules';
+    }
+
+    private function cleanModules()
```
PierreRambaud: could be in setUp & teardown?

upgrade/upgrade-5-0-0.php
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0  Academic Free License (AFL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+function upgrade_module_5_0_0($module)
+{
+    //Install module controller (hidden)
+    $controller_id_tab = (int) Tab::getIdFromClassName('AutoupgradeModule');
```
PierreRambaud: Could we prevent duplicate code from auotupgrade.php and add this in a function?

classes/Module/AddonsClientInterface.php
jolelievre: The thing is not all the classes in this namespace are about Addons
The ModuleDisabler is about modules in PrestaShop, doesn't matter if they come from Addons or not, so the namespace `Module` seemed more appropriate

classes/Module/AddonsClientInterface.php
jolelievre: If the namespace was `Addons` I would agree, but I prefer `Module` (see previous comment)

classes/Module/AddonsCurlClient.php
jolelievre: Not an exact copy, but the code comes from the core yes, but I prefer to have a dedicated class to manage this call inside the module. Since we don't know on which version the module is installed we can't rely on the core classes because they are modified from a version to another.
I think that's why the `Tools` class was copied in `Tools14`

jolelievre: By the way I think I copied this code from the `Tools` class but I can't remember which version (since I was working on different versions at the moment). I also removed a few switch cases because the API simply didn't work with these parameters.

classes/Module/AddonsCurlClient.php
jolelievre: Yes, it seems to be a fallback (I copied the code stupidly). I can use a loop like in `ps_mbo` if you prefer:

```php
        foreach ($protocols as $protocol) {
            if ($content = Tools::file_get_contents($protocol.'://'.$end_point, false, $context)) {
                return $content;
            }
        }
```

classes/Module/ModuleDisabler.php
jolelievre: Ok, but then we keep the `Module` namespace? not Addons?

classes/Module/AddonsClientInterface.php
jolelievre: Oooh, I just got it. Maybe you suggest adding another `Addons` namespace, in addition to the `Module` namespace?

classes/Module/ModuleRepository.php
jolelievre: Right! I'm used to short array now, but in this module I think use the `array()` notation since we're not sure of the PHP version? wdyt?

classes/Module/ModuleRepository.php
jolelievre: Yep, better

classes/UpgradeContainer.php
jolelievre: In this case since the version comes from `UpgraderContainer::getProperty` which uses either constants or files content, so we should be safe.
`PrestashopConfiguration::getPrestaShopVersion` already throws an exception if it can't find the version

classes/UpgradeTools/ModuleAdapter.php
jolelievre: In case it's overridden?

tests/ModuleRepositoryTest.php
jolelievre: Probably, I copied this from `ModuleDisablerTest` where I couldn't because there's a dependency between two tests, but in this case I think I can

upgrade/upgrade-5-0-0.php
jolelievre: I bothers me as well, but then we need to make this method public in `autoupgrade.php` which I'm not a big fan either..

classes/Module/ModuleRepository.php
PierreRambaud: Short array came with 5.4 and this module require 5.6+.

classes/Module/ModuleDisabler.php
PierreRambaud: Depends on the context.
If this class has for aim to disable a module, keep the `Module` namespace

classes/Module/AddonsCurlClient.php
PierreRambaud: I was only asking because of snake case and some parts about `services` and `hosted_module` :sweat_smile:

classes/Module/AddonsClientInterface.php
PierreRambaud: Yep, One for `Addons` and one for `Module`  :sweat_smile:

classes/UpgradeTools/ModuleAdapter.php
PierreRambaud: Yes, even for us we we have another Disabler based on this class. If we have setter / getter, always prefer them.

upgrade/upgrade-5-0-0.php
PierreRambaud: As you wish my lord :pray:

PierreRambaud:
ping @Quetzacoalt91 Can you also have a check on it?

classes/Addons/CurlClient.php
```diff
@@ -0,0 +1,159 @@
...
+                $postData .= '&method=listing&action=native';
+
+                break;
+
+            case 'service':
+                $postData .= '&method=listing&action=service';
+
+                break;
+
+                break;
```
Progi1984, line 87: Why 2 break ?

classes/Addons/CurlClient.php
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * @param string $requestContent
+     *
+     * @return array|bool
+     */
+    private function parseRequest($requestContent)
+    {
+        if (empty($requestContent)) {
+            false;
```
Progi1984, line 144:
```suggestion
            return false;
```

classes/Exception/AutoupgradeException.php
```diff
@@ -0,0 +1,33 @@
+<?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
```
Progi1984, line 3: Please update headers

classes/State.php
```diff
@@ -232,6 +241,18 @@ public function setInstallVersion($install_version)
         return $this;
     }
 
+    /**
+     * @param string $originVersion
+     *
+     * @return self
+     */
+    public function setOriginVersion($originVersion)
```
Progi1984, line 249:
```suggestion
    public function setOriginVersion(string $originVersion)
```

classes/UpgradeContainer.php
```diff
@@ -163,7 +163,7 @@ class UpgradeContainer
 
     public function __construct($psRootDir, $adminDir, $moduleSubDir = 'autoupgrade')
     {
-        $this->autoupgradeWorkDir = $adminDir . DIRECTORY_SEPARATOR . $moduleSubDir;
+        $this->autoupgradeWorkDir = $adminDir . '/' . $moduleSubDir;
```
Progi1984, line 166: Why changing this ?

PierreRambaud: There are many articles and benchmark saying using DIRECTORY_SEPARATOR is useless, it's an old const for very old windows system.

Progi1984: Thanks for the answer. I didn't know ;)

Progi1984:
Ping @jolelievre

classes/State.php
PierreRambaud: We can't do this kind of changes in the autoupgrade module, it must be compatible with older PS versions :)

views/templates/main.twig
```diff
@@ -20,6 +20,8 @@
             {{ 'Double-check the integrity of your backup and that you can easily manually roll-back if necessary.'|trans }}<br>
             {{ 'If you do not know how to proceed, ask your hosting provider.'|trans }}
         </p>
+        <a target="_blank" href="{{ link.getAdminLink('AutoupgradeModule') }}&version=1.6.1">List modules 1.6.1</a>
+        <a target="_blank" href="{{ link.getAdminLink('AutoupgradeModule') }}&version=1.7.6">List modules 1.7.6</a>
```
PierreRambaud, line 24:
```suggestion
        <a target="_blank" href="{{ link.getAdminLink('AutoupgradeModule') }}&amp;version=1.6.1">List modules 1.6.1</a>
        <a target="_blank" href="{{ link.getAdminLink('AutoupgradeModule') }}&amp;version=1.7.6">List modules 1.7.6</a>
```

tests/ModuleRepositoryTest.php
```diff
@@ -0,0 +1,206 @@
...
+    private $tempDisabledModulesDir;
+
+    /** @var Filesystem */
+    private $fileSystem;
+
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->fileSystem = new Filesystem();
+        $this->tempDir = sys_get_temp_dir() . '/module_disabler';
```
PierreRambaud, line 49: Just notice you should use a test directory, instead of the system directory, some container doesn't have /tmp :sweat_smile:

matks:
Hi @jolelievre 

Since we had no news from you for more than 30 days (and some more), I will follow [our guidelines](https://devdocs.prestashop.com/1.7/project/maintainers-guide/reviewing-pull-requests/#stale-pull-requests) and I'll close this pull request as it seems stale. Feel free to reopen or open another one if you think it's still relevant 😉 

Thanks!

PR #341, opened 2019-11-20
SQL upgrade files can use DB_NAME placeholder

jolelievre:
SQL upgrade files can use DB_NAME placeholder

Which will be useful for 1.7.7.0 version (see SQL upgrade file)

Test: to validate this modification you will need to create a build based on this PR https://github.com/PrestaShop/PrestaShop/pull/16465 which uses the new feature (so you can validate 2 PRs at once, amazing!)


PierreRambaud:
Thanks @jolelievre

PR #357, opened 2020-11-25
Fix probably bad directory name

matks:
| Questions     | Answers
| Description?  | Fix bad directory name
| Type?         | improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  |


matks:
@Quetzacoalt91 it looks like a mistake but I have no idea how to verify it was indeed a mistake and that this PR fixes something 😅

Quetzacoalt91 (approved):
wow, you're right

atomiix (approved):

PierreRambaud (approved):

PierreRambaud:
Thanks

PR #359, opened 2020-11-26
Introduce nightly upgrade tests

matks:
| Questions     | Answers
| Description?  | Introduce nightly upgrade tests. See details below.
| Type?         | improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | First part of https://github.com/PrestaShop/PrestaShop/issues/21989
| How to test?  | Travis builds should pass

This PR adds a Travis build that downloads the latest 1.7.7.x ZIP built by nightlies, and perform a CLI autoupgrade using `archive` channel.

To achieve this, I needed to allow the Upgrade Step UpdateConfig to be executed in CLI. This step writes on disk a file `config.var` that is needed by autoupgrade process when channel `archive` is used. In Web context, the UpdateConfig is called before starting the upgrade process.

It has some limitations (that I hope to solve in future PRs):
- requires to specify, as environment variables, the target version (1.7.7.0) and the target branch (1.7.7.x)
- requires to provide a configuration file to be used by UpdateConfig to generate the `config.var` (the content of this configuration file comes, in web context, from a form filled by the user)
- generates the download URL of the ZIP without safety it is valid

In future PRs, I aim to:
- make this process run every night
- create a API endpoint on nightly.prestashop.com that would return the latest available ZIP files and we can use them instead of "guessing" the download URL ; it could also provide the data needed by UpdateConfig


classes/TaskRunner/Miscellaneous/UpdateConfig.php
```diff
@@ -32,61 +32,74 @@
...
+     * @var bool
+     */
+    protected $isRunInCLI = false;
+
+    /**
+     * Data being passed by CLI entry point
+     *
+     * @var array
+     */
+    protected $cliParameters;
```
Progi1984, line 45:
```suggestion
    protected $cliParameters = [];
```

classes/TaskRunner/Miscellaneous/UpdateConfig.php
```diff
@@ -95,6 +108,40 @@ public function run()
         }
     }
 
+    /**
+     * @param array $parameters
+     */
+    public function inputCLIParameters($parameters)
```
Progi1984, line 114:
```suggestion
    public function inputCLIParameters($parameters): array
```

classes/TaskRunner/Miscellaneous/UpdateConfig.php
```diff
@@ -95,6 +108,40 @@ public function run()
...
+    protected function fetchConfigurationData()
+    {
+        if ($this->isRunInCLI) {
+            return $this->getCLIParameters();
+        } else {
+            return $this->getRequestParams();
+        }
+    }
+
+    protected function getCLIParameters()
```
Progi1984, line 132:
```suggestion
    protected function getCLIParameters(): array
```

classes/TaskRunner/Miscellaneous/UpdateConfig.php
```diff
@@ -95,6 +108,40 @@ public function run()
...
+    public function inputCLIParameters($parameters)
+    {
+        $this->isRunInCLI = true;
+        $this->cliParameters = $parameters;
+    }
+
+    /**
+     * Fetch data from either $_REQUEST (web context) or given input (CLI context)
+     */
+    protected function fetchConfigurationData()
```
Progi1984, line 123:
```suggestion
    protected function fetchConfigurationData(): array
```

classes/TaskRunner/Upgrade/UpgradeNow.php
```diff
@@ -47,7 +47,7 @@ public function run()
             $upgrader->checkPSVersion(false, array('minor'));
         }
 
-        if ($upgrader->isLastVersion()) {
+        if (!in_array($channel, ['private','archive','directory']) && $upgrader->isLastVersion()) {
```
Progi1984, line 50:
```suggestion
        if (!in_array($channel, ['private', 'archive', 'directory']) && $upgrader->isLastVersion()) {
```

cli-updateconfig.php
```diff
@@ -0,0 +1,47 @@
...
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ */
```
Progi1984, line 26:
```suggestion
 */
 
declare(strict_types=1);
```

cli-updateconfig.php
```diff
@@ -0,0 +1,47 @@
...
+
+if (PHP_SAPI !== 'cli') {
+    echo 'This script must be called from CLI';
+    exit(1);
+}
+
+$inputConfigurationFile = getopt('', array('from::'))['from'];
+if (!file_exists($inputConfigurationFile)) {
+    echo sprintf('Invalid input configuration file %s', $inputConfigurationFile).PHP_EOL;
```
Progi1984, line 35:
```suggestion
    echo sprintf('Invalid input configuration file %s', $inputConfigurationFile) . PHP_EOL;
```

tests/downloadNightlyZip.sh
```diff
@@ -0,0 +1,19 @@
...
+    exit 1
+fi
+
+if [[ -z "${PS_TARGET_BRANCH}" ]]; then
+    echo "Variable PS_TARGET_BRANCH must be defined"
+    exit 1
+fi
+
+
+YESTERDAY="$(date -d yesterday '+%Y-%m-%d')"
```
Progi1984, line 14: Are you sure a build is done in week end ?

.travis.yml
```diff
@@ -7,25 +7,43 @@ php:
 env:
   - VERSION=1.7.2.2
     CHANNEL=major
+    TEST_TYPE=standard
```
eternoendless, line 10: I think "use_channel" or "from_channel" would be more declarative. If you choose "use_" then you should rename "from_archive" to "use_archive"

Progi1984 (commented):
Need some changes

.travis.yml
```diff
@@ -7,25 +7,43 @@ php:
...
+      bash tests/downloadNightlyZip.sh;
+    fi
   - docker-compose up -d
 
 before_script:
   - bash -c 'while [[ "$(curl -L -s -o /dev/null -w %{http_code} http://localhost:8001/index.php)" != "200" ]]; do sleep 5; done'
   - docker exec -u www-data -ti prestashop_autoupgrade cp modules/autoupgrade/ -R admin-dev
+  - if [ "$TEST_TYPE" == "from_archive" ]; then
+      docker exec -u www-data -ti prestashop_autoupgrade mkdir admin-dev/autoupgrade/download;
+      docker exec -u www-data -ti prestashop_autoupgrade cp modules/autoupgrade/prestashop.zip admin-dev/autoupgrade/download/prestashop.zip;
```
eternoendless, line 44:
```suggestion
      docker exec -u www-data -ti prestashop_autoupgrade mv modules/autoupgrade/prestashop.zip admin-dev/autoupgrade/download/prestashop.zip;
```

tests/downloadNightlyZip.sh
matks: This is one of the current limitations of this PR. In theory a build is computed every night, weekend included. In practice sometimes the build process fails.

A good solution will be
> create a API endpoint on nightly.prestashop.com that would return the latest available ZIP files and we can use them instead of "guessing" the download URL ; it could also provide the data needed by UpdateConfig

But this is another repo, another PR, another project 😅

classes/TaskRunner/Miscellaneous/UpdateConfig.php
matks: I am not sure 🤔  this code must remain php5.6-compliant if I am not wrong. Is this style compliant with php5.6 ?

cli-updateconfig.php
matks: This one I am sure we cannot do it. We have to remain php5.6 compliant because this module must be able to upgrade 1.6 shops.

cli-updateconfig.php
```diff
@@ -0,0 +1,47 @@
...
+    exit(1);
+}
+$inputData = json_decode(file_get_contents($inputConfigurationFile), true);
+
+require_once realpath(dirname(__FILE__) . '/../../modules/autoupgrade') . '/ajax-upgradetabconfig.php';
+$container = autoupgrade_init_container(dirname(__FILE__));
+
+$container->setLogger(new PrestaShop\Module\AutoUpgrade\Log\StreamedLogger());
+$controller = new \PrestaShop\Module\AutoUpgrade\TaskRunner\Miscellaneous\UpdateConfig($container);
```
eternoendless, line 44: Use "use" instead of FQCN

cli-updateconfig.php
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ */
+
```
eternoendless, line 27: Consider explaining usage in a comment. What does this script do? What are its parameters?

cli-updateconfig.php
Quetzacoalt91: Indeed, at the moment this module is compatible from PHP 5.6

classes/TaskRunner/Miscellaneous/UpdateConfig.php
```diff
@@ -95,6 +108,40 @@ public function run()
...
+    {
+        if ($this->isRunInCLI) {
+            return $this->getCLIParameters();
+        } else {
+            return $this->getRequestParams();
+        }
```
eternoendless, line 129: You can remove the useless else after return
```suggestion
        } 
        return $this->getRequestParams();
```

eternoendless, line 129: You can remove the useless else after return
```suggestion
        if ($this->isRunInCLI) {
            return $this->getCLIParameters();
        }
        return $this->getRequestParams();
```

classes/TaskRunner/Miscellaneous/UpdateConfig.php
eternoendless: Can't use PHP 7 features in the autoupgrade module

classes/TaskRunner/Miscellaneous/UpdateConfig.php
eternoendless: Can't use PHP 7 features in the autoupgrade module

classes/TaskRunner/Miscellaneous/UpdateConfig.php
eternoendless: Can't use PHP 7 features in the autoupgrade module

classes/TaskRunner/Miscellaneous/UpdateConfig.php
```diff
@@ -32,61 +32,74 @@
...
 
         // Was coming from AdminSelfUpgrade::currentParams before
-        $request = $this->getRequestParams();
+        $inputConfigurationData = $this->fetchConfigurationData();
+
         $config = array();
         // update channel
-        if (isset($request['channel'])) {
-            $config['channel'] = $request['channel'];
+        if (isset($inputConfigurationData['channel'])) {
```
eternoendless, line 57: Consider using a shorter variable name like `$input` or `$params`

tests/downloadNightlyZip.sh
Progi1984: But have you checked, we use this API every day from Presthubot :

https://api-nightly.prestashop.com/reports

matks: Ooooh ❤️  that is a very good starting point indeed.

I see however this returns very old reports like 1.7.6.x reports so I will need nonetheless to create a 2nd endpoint with just the latest results. But this is great: I should be able to reuse the logic behind this to build it.

cli-updateconfig.php
matks: Fixed in https://github.com/PrestaShop/autoupgrade/pull/359/commits/2aacf592d1a8481f2c032f07cc24c7c819db7f2b

Now `cli-updateconfig.php --help` prints
```
Usage: php cli-updateconfig.php [--from=<configurationFile>]
```

classes/TaskRunner/Miscellaneous/UpdateConfig.php
matks: Fixed in https://github.com/PrestaShop/autoupgrade/pull/359/commits/46d1d8e3942284bbf2b1fe15b7bd15976ab4d3da

cli-updateconfig.php
matks: Fixed in https://github.com/PrestaShop/autoupgrade/pull/359/commits/46d1d8e3942284bbf2b1fe15b7bd15976ab4d3da

.travis.yml
matks: Fixed in https://github.com/PrestaShop/autoupgrade/pull/359/commits/a08ba3e030046b9abaf4c02ff5ca92106d9f175d

tests/downloadNightlyZip.sh
matks: @Progi1984 Parsing the big JSON returned by https://api-nightly.prestashop.com/reports will require a small parser (I will not do this in bash 😆 ). I would like to avoid this PR getting any bigger and implement this in another PR as an improvement.

Progi1984: @matks Use `jq` (https://stedolan.github.io/jq/) and your life will be more easier ;)

Progi1984: (and it's available on Travis : https://travis-ci.com/github/PrestaShop/autoupgrade/jobs/450892848#L52)

Progi1984: Done :

```bash
$ curl -sS https://api-nightly.prestashop.com/reports | jq -r 'first(.[] | select(.version == "1.7.6.x")) | .download'
null
$ curl -sS https://api-nightly.prestashop.com/reports | jq -r 'first(.[] | select(.version == "1.7.7.x")) | .download'
https://storage.googleapis.com/prestashop-core-nightly/2020-12-01-1.7.7.x-prestashop_1.7.7.0.zip
$ curl -sS https://api-nightly.prestashop.com/reports | jq -r 'first(.[] | select(.version == "develop")) | .download'
https://storage.googleapis.com/prestashop-core-nightly/2020-12-01-develop-prestashop_1.7.8.0.zip

```

matks:
I need to use https://futurestud.io/tutorials/github-actions-trigger-builds-on-schedule-cron

PR #360, opened 2020-11-26
Move PHPUnit Tests to GA

matks:
| Questions     | Answers
| Description?  | Move PHPUnit Tests to GitHub Action. Also test on php5.6 and php7.2!
| Type?         | improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? |
| How to test?  | GA should pass

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -1,0 +1,21 @@
+name: PHP tests
+on: [push, pull_request]
+jobs:
+  build:
+    strategy:
+      matrix:
+        php-versions: ['5.6', '7.2']
+    runs-on: ubuntu-latest
+    name: PHP Unit
+    steps:
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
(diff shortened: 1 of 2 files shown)


matks:
The GA pass on my fork https://github.com/matks/autoupgrade/pull/3

.github/workflows/php.yml
```diff
@@ -0,0 +1,21 @@
+name: PHP tests
+on: [push, pull_request]
+jobs:
+  build:
+    strategy:
+      matrix:
+        php-versions: ['5.6', '7.2']
```
atomiix, line 7: Maybe add 7.3 as it should be compatible with PrestaShop 1.7.7.0 that is itself compatible with php 7.3?

matks: I'm going to even see if it works for php7.4. 1.7.8 must be php7.4 compatible.

atomiix (approved):

matks:
Thanks for review!

PR #361, opened 2020-11-26
Test upgrades for PS1.7 with php7.1

matks:
| Questions     | Answers
| Description?  | Test upgrades for PS1.7 with php7.1
| Type?         | improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? |
| How to test?  | Travis test should pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ services: docker
 
 php:
   - '5.6'
+  - '7.1'
 
 env:
   - VERSION=1.7.2.2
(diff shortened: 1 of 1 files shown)


atomiix (approved):

matks:
Thank you for review!

matks:
Just realized this test is probably useless. Because
- phpunit tests will be moved in GitHub Actions
- phpstan tests and integration tests are run in a docker image so the PHP version used by Travis does not matter

PR #369, opened 2020-12-01
Avoid drop of some tables like guests

Quetzacoalt91:
| Questions     | Answers
| Description?  | Fix deletion of some tables by mistake
| Type?         | bug fix
| BC breaks?    | Nope
| Deprecations? | Nope
| Fixed ticket? | /
| How to test?  | Backup, upgrade then restore, the customer data must remain in DB

### DIFF
--- a/classes/TaskRunner/Rollback/RestoreDb.php
+++ b/classes/TaskRunner/Rollback/RestoreDb.php
@@ -141,7 +141,7 @@ class RestoreDb extends AbstractTask
 
                 $tables_after_restore = array_unique($tables_after_restore);
                 $tables_before_restore = $databaseTools->getAllTables();
-                $tablesToRemove = array_diff($tables_before_restore, $tables_after_restore);
+                $tablesToRemove = array_diff($tables_before_restore, $tables_after_restore, $ignore_stats_table);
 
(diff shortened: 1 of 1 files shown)


matks:
2 GitHub Actions fail but I think it's weird items. There is a PHP5.6 GA build successfull. The other must be bad duplicates.

Quetzacoalt91:
This seems to be related to GitHub hitself. 
image

Can you restart the workflow just in case?

Progi1984:
@Quetzacoalt91 I restarted them this morning : same result.

atomiix:
@Quetzacoalt91 isn't there something to do here too? https://github.com/PrestaShop/autoupgrade/blob/master/classes/TaskRunner/Upgrade/BackupDb.php#L110

atomiix:
@Quetzacoalt91 I don't remember why I said that 😅
But I don't think it is useful add the `guest` table to the list of data to save.

atomiix (approved):

Progi1984 (approved):

PierreRambaud (approved):

Progi1984:
Thanks @Quetzacoalt91 & @sarahdib

PR #379, opened 2021-02-08
Unescape special chars on translate chain

NeOMakinG:
| Questions         | Answers
| Description?      | There was encoded html chars on the configuration page
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/23147.
| How to test?      | See issue


NeOMakinG:
@PierreRambaud close it if you think this is not the right way at all

PierreRambaud (changes requested):
As I said you, I don't think it's the right way to fix it because all translations can fail depending on the language you use.
The main problem is because smarty is calling a twig template, and the escape is done twice.

PR #387, opened 2021-03-11
Add Nightly upgrades using User Infterface

boubkerbribri:
| Questions         | Answers
| Description?      | Add cron to test upgrade from "1.7.4.4", "1.7.5.2", "1.7.6.9", "1.7.7.1" to today builds on nightly (1.7.7.2 and 1.7.8.0) with user interface
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | no
| How to test?      | https://github.com/boubkerbribri/autoupgrade/actions/runs/642660002
| Possible impacts? | none


tests/e2e/scenarios/02_upgrade.js
```diff
@@ -0,0 +1,136 @@
...
+  it('should put the shop under maintenance and check if the checklist is all green', async () => {
+    await upgradeModulePage.putShopUnderMaintenance(page);
+
+    for (let i = 1; i <= 10; i++) {
+      const textResult = await upgradeModulePage.getRowImageContent(page, i);
+      await expect(textResult).to.equal('ok');
+    }
+  });
+
+  it('should click on \'UPGRADE PRESTASHOP NOW\' and wait for the upgrade', async () => {
```
atomiix, line 119: I think you should also activate the "Switch to the default theme" when doing a major upgrade (1.6 -> 1.7)

atomiix (commented):
Just one comment but seems good overall 👍

tests/e2e/scenarios/02_upgrade.js
boubkerbribri: For the now, this does not support upgrades from 1.6, because the the prestashop_test_lib does not allow it .

atomiix (dismissed):

.github/workflows/e2e_nightly_upgrade.yml
```diff
@@ -0,0 +1,204 @@
...
+
+      - id: set-target-versions
+        shell: bash
+        run: echo "::set-output name=TARGET_VERSIONS::$(node ./scripts/getTargetVersions.js)"
+        env:
+          API_URL: https://api-nightly.prestashop.com/reports
+        working-directory: ./tests/e2e
+
+    outputs:
+      target_versions: ${{ steps.set-target-versions.outputs.TARGET_VERSIONS }}
```
PierreRambaud, line 35: use env instead https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

.github/workflows/e2e_nightly_upgrade.yml
```diff
@@ -0,0 +1,204 @@
...
+        with:
+          name: failed-steps-for-test-${{ matrix.ps_start_version }}-${{ matrix.ps_target_version.for_test }}
+          path: ./tests/e2e/screenshots/
+
+      # Upload Mochawesome reports
+      - name: Rename Mochawesome report
+        if: ${{ always() }}
+        run: |
+          mkdir -p ./json-reports
+          cp ./tests/e2e/mochawesome-report/mochawesome.json ./json-reports/upgrade-from-${{ matrix.ps_start_version }}-to-${{ matrix.ps_target_version.for_test }}.json
```
PierreRambaud, line 124: A lot of tasks should be in dedicated scripts instead, it will be easier to manage, use and reproduce (as it's done in the PrestaShop Core repository)

.github/workflows/e2e_nightly_upgrade.yml
```diff
@@ -0,0 +1,204 @@
...
+          path: ${{ env.report_path }}
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+
+      - name: Combine reports
+        shell: bash
+        run: ./tests/e2e/scripts/combine-reports.py "${{ env.report_path }}" "${{ env.report_path }}/${{ env.combined_report_name }}"
```
PierreRambaud, line 159: Same about combine task, it's a little bit complicated and can be improved by using dedicated scripts

PierreRambaud (commented):
Why using Dockerfile inside .github directory, I don't get it?
And few comments, I really think GitHub actions have to be easy to read, you should prefer bash scripts instead of multiple yaml tasks.

boubkerbribri:
> Why using Dockerfile inside .github directory, I don't get it?
> And few comments, I really think GitHub actions have to be easy to read, you should prefer bash scripts instead of multiple yaml tasks.

I used the one who already was there to not create a new one, maybe we can ask @atomiix about this.

atomiix:
I've put the Dockerfile in the .github folder because it is not intended to be used for something else that the Github actions, but if you think we should put it somewhere else, I'm ok with that :)

atomiix (approved):

.github/workflows/e2e_nightly_upgrade.yml
boubkerbribri: I did the modification https://github.com/PrestaShop/autoupgrade/pull/387/commits/ef3122bdad9237833bb2bd88f0973e7c1f7795f7 but I got a context error `Unrecognized named-value: 'env'.` So I deleted it.

.github/workflows/e2e_nightly_upgrade.yml
boubkerbribri: done

.github/workflows/e2e_nightly_upgrade.yml
boubkerbribri: done

boubkerbribri:
Looks good now [runs/2121453573](https://github.com/boubkerbribri/autoupgrade/runs/2121453573)

atomiix (dismissed):

.github/workflows/e2e_nightly_upgrade.yml
PierreRambaud: Because it should be only one job, they are link, one can't work without the other.

atomiix (dismissed):

PierreRambaud (approved):

jolelievre (approved):

matks:
THIS IS AMAZING

PR #400, opened 2021-08-24
Added support of Twig 3

Progi1984:
| Questions         | Answers
| Description?      | Fixed support of Twig 3 : Like `develop` is now based on Twig 3 (Thanks @atomiix), autoload load firstly the dependency of the core and after the dependency of the module. It's why I detect if Twig 3 class is here or not.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#25624
| How to test?      | Cf PrestaShop/PrestaShop#25624 : check on PS 1.6/1.7/develop the display of the module
| Possible impacts? | N/A

### DIFF
--- a/classes/Twig/TransFilterExtension3.php
+++ b/classes/Twig/TransFilterExtension3.php
@@ -25,36 +25,38 @@
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  */
 
-namespace PrestaShop\Module\AutoUpgrade\Twig\Block;
+namespace PrestaShop\Module\AutoUpgrade\Twig;
 
-use PrestaShop\Module\AutoUpgrade\BackupFinder;
-use Twig_Environment;
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFilter;
 
-class RollbackForm
+/**
+ * Filter (Support for Twig 3)
+ */
+class TransFilterExtension3 extends AbstractExtension
 {
-    /**
-     * @var Twig_Environment|\Twig\Environment
-     */
-    private $twig;
+    const DOMAIN = 'Modules.Autoupgrade.Admin';
 
     /**
-     * @var BackupFinder
+     * @var Translator
      */
-    private $backupFinder;
+    private $translator;
 
-    public function __construct($twig, BackupFinder $backupFinder)
+    public function __construct($translator)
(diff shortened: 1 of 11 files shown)


classes/Twig/TransFilterExtension3.php
```diff
@@ -0,0 +1,62 @@
...
+namespace PrestaShop\Module\AutoUpgrade\Twig;
+
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFilter;
+
+/**
+ * Filter (Support for Twig 3)
+ */
+class TransFilterExtension3 extends AbstractExtension
```
PierreRambaud, line 37:
```suggestion
class TransFilterExtension extends AbstractExtension
```

Shouldn't be specify in the name of the method

PierreRambaud: There is maybe another way to make the first class compatible to the other twig version

Progi1984: No, because `TransFilterExtension` always exists and is compatible with Twig 1.

PierreRambaud (approved):

classes/UpgradeContainer.php
```diff
@@ -415,19 +416,28 @@ public function getTranslator()
...
         if (null !== $this->twig) {
             return $this->twig;
         }
 
-        // Using independant template engine for 1.6 & 1.7 compatibility
-        $loader = new Twig_Loader_Filesystem();
-        $loader->addPath(realpath(__DIR__ . '/..') . '/views/templates', 'ModuleAutoUpgrade');
-        $twig = new Twig_Environment($loader);
-        $twig->addExtension(new TransFilterExtension($this->getTranslator()));
+        if (class_exists(\Twig\Environment::class)) {
```
atomiix, line 427: I'm not sure about that because the class `\Twig\Environment` also exists in twig 1.x

PierreRambaud: You're right :O 
https://github.com/twigphp/Twig/blob/1.x/src/Environment.php#L1638

Progi1984: @atomiix Thanks. I have fixed with a reverse test.

PierreRambaud (approved):

atomiix (approved):

PierreRambaud:
Thank you @Progi1984

PR #406, opened 2021-09-13
Add second parameter for fopen in Tools14::copy

sowbiba:
| Questions         | Answers
| Description?      | Fopen requires secong parameter for open mode
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| How to test?      | CI green<br>If `allow_url_fopen` is set to false or 0 or Off in php.ini, curl will be used to download the PS binary, otherwise, fopen will be used. <br>To test this PR, use fopen
| Possible impacts? | -

### DIFF
--- a/classes/Tools14.php
+++ b/classes/Tools14.php
@@ -2410,7 +2410,7 @@ AddOutputFilterByType DEFLATE application/x-javascript
         }
 
         if (self::shouldUseFopen($source)) {
-            $sourceFile = fopen($source);
+            $sourceFile = fopen($source, 'rb');
             // If something else than false, the data was stored
             $result = (file_put_contents($destination, $sourceFile) !== false);
             fclose($sourceFile);


matks (approved):
Looks OK but if I understand correctly, using fopen with only 1 parameter is actually not possible. So this was wrong all this time?

sowbiba:
> Looks OK but if I understand correctly, using fopen with only 1 parameter is actually not possible. So this was wrong all this time?

It seems it was wrong since 2018

classes/Tools14.php
```diff
@@ -2410,7 +2410,7 @@ public static function copy($source, $destination, $stream_context = null)
         }
 
         if (self::shouldUseFopen($source)) {
-            $sourceFile = fopen($source);
+            $sourceFile = fopen($source, 'r');
```
PierreRambaud, line 2413:
```suggestion
            $sourceFile = fopen($source, 'rb');
```

The default behavior was in bytes, the destination file is `wb`, seems to be logic to have `rb` instead of `r`

sowbiba: The doc talks about Notepad issues for the reasons to add 'b' mode https://www.php.net/manual/fr/function.fopen.php

But why not :D

Done ;)

Progi1984 (approved):

matks (approved):

PierreRambaud:
No need QA, the CI was failing and CI is green now.

PR #411, opened 2021-09-21
Display the current PHP version

okom3pom:
| Questions         | Answers
| Description?      | Display php Version 
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes  https://github.com/PrestaShop/PrestaShop/issues/25656
| How to test?      | Open module see phpVersion
| Possible impacts? | Please indicate what parts of the software we need to check to make sure everything is alright.

### DIFF
--- a/classes/Twig/Block/UpgradeButtonBlock.php
+++ b/classes/Twig/Block/UpgradeButtonBlock.php
@@ -179,7 +179,7 @@ class UpgradeButtonBlock
             'downloadPath' => $this->downloadPath . DIRECTORY_SEPARATOR,
             'directoryVersionNumber' => $this->config->get('directory.version_num'),
             'manualMode' => $this->manualMode,
-            'phpVersion' => phpversion(),
+            'phpVersion' => PHP_VERSION,
         );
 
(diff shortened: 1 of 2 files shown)


classes/Twig/Block/UpgradeButtonBlock.php
```diff
@@ -179,6 +179,7 @@ public function render()
             'downloadPath' => $this->downloadPath . DIRECTORY_SEPARATOR,
             'directoryVersionNumber' => $this->config->get('directory.version_num'),
             'manualMode' => $this->manualMode,
+            'phpVersion' => phpversion(),
```
PrestaEdit, line 182: To PHP devs, could you say me if the use of phpversion() is better/slower/the same as the defined PHP_VERSION ? Just to know, not a big deal, of course :)

PierreRambaud: It's always faster to use predefined constant instead of calling a function.
```suggestion
            'phpVersion' => PHP_VERSION,
```

PierreRambaud (approved):

Progi1984 (approved):

PierreRambaud:
Thank you @okom3pom

PR #421, opened 2021-10-04
Check if override is disabled before update

okom3pom:
| Questions         | Answers
| Description?      | Check if overrides are disabled and let the user disable them
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#14781
| How to test?      | Change disable override setting and check the module.
| Possible impacts? | Please indicate what parts of the software we need to check to make sure everything is alright.

### DIFF
--- a/AdminSelfUpgrade.php
+++ b/AdminSelfUpgrade.php
@@ -76,6 +76,7 @@ class AdminSelfUpgrade extends AdminController
     public $keepMails;
     public $manualMode;
     public $deactivateCustomModule;
+    public $disableOverride;
 
     public static $classes14 = ['Cache', 'CacheFS', 'CarrierModule', 'Db', 'FrontController', 'Helper', 'ImportModule',
         'MCached', 'Module', 'ModuleGraph', 'ModuleGraphEngine', 'ModuleGrid', 'ModuleGridEngine',
@@ -174,9 +175,8 @@ class AdminSelfUpgrade extends AdminController
             // Create a dummy index.php file in the XML config directory to avoid directory listing
             if (!file_exists(_PS_ROOT_DIR_ . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'xml' . DIRECTORY_SEPARATOR . 'index.php') &&
                 (file_exists(_PS_ROOT_DIR_ . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'index.php') &&
-                 !@copy(_PS_ROOT_DIR_ . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'index.php', _PS_ROOT_DIR_ . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'xml' . DIRECTORY_SEPARATOR . 'index.php'))) {
+                    !@copy(_PS_ROOT_DIR_ . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'index.php', _PS_ROOT_DIR_ . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'xml' . DIRECTORY_SEPARATOR . 'index.php'))) {
                 $this->_errors[] = $this->trans('Unable to create the directory "%s"', [_PS_ROOT_DIR_ . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'xml'], 'Modules.Autoupgrade.Admin');
-
                 return;
             }
         }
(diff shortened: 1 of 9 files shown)


marionf:
Hello @okom3pom 

The [mockups](https://www.figma.com/file/GpRsUUCHcdhJPAaz7Eqzes/Autoupgrade?node-id=497%3A11710) are now up to date and you can add a toggle to let the user choose if he wants to disable or not the overrides ;)

matks:
Hi @matthieu-rolland can you confirm the status of this contribution? WIP and some more commits are coming or do we review it ?

matks:
ping @matthieu-rolland

matthieu-rolland:
> Hi @matthieu-rolland can you confirm the status of this contribution? WIP and some more commits are coming or do we review it ?

made some last fixes, it's ready to review again 👍

atomiix (approved):

AdminSelfUpgrade.php
```diff
@@ -43,7 +44,7 @@ class AdminSelfUpgrade extends AdminController
     // used for translations
     public static $l_cache;
     // retrocompatibility
-    public $noTabLink = array();
+    public $noTabLink = [];
```
jolelievre, line 47: I think you can't use this notation with brackets It's available since PHP 5.4 while 1.6 is compatible with older versions So this will prevent using the autoupgrade module when trying to upgrade from 1.6 to an intermediate PrestaShop version (if you don't want to upgrade to the latest version)

jolelievre (commented):
Sorry I didn't review much I think all the changes regarding array syntax should be reverted first

AdminSelfUpgrade.php
atomiix: Last versions of the autoupgrade module is compatible with php >= 5.6 so I guess we're safe here

Progi1984:
@okom3pom Needs a rebase

matthieu-rolland:
> @okom3pom Needs a rebase

I'm on it 👍

atomiix (dismissed):

AdminSelfUpgrade.php
```diff
@@ -76,6 +77,7 @@ class AdminSelfUpgrade extends AdminController
     public $keepMails;
     public $manualMode;
     public $deactivateCustomModule;
+    public $disableOverride;
```
Progi1984, line 80: Could you add the PHPDoc, please ?

Progi1984 (approved):
A small feedback but it's approved

PierreRambaud (dismissed):

florine2623 (dismissed):
Hello @okom3pom ,

Thanks for the PR !

Sorry for the long answer, I thought I already answered.

1. We should add a toggle "Disable all overrides" in Advanced Parameters > Performance" in the "upgrade options" block. ✅ 
Screenshot 2022-03-15 at 17 54 47

2. When this option is enabled or disabled from the module, the same option in "Advanced Parameters > Performance" is changed accordingly. ✅ 

https://user-images.githubusercontent.com/16019289/158430859-bf33b04e-a05f-4ade-867a-3257596c0119.mov

3. By default, the toggle should be on the same state at the same toggle in "Advanced Parameters > Performance" ✅  

https://user-images.githubusercontent.com/16019289/158431237-67ffa9fe-6667-47e4-bfca-13e5da9fed40.mov

Whether the toggle is modified from the autoupgrade module or "Advanced Parameters > Performance", they should always match.
After a few times enabling/disabling this feature, point 3 is not OK when I have to confirm resubmission of form. I have followed these steps, it won't work :
- Have your autoupgrade page in one tab
- Open your Advanced Parameters > Performance in another tab
- Switch "Disable all overrides" in Advanced Parameters > Performance tab
- Go to your first autoupgrade tab > refresh > NOK ⛔ 

https://user-images.githubusercontent.com/16019289/158433720-9fba7007-1aeb-4743-9397-cf5a699c976a.mov

I'm using Google Chrome, Version 99.0.4844.51

Could you take a look ?
Thanks!

matthieu-rolland:
So at first I couldn't reproduce this behavior, but then I managed to do it:

1/ push the form on the autoupgrade page
2/ see in `Advanced Parameters > Performance` that the value is the same
3/ change the value and push the form in `Advanced Parameters > Performance`
4/ Go bach to the autoupgrade page and refresh => the value didn't change

Since we pushed the autoupgrade form in step 1, when we refresh its page in step 4, the form is pushed again, hence the browser's alert at 10 seconds in the video

matthieu-rolland:
fixed !

atomiix (approved):

florine2623 (approved):
Thanks @matthieu-rolland ,

I've tested again this bug. It is looking good ✅ 

I also tested the others Upgrade Options, they all are well saved and the whole form works great ✅ 

I've made 3 different upgrades :
PS 1.7.6.5 -> PS 1.7.8.5 (Disable all overrides : YES)
PS 1.7.7.8 -> PS 1.7.8.5 (Disable all overrides : NO)
PS 1.7.8.4 -> PS 1.7.8.5 (Disable all overrides : YES)
All 3 upgrades were successful and the Override Option is well saved
I changed the override option and rollbacked all shop to their original version, the override option is the one from the previous version.

It is QA ✅ for me !

Progi1984:
Thanks @okom3pom, @matthieu-rolland  & @florine2623

PR #434, opened 2021-11-02
Use GenerateThemeMailTemplatesCommand during upgrade

jolelievre:
| Questions         | Answers
| Description?      | Use GenerateThemeMailTemplatesCommand during upgrade instead of legacy removed Language::installEmailsLanguagePack function
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/26503
| How to test?      | PrestaShop/PrestaShop#27276 must be merged before. Build the module based on this branch Install it and try to upgrade to 8.0.0 (you will probably need to create a release from the most recent develop branch). You shouldn't have anymore error related to mail generation (there might be other errors down on the road but it's not the purpose of this PR)
| Possible impacts? | The module now relies on `GenerateThemeMailTemplatesCommand` which was introduced in PrestaShop 1.7.6 which means the module would now be only compatible with target versions above 1.7.6 I don't know if we should handle retrocompatibility with older versions? If so then we might need to extract this new code in a dedicated class or service to handle the switch more cleanly

### DIFF
--- a/classes/TaskRunner/Upgrade/UpgradeDb.php
+++ b/classes/TaskRunner/Upgrade/UpgradeDb.php
@@ -64,7 +64,7 @@ class UpgradeDb extends AbstractTask
             return new CoreUpgrader16($this->container, $this->logger);
         }
 
-        if (version_compare($this->container->getState()->getInstallVersion(), '8.0.0.0', '<')) {
+        if (version_compare($this->container->getState()->getInstallVersion(), '8.0.0', '<')) {
             return new CoreUpgrader17($this->container, $this->logger);
         }
 
--- a/classes/TaskRunner/Upgrade/UpgradeDb.php
+++ b/classes/TaskRunner/Upgrade/UpgradeDb.php
@@ -31,6 +31,7 @@ use PrestaShop\Module\AutoUpgrade\TaskRunner\AbstractTask;
 use PrestaShop\Module\AutoUpgrade\UpgradeException;
 use PrestaShop\Module\AutoUpgrade\UpgradeTools\CoreUpgrader\CoreUpgrader16;
 use PrestaShop\Module\AutoUpgrade\UpgradeTools\CoreUpgrader\CoreUpgrader17;
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\CoreUpgrader\CoreUpgrader80;
 use PrestaShop\Module\AutoUpgrade\UpgradeTools\SettingsFileWriter;
 
 class UpgradeDb extends AbstractTask
@@ -59,11 +60,15 @@ class UpgradeDb extends AbstractTask
 
     public function getCoreUpgrader()
     {
-        if (version_compare($this->container->getState()->getInstallVersion(), '1.7.0.0', '<=')) {
+        if (version_compare($this->container->getState()->getInstallVersion(), '1.7.0.0', '<')) {
             return new CoreUpgrader16($this->container, $this->logger);
         }
 
-        return new CoreUpgrader17($this->container, $this->logger);
+        if (version_compare($this->container->getState()->getInstallVersion(), '8.0.0.0', '<')) {
+            return new CoreUpgrader17($this->container, $this->logger);
+        }
+
+        return new CoreUpgrader80($this->container, $this->logger);
     }
 
     public function init()
--- a/classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
+++ b/classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
@@ -30,7 +30,6 @@ namespace PrestaShop\Module\AutoUpgrade\UpgradeTools\CoreUpgrader;
 use PrestaShop\Module\AutoUpgrade\UpgradeException;
 use PrestaShop\Module\AutoUpgrade\UpgradeTools\ThemeAdapter;
 use PrestaShop\PrestaShop\Core\CommandBus\CommandBusInterface;
-use PrestaShop\PrestaShop\Core\Domain\MailTemplate\Command\GenerateThemeMailTemplatesCommand;
 use PrestaShop\PrestaShop\Core\Domain\Theme\Command\AdaptThemeToRTLLanguagesCommand;
 use PrestaShop\PrestaShop\Core\Domain\Theme\ValueObject\ThemeName;
 use PrestaShop\PrestaShop\Core\Exception\CoreException;
@@ -84,33 +83,7 @@ class CoreUpgrader17 extends CoreUpgrader
         \Language::installSfLanguagePack($lang_pack['locale'], $errorsLanguage);
 
         if (!$this->container->getUpgradeConfiguration()->shouldKeepMails()) {
-            $mailTheme = \Configuration::get('PS_MAIL_THEME', null, null, null, 'modern');
-
-            $frontTheme = _THEME_NAME_;
-            $frontThemeMailsFolder = _PS_ALL_THEMES_DIR_ . $frontTheme . '/mails';
-            $frontThemeModulesFolder = _PS_ALL_THEMES_DIR_ . $frontTheme . '/modules';
-
-            $generateCommand = new GenerateThemeMailTemplatesCommand(
-                $mailTheme,
-                $lang_pack['locale'],
-                true,
-                $frontThemeMailsFolder,
-                $frontThemeModulesFolder
-            );
-            /** @var CommandBusInterface $commandBus */
-            $commandBus = $this->container->get('prestashop.core.command_bus');
-
-            try {
-                $commandBus->handle($generateCommand);
-            } catch (CoreException $e) {
-                throw new UpgradeException(
-                    $this->container->getTranslator()->trans(
-                        'Cannot generate email templates: %s.',
-                        [$e->getMessage()],
-                        'Modules.Autoupgrade.Admin'
-                    )
-                );
-            }
+            \Language::installEmailsLanguagePack($lang_pack, $errorsLanguage);
         }
 
         if (!empty($errorsLanguage)) {
--- a/classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
+++ b/classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
@@ -30,6 +30,7 @@ namespace PrestaShop\Module\AutoUpgrade\UpgradeTools\CoreUpgrader;
 use PrestaShop\Module\AutoUpgrade\UpgradeException;
 use PrestaShop\Module\AutoUpgrade\UpgradeTools\ThemeAdapter;
 use PrestaShop\PrestaShop\Core\CommandBus\CommandBusInterface;
+use PrestaShop\PrestaShop\Core\Domain\MailTemplate\Command\GenerateThemeMailTemplatesCommand;
 use PrestaShop\PrestaShop\Core\Domain\Theme\Command\AdaptThemeToRTLLanguagesCommand;
 use PrestaShop\PrestaShop\Core\Domain\Theme\ValueObject\ThemeName;
 use PrestaShop\PrestaShop\Core\Exception\CoreException;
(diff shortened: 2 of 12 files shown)


classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
```diff
@@ -28,6 +28,9 @@
 namespace PrestaShop\Module\AutoUpgrade\UpgradeTools\CoreUpgrader;
 
 use PrestaShop\Module\AutoUpgrade\UpgradeException;
+use PrestaShop\PrestaShop\Core\CommandBus\CommandBusInterface;
```
jolelievre, line 31: I don't know if we are allowed to use core classes like this in the upgrade module Especially since these classes may not be available in the source version and/or the target version.

classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
```diff
@@ -87,7 +90,33 @@ protected function upgradeLanguage($lang)
         \Language::installSfLanguagePack($lang_pack['locale'], $errorsLanguage);
 
         if (!$this->container->getUpgradeConfiguration()->shouldKeepMails()) {
-            \Language::installEmailsLanguagePack($lang_pack, $errorsLanguage);
+            $mailTheme = \Configuration::get('PS_MAIL_THEME', null, null, null, 'modern');
```
jolelievre, line 93: Should we handle backward compatibility and use the `Language::installEmailsLanguagePack` for versions older than the 1.7.6? If so maybe this should be extracted in a dedicated tool class or service.

atomiix: I'm in favor to keep the backward compatibilty!

classes/UpgradeTools/CoreUpgrader/CoreUpgrader80.php
```diff
@@ -0,0 +1,135 @@
...
+    {
+        parent::initConstants();
+
+        /*if (!file_exists(SETTINGS_FILE_PHP)) {
+            throw new UpgradeException($this->container->getTranslator()->trans('The app/config/parameters.php file was not found.', array(), 'Modules.Autoupgrade.Admin'));
+        }
+        if (!file_exists(SETTINGS_FILE_YML)) {
+            throw new UpgradeException($this->container->getTranslator()->trans('The app/config/parameters.yml file was not found.', array(), 'Modules.Autoupgrade.Admin'));
+        }*/
+
```
PierreRambaud, line 48:
```suggestion
```

classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
PierreRambaud: Yes keep it

classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
PierreRambaud: Afaik you can do it:
image
It will only raise an error if the class is not found and executed by PHP

matks (commented):
PR is OK IMO, if you apply latest feedbacks from @PierreRambaud

matks:
PHPStan issue not related to change https://github.com/PrestaShop/PrestaShop/issues/26861

atomiix (dismissed):

classes/UpgradeTools/CoreUpgrader/CoreUpgrader80.php
```diff
@@ -0,0 +1,128 @@
...
+            $generateCommand = new GenerateThemeMailTemplatesCommand(
+                $mailTheme,
+                $lang_pack['locale'],
+                true,
+                $frontThemeMailsFolder,
+                $frontThemeModulesFolder
+            );
+            /** @var CommandBusInterface $commandBus */
+            $commandBus = $this->container->get('prestashop.core.command_bus');
```
atomiix, line 97: It seems that this `container` is not what you think 😅

classes/TaskRunner/Upgrade/UpgradeDb.php
```diff
@@ -59,11 +60,15 @@ public function run()
 
     public function getCoreUpgrader()
     {
-        if (version_compare($this->container->getState()->getInstallVersion(), '1.7.0.0', '<=')) {
+        if (version_compare($this->container->getState()->getInstallVersion(), '1.7.0.0', '<')) {
             return new CoreUpgrader16($this->container, $this->logger);
         }
 
-        return new CoreUpgrader17($this->container, $this->logger);
+        if (version_compare($this->container->getState()->getInstallVersion(), '8.0.0.0', '<')) {
```
PierreRambaud, line 67: :eyes:

classes/UpgradeTools/CoreUpgrader/CoreUpgrader80.php
jolelievre: t's not? What is it then? It's not the Symfony container?

atomiix: @jolelievre nope ^^' it's this : https://github.com/PrestaShop/autoupgrade/blob/dev/classes/UpgradeContainer.php

jolelievre: Damn 😅. I admit I didn't check if it worked at all... shame!

Does it mean there is no way to use the Symfony container or any services it might contain? Should I reimplement the whole thing? (seems unrealistic)
Or can I get the container with `SymfonyContainer::getInstance` for example?

atomiix: @jolelievre you should be able to do something like this in: https://github.com/PrestaShop/autoupgrade/blob/dev/classes/UpgradeTools/ModuleAdapter.php
```php
public function getCommandBus()
{
    if (null === $this->commandBus) {
        $this->commandBus = $this->symfonyAdapter
            ->initAppKernel()
            ->getContainer()
            ->get('prestashop.core.command_bus');
    }

    return $this->commandBus;
}
```
and then use `$commandBus = $this->container->getModuleAdapter()->getCommandBus();`

jolelievre (commented):
Thanks @atomiix for the finishing touch 😉
I'd like to approve but I can't approve my own PR 😅

jolelievre:
Oh and it seems like PHPStan is not happy about something

jolelievre (commented):
I can't approve this PR, seems legitimate I just have one suggestion

I didn't see any modif related to the bug you mentioned regarding `LegacyLoader`, is it fixed in the core?

classes/UpgradeTools/CoreUpgrader/CoreUpgrader80.php
```diff
@@ -90,8 +90,8 @@ protected function upgradeLanguage($lang)
                 $mailTheme,
                 $lang_pack['locale'],
                 true,
-                $frontThemeMailsFolder,
-                $frontThemeModulesFolder
+                is_dir($frontThemeMailsFolder) ? $frontThemeMailsFolder : '',
```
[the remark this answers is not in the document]
atomiix: These folders aren't supposed to always exist, if you check the [`themes/classic/` folder](https://github.com/PrestaShop/PrestaShop/tree/develop/themes/classic), you'll see no `mails` folder.
Also, I see that's what we do in [src/PrestaShopBundle/Controller/Admin/Improve/Design/MailThemeController.php](https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Controller/Admin/Improve/Design/MailThemeController.php#L119-L138)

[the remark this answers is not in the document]
atomiix: >I didn't see any modif related to the bug you mentioned regarding LegacyLoader, is it fixed in the core?

@jolelievre yep, it's fixed in the core ;)

kpodemski (dismissed):
I'm wondering if `ModuleAdapter:: getCommandBus` shouldn't throw an error if we're not in the context of 1.7+, but of course, this is an edge case scenario.

khouloudbelguith:
Hi @jolelievre,

1. I created a release from this PR: https://github.com/PrestaShop/PrestaShop/pull/27276
2. I installed PS1782 with English language & France country
3. I added the Algeria pack
4. I enabled multistore & created a new shop
5. I created some order
6. I installed the autoupgrade module using this PR
7. During the upgrade, I have an error  

image

What do you think, this error is related to the PR module or the develop PR?

PS: I tried to make an upgrade from PS1778 to PS1782 using this PR => ok

Thanks!

jolelievre:
Hi @khouloudbelguith 

sorry I cowardly left this PR's responsibility to @atomiix so I believe he is looking into it but can't reproduce the bug...

khouloudbelguith:
Hi @jolelievre, @atomiix,

I tried again with PS1782 fresh install & an upgrade to the PR & same error

https://user-images.githubusercontent.com/16067358/149154665-cabf5046-e4cd-45af-b8e9-a1890e54e3b4.mp4

Thanks!

atomiix:
@khouloudbelguith it seems that your PHP version is 7.1, can you try with PHP 7.2 or above as the develop branch is not compatible with PHP < 7.2?

khouloudbelguith (changes requested):
Hi @atomiix,

Yes, It is ok with PHP7.3
image
I tried to make an upgrade from PS1782 to https://github.com/PrestaShop/PrestaShop/pull/27276
But after the upgrade, I have a lot of exceptions:
1. Go to BO > Orders > View Order > See exception 
image
1. Go to BO > Orders > Shopping cart page > See exception 
image
 1. Go to BO > Customers page > View customer > see exception
image
1. Go to BO > Modules > Module Catalog > See exception 
image
1. Go to FO > Sign in > Se exception 
image


Thanks!

atomiix:
@khouloudbelguith I think we might be in an infinite loop as this PR #441 seems to be solving the issue you just had but to validate it you need this one 😅
Maybe we can merge this one as the upgrade itself went well, see if #411 solve the issue you had and if not, create a new issue. WDYT?

PierreRambaud (approved):

Progi1984 (approved):

khouloudbelguith (approved):
Hi @atomiix, @jolelievre,

It is ok :heavy_check_mark: 

I tried to upgrade PS1778 to PS1783 using this PR => OK (PHP7.3)
I tried to generate modern/classic emails in different languages (RTL & non RTL)
I tried to check the HTTP of emails 'Modern template / classic template)
I tried to translate some emails and resend them


I tried to upgrade from PS1769 to 178x branch using this PR
I tried to generate some emails with RTL to check this [issue](https://github.com/PrestaShop/PrestaShop/pull/27614): => ok

I tried to upgrade from PS1783 to this PR: https://github.com/PrestaShop/PrestaShop/pull/27276 => OK
I tried to upgrade from 178x branch to this PR: https://github.com/PrestaShop/PrestaShop/pull/27276 => OK

upgrade178x_develop_Pr
I tried to generate some emails in Arabic language & English language
I checked Gmail, outlook web version, outlook desktop version, maildev server

https://user-images.githubusercontent.com/16067358/154260004-31cc232b-d060-4173-b432-f67c2b6292f5.mp4


I found some bugs but not the scope of this PR, which are blocking

PS:

Before the upgrade to this PR:  https://github.com/PrestaShop/PrestaShop/pull/27276
I uninstall those modules: ps_metrics, ps_facebook, Prestashop Checkout, Google marketing, welcome
After the upgrade
1. Go to BO > Orders > View Order > See exception  & Add new order from BO, will be fixed with this https://github.com/PrestaShop/autoupgrade/pull/441
image
1. Go to BO > Orders > Shopping cart page > See exception => will be fixed with this https://github.com/PrestaShop/autoupgrade/pull/441
image
 1. Go to BO > Customers page > View customer > see exception  => will be fixed with this https://github.com/PrestaShop/autoupgrade/pull/441
image
1. Go to BO > Modules > Module Catalog > See exception => I will create a new issue.
image
1. Go to FO > Sign in > See exception  => will be fixed with this https://github.com/PrestaShop/autoupgrade/pull/441
image
1. Go to FO as a visitor & add a product to cart > see exception  => will be fixed with this https://github.com/PrestaShop/autoupgrade/pull/441
image
1. Go to BO > SEO & URLs > Reffers Tab > see exception => will be fixed here: https://github.com/PrestaShop/PrestaShop/pull/27098
image
1. Go to BO > SEO & URLs page > see exception => I will create a new issue
image

1. Go to BO > Design > Email theme > choose any theme (Classic or modern) > View Htpp or any action for **order_conf** > see exception  => will be fixed with this https://github.com/PrestaShop/autoupgrade/pull/441
image

1. The Security menu is not displayed => I will create a new issue.

Thanks!

Progi1984:
Thanks @jolelievre & @khouloudbelguith

PR #439, opened 2021-12-09
Fix untranslatable Tabs for shops upgraded to 1.7.8.0

Prestaworks:
| Questions         | Answers
| Description?      | Fix for issue 26933
| Type?             | bug fix
| BC breaks?        |  no
| Deprecations?     |  no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#26933.
| How to test?      | Follow instructions in issue
| Possible impacts? |


atomiix (approved):

PierreRambaud (changes requested):
Since it's a bug in 1.7.8.0, to fix it properly, shouldn't be a 1.7.8.3 script, and also checking if the wording/wording_domain are empty when updating :thinking: 
I would like @eternoendless 's opinion

Prestaworks:
> 

Is it really a bug in 1.7.8.0 ? it's working fine for new installations as far as I know.

PierreRambaud:
> > 
> 
> Is it really a bug in 1.7.8.0 ? it's working fine for new installations as far as I know.

Yes for new installation, but the bug is still present for shop that are already been upgraded

Prestaworks:
> > > 
> > 
> > 
> > Is it really a bug in 1.7.8.0 ? it's working fine for new installations as far as I know.
> 
> Yes for new installation, but the bug is still present for shop that are already been upgraded

I understand your point, but how far away is 1.7.8.3? (it's no problem to run the queries multiple times as all they do is correct the values)

PierreRambaud:
> > > > 
> > > 
> > > 
> > > Is it really a bug in 1.7.8.0 ? it's working fine for new installations as far as I know.
> > 
> > 
> > Yes for new installation, but the bug is still present for shop that are already been upgraded
> 
> I understand your point, but how far away is 1.7.8.3? (it's no problem to run the queries multiple times as all they do is correct the values)

With your actual request, you can break changes as the menu is now translatable. If you add a check to verify if `wording` and `wording_domain` are empty when you update queries, that's ok :)

Prestaworks:
> > > > > 
> > > > 
> > > > 
> > > > Is it really a bug in 1.7.8.0 ? it's working fine for new installations as far as I know.
> > > 
> > > 
> > > Yes for new installation, but the bug is still present for shop that are already been upgraded
> > 
> > 
> > I understand your point, but how far away is 1.7.8.3? (it's no problem to run the queries multiple times as all they do is correct the values)
> 
> With your actual request, you can break changes as the menu is now translatable. If you add a check to verify if `wording` and `wording_domain` are empty when you update queries, that's ok :)

Are you saying that wording and wording_domain may change on these items?

PierreRambaud:
If you upgrade to 1.7.8.0, your fix is legit, but we still need a fix for the next release.
Your fix will not work with 1.7.8.3 because if someone changes translations it will break his.

I'm also not even sure the https://github.com/PrestaShop/autoupgrade/pull/439/files#diff-b5a927db4ced9372b0307b5bb47e3ee232066f7cc28813ff6f12ee1f08ee95b3R282 about linklist should be here as it's related to ps_linklist module :thinking:

Prestaworks:
> 

Can you explain what you mean by "if someone changes the translations" and what will break?

PierreRambaud:
Imager are better than long story:
image

eternoendless:
Yep, the magic happens here: https://github.com/PrestaShop/PrestaShop/blob/1.7.8.x/src/PrestaShopBundle/Controller/Api/TranslationController.php#L348-L361

The translator needs to find the wording & domain so that it knows which tab it has to update into tab_lang.

upgrade/sql/1.7.8.0.sql
```diff
@@ -159,6 +159,129 @@ ALTER TABLE `PREFIX_log`
 ALTER TABLE `PREFIX_tab` ADD `wording` VARCHAR(255) DEFAULT NULL AFTER `icon`;
 ALTER TABLE `PREFIX_tab` ADD `wording_domain` VARCHAR(255) DEFAULT NULL AFTER `wording`;
 
+UPDATE `PREFIX_tab` SET `wording`='Dashboard', `wording_domain`='Admin.Navigation.Menu' WHERE `class_name`='AdminDashboard';
```
eternoendless, line 162: I'd say to run the changes in 1.7.8.3, it's highly unlikely a tab will have been reworded by hand in `tab`. Also, you can add this to every line to make extra sure that it hasn't been set already:
```suggestion
UPDATE `PREFIX_tab` SET `wording`='Dashboard', `wording_domain`='Admin.Navigation.Menu' WHERE `class_name`='AdminDashboard' AND COALESCE(`wording`, '') = '' AND COALESCE(`wording_domain`, '') = '';
```

Prestaworks: Will 1.7.8.3 be released in the near future?

eternoendless: https://github.com/PrestaShop/PrestaShop/issues/26969

Prestaworks: ok, not so far away then :)
Well it sounds that you like the updates moved to 1.7.8.3 then? should I do that change?

eternoendless: Yes please!

atomiix (approved):

upgrade/sql/1.7.8.0.sql
PierreRambaud: What's the aim of using `COALESCE` with a single value? `COALESCE` is a shortcut for `CASE WHEN`, it returns the first non-null value :thinking: 
Using `wording = '' AND wording_domain = ''` is enough.

PierreRambaud (dismissed):
I don't understand the usage of `COALESCE`. 
https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#function_coalesce

upgrade/sql/1.7.8.0.sql
PierreRambaud: After checking the `Entity\Tab.php` file, the field is nullable by default, and we authorize empty value, so `IS NULL` should be enough

Prestaworks: removed the columns from the table?

eternoendless: No it's not enough, because both `wording` and `wording_domain` are varchar but default to NULL. So "empty" can be either NULL or empty string. The coalesce works as a shorthand to `(wording = '' OR wording IS NULL)`. 

If we don't coalesce to string and you use IS NULL then we risk leaving out tabs that may have been edited by hand by a developer exploring the original issue, then emptied out to prepare for the update (but not set to NULL). 

I can't see any scenario where you'd like the domain and wording of these tabs empty on purpose.

PierreRambaud: There is a difference between null and non-null values.
> I can't see any scenario where you'd like the domain and wording of these tabs empty on purpose.

Even if you can't see a scenario, I see a lot of them where you can remove unneeded wording or wording_domain and leave fields empty because you don't care about translations in your BO.
If the field mustn't be empty, we have to add a non-empty constraint directly in the BO too and remove the nullable field :thinking:

eternoendless (approved):
LGTM

upgrade/sql/1.7.8.0.sql
eternoendless: I honestly think your use case is purely hypothetical. I can't imagine a merchant calling their tech guy and say "hey can you please make it so that I can no longer translate the menu items?"

PierreRambaud: I follow what has been done in the code, nothing more :/
We have to be consistent between what we created with `Tab` entity and this one.

eternoendless: I understand but we don't control what people might have put in their database. This accounts for that while leaving customization as is.

eternoendless:
I think I fixed a bug related to this feature. It looks like the reverse translation is not triggered when wording and wording_domain are empty: https://github.com/PrestaShop/PrestaShop/pull/27011

eternoendless:
Thank you @Prestaworks

PR #443, opened 2021-12-29
Fixed translations in AutoUpgrade

Progi1984:
| Questions         | Answers
| Description?      | Fixed translations in AutoUpgrade
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | **BEFORE** : Translations for module AutoUpgrade are not available in [BackOffice](https://devdocs.prestashop.com/8/modules/creation/module-translation/classic-system/#creating-translation-dictionary-files)<br/>**AFTER** : Translations for module AutoUpgrade are available in [BackOffice](https://devdocs.prestashop.com/8/modules/creation/module-translation/classic-system/#creating-translation-dictionary-files)


classes/UpgradePage.php
```diff
@@ -177,7 +177,7 @@ public function display($ajaxResult)
         $errMessageData = $this->getErrorMessage();
         if (!empty($errMessageData)) {
             return $twig
-                ->render('@ModuleAutoUpgrade:error.twig', $errMessageData);
+                ->render('@ModuleAutoUpgrade:error.html.twig', $errMessageData);
```
atomiix, line 180:
```suggestion
                ->render('@ModuleAutoUpgrade/error.html.twig', $errMessageData);
```

atomiix (commented):
A little change to be sur to be compatible with PrestaShop 8.0. Also, can you explain why adding the `.html` solves the translation issue? 😅

views/templates/block/activityLog.html.twig
```diff
@@ -1,23 +1,23 @@
 <div class="bootstrap" id="activityLogBlock" style="display:none">
     <div class="panel">
         <div class="panel-heading">
-            {{ 'Activity Log'|trans }}
+            {{ 'Activity Log'|trans({}, 'Modules.Autoupgrade.Admin') }}
```
eternoendless, line 4: Note that it's useless to add translation domain, this module uses its own translator, which goes straight to the legacy translation system.

See:
- https://github.com/PrestaShop/autoupgrade/blob/v4.13.0/classes/UpgradeTools/Translator.php
- https://github.com/PrestaShop/autoupgrade/blob/v4.13.0/classes/Twig/TransFilterExtension.php

The wordings must be added to the legacy dictionaries as explained in the docs

eternoendless (changes requested):

views/templates/block/activityLog.html.twig
eternoendless: What we could do is reimplement the translator within the module to use xliff files. That way the dictionaries could be created using the BO interface in 1.7.8, then reused everywhere. And it also would allow to break one dependency from the core.

classes/UpgradePage.php
PierreRambaud: Are you sure it's working with all Symfony versions?

atomiix: Yep: [https://symfony.com/doc/2.8/templating/namespaced_paths.html](https://symfony.com/doc/2.8/templating/namespaced_paths.html)

PR #448, opened 2022-01-13
Add rtl files generation

tegbessou:
| Questions         | Answers
| Description?      | Regenerate RTL files when upgrading PS version
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Relative to PrestaShop/PrestaShop#27307.
| How to test?      | Install PS 1.7.7.8, then add to this PS an RTL language. Then upgrade your PS version to 1.7.8.3, then go to the homepage you shouldn't see the bug describe by the issue and your RTL files must be regenerated. You also have to try an upgrade by disabling the RTL regenerate option
| Possible impacts? | This can impact the all upgrade process


classes/Parameters/UpgradeConfiguration.php
```diff
@@ -149,13 +149,21 @@ public function shouldSwitchToDefaultTheme()
...
-     * @return bool True if we are allowed to update th default theme files
+     * @return bool True if we are allowed to update the default theme files
      */
     public function shouldUpdateDefaultTheme()
     {
         return (bool) $this->get('PS_AUTOUP_UPDATE_DEFAULT_THEME');
     }
 
+    /**
+     * @return bool True if we are allowed to update RTL files
```
PierreRambaud, line 160:
```suggestion
     * @return bool True if we should update RTL files
```

tegbessou:
I'm waiting for wording form @MatShir

classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
```diff
@@ -112,4 +118,54 @@ protected function upgradeLanguage($lang)
...
+        $this->removeExistingRTLFiles($themes);
+
+        foreach ($themes as $theme) {
+
+            $adaptThemeToTRLLanguages = new AdaptThemeToRTLLanguagesCommand(
+                new ThemeName($theme['name'])
+            );
+
+            /** @var CommandBusInterface $commandBus */
+            $commandBus = $this->container->getModuleAdapter()->getCommandBus();
```
PierreRambaud, line 147: It will not work for all 1.7 version, you need to be sure the command bus is available
The AdaptThemeToRTLLanguagesCommand has been added in [1.7.6.0](https://github.com/PrestaShop/PrestaShop/blob/1.7.6.0/src/Core/Domain/Theme/Command/AdaptThemeToRTLLanguagesCommand.php)

tests/phpstan/phpstan.neon
```diff
@@ -20,5 +20,6 @@ parameters:
 		- '#AppKernel#'
 		- '#Access to offset 0 on an unknown class mysqli_result.#'
 		- '#Iterating over an object of an unknown class mysqli_result.#'
+		- '#PrestaShop\\PrestaShop\\Core\\CommandBus\\TacticianCommandBusAdapter#'
 
 	level: 5
```
PierreRambaud, line 25:
```suggestion
	level: 5

```

classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
tegbessou: So I need to verify with `class_exists ` if the class exists before launching my script. If I do this it's okay for you?

tegbessou:
The wording was in the issue! So you can review thx 😄

classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
matks: If you do this, this process will only work if target version is higher than 1.7.6

- if user upgrades from 1.7.3.0 to 1.7.7.5 ✅ 
- if user upgrades from 1.7.7.8 to 1.7.8.2 ✅ 
- if user upgrades from 1.7.3.0 to 1.7.5.1 🟥

classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
```diff
@@ -112,4 +118,58 @@ protected function upgradeLanguage($lang)
...
+    protected function updateTheme()
+    {
+        parent::updateTheme();
+
+        $this->updateRTLFiles();
+    }
+
+    protected function updateRTLFiles()
+    {
+        if (!class_exists(AdaptThemeToRTLLanguagesCommand::class)) {
```
atomiix, line 131: I think this will fail if the class doesn't exist?
```suggestion
        if (!class_exists('PrestaShop\PrestaShop\Core\Domain\Theme\Command\AdaptThemeToRTLLanguagesCommand')) {
```

tegbessou: No, I've tested this and this work!

matthieu-rolland (approved):
LGTM 👍

classes/UpgradeTools/CoreUpgrader/CoreUpgrader17.php
PierreRambaud: I confirm it works

PierreRambaud (approved):

saulaski (approved):

matks (approved):
As discussed, PR is OK for given scope

HanaRebaiQA:
Hello @tegbessou 
Thanks for the fix. @khouloudbelguith  and me are checked this PR and verified that the issue is fixed. 
https://watch.screencastify.com/v/eWIGzMfG9zZqXfKbuU61

Here the tested versions :
1. Upgrade de 16124 => 178x & RTL option enabled
Rollback 178x => 16124 ==> ok
2. Upgrade de 16124 => 178x & RTL option disabled
Rollback 178x => 16124  ==> ok
3. Upgrade 1778 => 178x & RTL option enabled
Rollback 178x => 1778 => ok
4. Upgrade 1782 => 178x & RTL option enabled
Rollback 178x => 1782 => ok
5. Upgrade de 1769 => 178x & RTL option disabled
Rollback 178x => 1769 => ok

So, we can confirm this PR : QA:heavy_check_mark: 

Thank you!

tegbessou:
Thank you for your QA 😄

atomiix:
Thanks @tegbessou & @HanaRebaiQA!

PR #442, opened 2021-12-27
Fix line not compatible with PHP < 7

Prestaworks:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Fix for issue 27087
| Type?             | bug fix
| BC breaks?        |  no
| Deprecations?     |  no
| Fixed ticket?     | Related to PrestaShop/PrestaShop#27087.
| How to test?      | See Issue for details.
| Possible impacts? |

PR #454, opened 2022-01-20
Fix dirname call for PHP <7.0

bfontaine:
`dirname(..., 3)` in PHP 7.0+ is equivalent to `dirname(dirname(dirname(...)))` in PHP <7.0.

| Questions         | Answers
| Description?      | Changing a `dirname()` call to avoid using the second argument which is only supported by PHP 7.0+
| Type?             | bug fix / improvement
| BC breaks?        | I don’t know. What’s BC?
| Deprecations?     | I don’t know
| Fixed ticket?     | _none_
| How to test?      | Run the autoupgrade script using PHP 5.6
| Possible impacts? |


kpodemski:
Thank you @bfontaine 

This fix has already been submitted by another contributor, here:
https://github.com/PrestaShop/autoupgrade/pull/442

Anyway, thank you for your PR, and we will make sure to implement the above fix in one of the next releases.

PR #465, opened 2022-03-02
Upgrade process will not stop due to existing hooks

Prestaworks:
| Questions         | Answers
| Description?      | Fix for ticket 27184
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#27184
| How to test?      | Upgrade from 1.6 to 1.7 succeeds without warnings.
| Possible impacts? | Nothing outside the upgrade process

### DIFF
--- a/upgrade/sql/1.7.0.0.sql
+++ b/upgrade/sql/1.7.0.0.sql
@@ -154,7 +154,7 @@ CREATE TABLE `PREFIX_module_access` (
 
 /* PHP:add_quick_access_tab(); */;
 
-INSERT INTO `PREFIX_hook` (`name`, `title`, `description`, `position`) VALUES
+INSERT IGNORE INTO `PREFIX_hook` (`name`, `title`, `description`, `position`) VALUES
   ('actionValidateCustomerAddressForm', 'Customer address form validation', 'This hook is called when a customer submit its address form', '1'),
   ('displayAfterCarrier', 'After carriers list', 'This hook is displayed after the carrier list in Front Office', '1'),
   ('displayCarrierExtraContent', 'Display additional content for a carrier (e.g pickup points)', 'This hook calls only the module related to the carrier, in order to add options when needed.', '1'),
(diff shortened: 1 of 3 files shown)


upgrade/sql/1.7.0.0.sql
```diff
@@ -154,7 +154,7 @@ CREATE TABLE `PREFIX_module_access` (
 
 /* PHP:add_quick_access_tab(); */;
 
-INSERT INTO `PREFIX_hook` (`name`, `title`, `description`, `position`) VALUES
+INSERT IGNORE INTO `PREFIX_hook` (`name`, `title`, `description`, `position`) VALUES
```
saulaski, line 157: What about REPLACE to force changes from the upgrade, instead of keeping the previous obsolete values ?

Prestaworks: The important thing is the name column, I don't thing we need to write a replace query for this?

atomiix (dismissed):

upgrade/sql/1.7.0.0.sql
saulaski: Not 100% sure either replacing is important. I just assumed that upgrade could fix a typo in description, or update the position.

Prestaworks: I think the position column was related to if the hook allowed drag and drop of the order of the modules.

saulaski: Most likely yes, then the command shouldn't replace current position. Your PR is fine :)

saulaski (approved):

Progi1984 (dismissed):

Progi1984 (approved):

PierreRambaud (approved):

saulaski (approved):

atomiix (approved):

kpodemski:
@florine2623 I can confirm this issue. Third-party modules are using these three hooks mentioned in the issue, and if they exist in the DB, the upgrade shows warnings.

matks:
I put back the Waiting for QA again, hoping thanks to @kpodemski we can reproduce the bug

eternoendless (approved):

eternoendless:
Thank you @Prestaworks

PR #474, opened 2022-04-09
State field will be longer from 1.7.8.6, not .4

kpodemski:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | As discussed here: https://github.com/PrestaShop/autoupgrade/pull/455#issuecomment-1092590347
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | n/a
| How to test?      | State field should be longer after upgrading to 1.7.8.6

PR #473, opened 2022-04-05
Add pages to enable maintenance mode before upgrade

boubkerbribri:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Add pages to enable maintenance mode before upgrade
| Type?             | refacto
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | no
| How to test?      | Tested with workflow
| Possible impacts? | none

PR #469, opened 2022-03-16
Accept more characters in the product name

kpodemski:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | https://github.com/PrestaShop/PrestaShop/pull/27933
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/9644
| How to test?      | After upgrade from 1.7 to 8 `order_detail` should have a column `product_name` of the `TEXT` type
| Possible impacts? | Please indicate what parts of the software we need to check to make sure everything is alright.
| Sponsor company | impSolutions

PR #458, opened 2022-02-03
Add displayCheckoutBeforeConfirmation hook

Hlavtox:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Add new https://github.com/PrestaShop/PrestaShop/pull/27437 hook during upgrade
| Type?             | new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| How to test?      | Install a new shop, make an upgrade to 8.0.0, check the hook is in DB
| Possible impacts? | No

PR #481, opened 2022-04-25
Release 4.15.0

matks:
| Questions         | Answers
| Description?      | Release 4.15.0

## How to test

I provided below a list of the changes between 4.15.0 and 4.14.0 . I think only the  "Upgrade to 1.7.8.6" and "New features and improvements" changes can impact the release.

### Continuous integration improvements and bug fixes
https://github.com/PrestaShop/autoupgrade/pull/459
https://github.com/PrestaShop/autoupgrade/pull/460
https://github.com/PrestaShop/autoupgrade/pull/461
https://github.com/PrestaShop/autoupgrade/pull/471
https://github.com/PrestaShop/autoupgrade/pull/473
https://github.com/PrestaShop/autoupgrade/pull/477
https://github.com/PrestaShop/autoupgrade/pull/478

### Version bump
https://github.com/PrestaShop/autoupgrade/pull/480

#### Upgrade to 8.0.0
https://github.com/PrestaShop/autoupgrade/pull/456
https://github.com/PrestaShop/autoupgrade/pull/422
https://github.com/PrestaShop/autoupgrade/pull/464
https://github.com/PrestaShop/autoupgrade/pull/469
https://github.com/PrestaShop/autoupgrade/pull/446
https://github.com/PrestaShop/autoupgrade/pull/458

#### Upgrade to 1.7.8.6
https://github.com/PrestaShop/autoupgrade/pull/455
https://github.com/PrestaShop/autoupgrade/pull/474

### New features and improvements
https://github.com/PrestaShop/autoupgrade/pull/472
https://github.com/PrestaShop/autoupgrade/pull/451
https://github.com/PrestaShop/autoupgrade/pull/434
https://github.com/PrestaShop/autoupgrade/pull/467
https://github.com/PrestaShop/autoupgrade/pull/421
https://github.com/PrestaShop/autoupgrade/pull/440
https://github.com/PrestaShop/autoupgrade/pull/445
https://github.com/PrestaShop/autoupgrade/pull/450


composer.json
```diff
@@ -1,38 +1,39 @@
...
+  "homepage": "https://github.com/PrestaShop/autoupgrade",
+  "license": "AFL-3.0",
+  "authors": [
+    {
+      "name": "PrestaShop SA",
+      "email": "contact@prestashop.com"
+    }
+  ],
+  "require": {
+    "php": ">=5.6",
```
jolelievre, line 13: The PHP version here doesn't match with the module's dependency requiring PHP 7.2 minimum

classes/UpgradeSelfCheck.php
```diff
@@ -29,13 +29,14 @@
...
 use ConfigurationTest;
+use Shop;
 
 class UpgradeSelfCheck
 {
     /**
      * Recommended PHP Version. If below, display a notice.
      */
-    const RECOMMENDED_PHP_VERSION = 70103;
+    const RECOMMENDED_PHP_VERSION = 70205;
```
jolelievre, line 39: IMHO this is a hug change that would required release a new major version, not a minor one.
So 5.0.0 should be the new version number

config.xml
```diff
@@ -2,7 +2,7 @@
 <module>
   <name>autoupgrade</name>
   <displayName><![CDATA[1-Click Upgrade]]></displayName>
-  <version><![CDATA[4.14.0]]></version>
+  <version><![CDATA[4.15.0]]></version>
```
jolelievre, line 5:
```suggestion
  <version><![CDATA[5.0.0]]></version>
```

jolelievre (commented):
I have nothing to say agains all these changes, except for the version number.

IMHO changing the PHP min version is a modification worth releasing a new major version. Ad actually, even including all the upgrade scripts in this module was already a modification worth a major release which should have been done in v4.13 already (at the time those scripts were integrated)

classes/UpgradeSelfCheck.php
kpodemski: The question is: is this blocking an upgrade process, or is it simply showing a warning?

kpodemski: There's no way to run PrestaShop 1.6 on PHP 7.2 without changes in the core. Even if this is just a warning, it should be reverted. The `autoupgrade` module should stay with the minimum compatibility of PHP 7.1. Ideally, it would be 5.6.

This is the same problem we discuss on Slack @matks - this is the time to introduce partial updates, for example: from 1.6 to 1.7.8, from 1.7.8 to 8, etc.

matks: PR is here https://github.com/PrestaShop/autoupgrade/pull/440

I need an autoupgrade release to be able to deliver 1.7.8.6 . I will not start a revert process and another set of high-risk changes in autoupgrade 😄  and delay 1.7.8.6 further. So if more changes are required I will create a `4.x` branch from tag 4.14.0 and only apply the 2 commits I need and release that. We can keep the big issues for later.

kpodemski: It's a must-have to test it on PHP 7.1 and 1.6 to check if it is still possible to upgrade from 1.6 to 1.7, and it's just a warning.

kpodemski: I tested it, and from what I see, it's only a warning that doesn't prevent me from upgrading.

btw. Something has changed, and we scan files from the disk before opening a module configuration page. I had to wait for 5 minutes to open `autoupgrade` 😭

matks: @kpodemski v5.0.0 is not released yet ;) (and we could always deliver a patch 5.0.1 if needed) so if you find issues, let's write bug reports

I think I'll have to do my `4.x` branch just for 1.7.8.6, other changes might not be safe to release now

jolelievre: It's not just about the warning, all the PHP files have been modified to use short array syntax `[]` which is only usable in PHP 7.2

So if you plan on creating a branch be careful to start it before all those files where modified.

kpodemski: @jolelievre hm? short array? it's from 5.4

kpodemski: + I just installed it on 7.1 and 1.6.1.24 and it's displaying everything properly.

MeKeyCool: I have some troubles to understand the subject.
I'll share my understanding of versioning and how it should work then please enlighten me about what I miss if necessary.

1. `PrestaShop/autoupgrade` is an independent project respecting semver rules.
  * So if you introduce a breaking change => version of project should change from 4.y.z to 5.0.0 on the concerned branch
  * Then it is not clear for me, does changing `UpgradeSelfCheck::RECOMMENDED_PHP_VERSION` from 7.1 to 7.2 introduces a BC ? The comment (`Recommended PHP Version. If below, display a notice.`) implies that it is only about log/notice, is the comment right ? If this change necessary for bug fix, how could it be warning only ?
2. Is the bug you are looking to solve is in Prestashop 1.6 ? 
  * If it is not, what the matter with Prestashop 1.7 has dependency on autoupgrade 4 and Prestashop 1.6 has dependency on autoupgrade 5 ? (furthermore, Prestashop 1.7 has already minimum [PHP dependency on 7.2](https://github.com/PrestaShop/PrestaShop/blob/533ba0179b5c03f4f3a55c5291f7e3bd7024fe2a/composer.json#L17) so changing Prestashop 1.7 dependency to autoupgrade 4 to autoupgrade 5 shouldn't change anything, does it ? Does autoupgrade change breaks something else than php minimum version ? 
  * If it is, is the bug fix a security issue ? If it is not, I thought we won't solve issues. :thinking: 
 
 Then everything looks simple for me but I thing I didn't get some key elements ^^'

kpodemski: Guys. Everything seems to be working fine, it's just this warning that should stay on PHP 7.1 instead of 7.2. I don't see any other issues 🤷

jolelievre: @kpodemski my bad about the short array So indeed the PHP minimum version shouldn't be modified and that should do it

jolelievre: I thought it was the minimum version for the module, but it turns out it's used to check the server target version so no problem if we revert it I guess this should be dynamically filled based on PS target version but I believe @matthieu-rolland was working on such a feature

kpodemski, line 39:
```suggestion
    const RECOMMENDED_PHP_VERSION = 70103;
```

This should be changed in the future for a version dedicated to PrestaShop 8. We shouldn't allow people to upgrade directly from < 1.7.7 to 8.

PierreRambaud: The aim of this module is to allow the old version of the PrestaShop project to be updated, the more your support, the best is it.
Imho, we should keep at least the 7.1, or as @jolelievre said earlier, we could update the major version.
For example, version 4 is for 1.6 & 8, then version 5 is for 1.7 & 8, etc.

jolelievre: Agreed

PR #496, opened 2022-08-11
Added stock_mvt tables upgrade

margud:
| Questions         | Answers
| Description?      | Changing the stock_mvt tables quantity input to unsigned. It is always positive number and by doing this we can insert twice as big int and after this [PR](https://github.com/PrestaShop/PrestaShop/pull/29081) we need that.
| BC breaks?        |  no
| Deprecations?     | no
| Fixed ticket?     | Fixes [27313](https://github.com/PrestaShop/PrestaShop/issues/27313).
| How to test?      | Run upgrade and check if table stock_mvt had the column physical_quantity changed to UNSIGNED.


jolelievre (commented):
Thanks @margud

jolelievre (approved):

jolelievre (commented):
Hi @margud 

this one needs to be rebased because conflicts arose when another one was merged.
Don't hesitate to add you change in the middle of the file instead of the bottom, this should reduce the risk for future conflicts 😉

Robin-Fischer-PS:
Hi,

Once ready, this one can be tested by a dev :)

jolelievre (changes requested):
As I suggested in the related PR, this should be handled in v8.1 instead, so the modification should be moved into the `8.1.0.sql` file instead

PR #497, opened 2022-08-16
Fix version compare check to use CoreUpgrader 8 with 8.0 version

FabienPapet:
This PR fixes CoreUpgrader file

If you want to update to version 8.0, the core upgrader uses the CoreUpgrader17 instead of CoreUpgrader8.

(the selected core upgrader depends on the targeted version and not from the source version) 

Fixes https://github.com/PrestaShop/PrestaShop/issues/29248

## HOW TO TEST:

1/ In the autoupgrade form, choose upgrade via local archive
2/ In the field for the upgrade version, you should only be able to enter versions with a patch precision, example:

- 1.7.7.8
- 8.0.0
- 1.6.1.24

Examples of wrong versions (you will get an error message):

- 1.7.2
- 8.0
- 8.0.0.0

(Notice how for PS 8 and above, the version is made of 3 numbers, while it's 4 numbers for older versions)

### DIFF
--- a/js/upgrade.js
+++ b/js/upgrade.js
@@ -804,7 +804,7 @@ $(document).ready(function() {
         var archive_prestashop = $("select[name=archive_prestashop]").val();
         var archive_num = $("input[name=archive_num]").val();
         var archive_xml = $("select[name=archive_xml]").val();
-        if (archive_num == "" || !archive_num.match(versionNumberRegex)) {
+        if (archive_num == "" || (archive_num !== undefined && !archive_num.match(versionNumberRegex))) {
           showConfigResult(input.translation.needToEnterArchiveVersionNumber, "error");
           return false;
         }
@@ -820,7 +820,7 @@ $(document).ready(function() {
         params.channel = "directory";
         params.directory_prestashop = $("select[name=directory_prestashop] option:selected").val();
         let directory_num = $("input[name=directory_num]").val();
-        if (directory_num == "" || !directory_num.match(versionNumberRegex)) {
+        if (directory_num == "" || (directory_num !== undefined && !directory_num.match(versionNumberRegex))) {
           showConfigResult(input.translation.needToEnterDirectoryVersionNumber, "error");
           return false;
         }
--- a/js/upgrade.js
+++ b/js/upgrade.js
@@ -820,7 +820,7 @@ $(document).ready(function() {
         params.channel = "directory";
         params.directory_prestashop = $("select[name=directory_prestashop] option:selected").val();
         let directory_num = $("input[name=directory_num]").val();
-        if (directory_num !== undefined && (directory_num == "" || !directory_num.match(versionNumberRegex))) {
+        if (directory_num == "" || !directory_num.match(versionNumberRegex)) {
           showConfigResult(input.translation.needToEnterDirectoryVersionNumber, "error");
           return false;
         }
--- a/js/upgrade.js
+++ b/js/upgrade.js
@@ -11,7 +11,7 @@ if (typeof input === 'undefined') {
     ajaxUpgradeTabExists: true,
     currentIndex: 'page.php',
     tab: input.tab,
-    versionNumberRegex: '^(1\\.(6|7)\\.(\\d)\\.(\\d+)$)|^(8\\.(\\d).(\\d)$)',
+    versionNumberRegex: /^(1\\.(6|7)\\.(\\d)\\.(\\d+)$)|^(8\\.(\\d).(\\d)$)/,
     channel: 'major',
     translation: {
       confirmDeleteBackup: "Are you sure you want to delete this backup?",
--- a/js/upgrade.js
+++ b/js/upgrade.js
@@ -11,6 +11,7 @@ if (typeof input === 'undefined') {
     ajaxUpgradeTabExists: true,
     currentIndex: 'page.php',
     tab: input.tab,
+    versionNumberRegex: '^(1\\.(6|7)\\.(\\d)\\.(\\d+)$)|^(8\\.(\\d).(\\d)$)',
     channel: 'major',
     translation: {
       confirmDeleteBackup: "Are you sure you want to delete this backup?",
@@ -47,7 +48,7 @@ if (typeof input === 'undefined') {
       linkAndMd5CannotBeEmpty: "Link and MD5 hash cannot be empty",
       needToEnterArchiveVersionNumber: "You need to enter the version number associated with the archive.",
       noArchiveSelected: "No archive has been selected.",
-      needToEnterDirectoryVersionNumber: "You need to enter the version number associated with the directory.",
+      needToEnterDirectoryVersionNumber: "You must enter the full version number of the version you want to upgrade. The full version number can be present in the zip name (ex: 1.7.8.1, 8.0.0).",
       confirmSkipBackup: "Please confirm that you want to skip the backup.",
       confirmPreserveFileOptions: "Please confirm that you want to preserve file options.",
       lessOptions: "Less options",
@@ -801,7 +802,7 @@ $(document).ready(function() {
         var archive_prestashop = $("select[name=archive_prestashop]").val();
         var archive_num = $("input[name=archive_num]").val();
         var archive_xml = $("select[name=archive_xml]").val();
-        if (archive_num == "") {
+        if (archive_num == "" || !archive_num.match(input.versionNumberRegex)) {
           showConfigResult(input.translation.needToEnterArchiveVersionNumber, "error");
           return false;
         }
@@ -816,8 +817,8 @@ $(document).ready(function() {
       } else if ($newChannel === "directory") {
         params.channel = "directory";
         params.directory_prestashop = $("select[name=directory_prestashop] option:selected").val();
-        var directory_num = $("input[name=directory_num]").val();
-        if (directory_num == "" || directory_num.indexOf(".") == -1) {
+        let directory_num = $("input[name=directory_num]").val();
+        if (directory_num == "" || !directory_num.match(input.versionNumberRegex)) {
           showConfigResult(input.translation.needToEnterDirectoryVersionNumber, "error");
           return false;
         }
--- a/js/upgrade.js
+++ b/js/upgrade.js
@@ -11,7 +11,6 @@ if (typeof input === 'undefined') {
     ajaxUpgradeTabExists: true,
     currentIndex: 'page.php',
     tab: input.tab,
-    versionNumberRegex: /^(1\\.(6|7)\\.(\\d)\\.(\\d+)$)|^(8\\.(\\d).(\\d)$)/,
     channel: 'major',
     translation: {
       confirmDeleteBackup: "Are you sure you want to delete this backup?",
@@ -669,7 +668,7 @@ $(document).ready(function() {
           $("#changedList").toggle();
         });
 
-        $(".toggleSublist").die().live("click", function(e) {
+        $('body').on().on("click", '.toggleSublist', function(e) {
           e.preventDefault();
           $(this).parent().next().toggle();
         });
@@ -727,7 +726,7 @@ $(document).ready(function() {
           $("#diffList").toggle();
         });
 
-        $(".toggleSublist").die().live("click", function(e) {
+        $('body').on().on("click", '.toggleSublist', function(e) {
           e.preventDefault();
           // this=a, parent=h3, next=ul
           $(this).parent().next().toggle();
@@ -767,10 +766,13 @@ $("input[name=btn_adv]").click(function(e) {
 });
 
 $(document).ready(function() {
-  $("input[name|=submitConf]").bind("click", function(e) {
+  $("input[name|=submitConf], input[name=submitConf-channel]").bind("click", function(e) {
+
     var params = {};
     var $newChannel = $("select[name=channel] option:selected").val();
     var $oldChannel = $("select[name=channel] option.current").val();
+    var versionNumberRegex = /^(1\\.(6|7)\\.(\\d)\\.(\\d+)$)|^(8\\.(\\d).(\\d)$)/;
+
     $oldChannel = "";
 
     if ($oldChannel != $newChannel) {
@@ -802,7 +804,7 @@ $(document).ready(function() {
         var archive_prestashop = $("select[name=archive_prestashop]").val();
         var archive_num = $("input[name=archive_num]").val();
         var archive_xml = $("select[name=archive_xml]").val();
-        if (archive_num == "" || !archive_num.match(input.versionNumberRegex)) {
+        if (archive_num == "" || !archive_num.match(versionNumberRegex)) {
           showConfigResult(input.translation.needToEnterArchiveVersionNumber, "error");
           return false;
         }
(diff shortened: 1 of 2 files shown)


FabienPapet:
php > var_dump(version_compare("8.0", "8.0.0", '<'));
bool(true)
php > var_dump(version_compare("8.0.0", "8.0.0", '<'));
bool(false)
php > var_dump(version_compare("8", "8.0.0", '<'));
bool(true)

jolelievre (dismissed):
Thanks @FabienPapet

matthieu-rolland (approved):

matthieu-rolland:
@PrestaShop/qa-team the bug happened when upgrading via local archive and putting 8.0 instead of 8.0.0, now it should work even if you put `8.0` as the target version

matthieu-rolland (dismissed):

js/upgrade.js
```diff
@@ -817,7 +817,7 @@ $(document).ready(function() {
         params.channel = "directory";
         params.directory_prestashop = $("select[name=directory_prestashop] option:selected").val();
         var directory_num = $("input[name=directory_num]").val();
-        if (directory_num == "" || directory_num.indexOf(".") == -1) {
+        if (directory_num == "" || !directory_num.match('^[0-9]{1}\\.[0-9]{1}\\.[0-9]{1}(\\.[0-9]{1})?$')) {
```
jolelievre, line 820: Hi @FabienPapet and @MatShir 

The idea to force a valid version is a good one, however the test right doesn't seem strict enough. We actually need the whole version including the patch number because our upgrade scripts can vary depending on the patch target (for example we have a `1.7.8.1.sql` file to fix some bugs in the database).

But to apply the upgrade script correctly we really need the whole version which is not formatted the same way starting 8.0. So the regexp (or any other check) should take this into account and validate the input differently for vers 1.6.*, 1.7.* and 8.0.*

jolelievre (changes requested):
The principle is good but the rule is not accurate enough

js/upgrade.js
jolelievre: Just an exmaple to illustrate more the problem With the current regexp the last part is optional, meaning if you input `1.7.5` it is considered valid and yet you don't have the patch version which will result in errors during the upgrade

matthieu-rolland:
@FabienPapet I added a commit with some wording modification

About the regex: did you test it in the form ? I tried your PR and it seems like I can put any value I want in the field, I don't think it comes from your regex, more like it's not even taken into account, but maybe I'm missing something ? 🤔 

the regex seems good otherwise 👍

jolelievre (dismissed):

matthieu-rolland (approved):

HanaRebaiQA (commented):
Hello @matthieu-rolland & @FabienPapet 

I have checked this PR and as discussed on Slack, here is the detected issue when started the test and after this [commit](https://github.com/PrestaShop/autoupgrade/pull/497/commits/75cd544090848599dd4252ad9b2979c33e988174) :

1. Install PS1778 OR PS1786
2. Uninstall problematic modules (welcome, ps_mbo, pd_metrics, facebook, psxmarketingwithgoogle)
3. Install the `autoupgrade v4.15` module from dev branch
4. Install a new 8.0.x branch
5. Create a release from 8.0.0
6. Configure the autoupgrade module
7. Use a `8.0` as a version 
8. Upgrade to 8.0.0 branch
9. Connect to the shop 8.0.0
10. See error > An exception is thrown and impossible to connect to the shop


image

**PS1786 :** 

https://user-images.githubusercontent.com/92912932/185649859-62f34093-9d13-4bb9-aebb-6db46429ce09.mp4

**PS1778 :** 

https://user-images.githubusercontent.com/92912932/185649930-3d18bba6-7831-4c58-971b-e4619b06a0ad.mp4

I found another issues which are not related to this PR : [#29394](https://github.com/PrestaShop/PrestaShop/issues/29394
), [#29395](https://github.com/PrestaShop/PrestaShop/issues/29395), [#29398](https://github.com/PrestaShop/PrestaShop/issues/29398), [#29399](https://github.com/PrestaShop/PrestaShop/issues/29399)

The test phase is still in progress and i didn't check again following the [last commit](https://github.com/PrestaShop/autoupgrade/pull/497/commits/b14310d34efe9363ea07b22557d7c176801bd1bd).

We will give you our next feedback after following our tests.

Thank You!

khouloudbelguith (commented):
Hello @FabienPapet,

- I tried an upgrade from 16.1.20 => **1.6.2** (local archive = 1.6.1.24): I don't have any error, => NOK as the expected behavior "I should have an error message that the version is NOK": after the upgrade, the version displayed is **1.6.2**

https://user-images.githubusercontent.com/16067358/185946604-f64eaf48-4fce-4ede-b21c-42969dae3545.mp4

- I tried an upgrade from 1.6.1.20 => **1.7.4** (local archive = 1.7.4.4): I don't have any error, => NOK as the expected behavior "I should have an error message that the version is NOK": after the upgrade, the version displayed **1.7.4.4**

https://user-images.githubusercontent.com/16067358/185946738-9156c655-4b5e-45c6-a2e0-ebc9099d092f.mp4

- I tried an upgrade from 1.7.8.7 => **8.0.0.0** (local archive = 8.0.0)
I don't have any error, => NOK as the expected behavior "I should have an error message that the version is NOK": after the upgrade, the version displayed is 8.0.0 

https://user-images.githubusercontent.com/16067358/185947287-bde83e97-69dc-4a0b-9735-d8c6c443cdea.mp4

- I tried an upgrade from 1.7.8.7 => **8** (local archive = 8.0.0)
I don't have any error, => NOK as the expected behavior "I should have an error message that the version is NOK".
The upgrade process is stopped 

https://user-images.githubusercontent.com/16067358/185947484-c05df361-8adf-4fcd-ba03-013d85a0ee2c.mp4

- I tried an upgrade from 1.7.8.7 => **8.0** (local archive = 8.0.0)
I don't have any error, => NOK as the expected behavior "I should have an error message that the version is NOK".
After the upgrade, I cannot connect to BO, I cannot sign in to the FO, I cannot create a new account from the FO, I cannot add a product to a cart from the FO 

https://user-images.githubusercontent.com/16067358/185948039-8a463e10-9459-4694-baba-fecb39544b73.mp4

I enabled the debug mode, I have an error when trying to connect to BO


- I tried an upgrade from 1.7.7.8 => **1.7.8** (local archive = 1.7.8.7)
I don't have any error, => NOK as the expected behavior "I should have an error message that the version is NOK".
After the upgrade, I enabled debug mode, I have an exception in the BO and in the FO 

https://user-images.githubusercontent.com/16067358/185949489-ac3718ad-e7ce-4ab6-8634-cf84f82b50a4.mp4

- I tried an upgrade from 1.7.7.8 => **8,0,0** (local archive = 8.0.0)
I don't have any error, => NOK as the expected behavior "I should have an error message that the version is NOK".
The upgrade process is stopped 

https://user-images.githubusercontent.com/16067358/185956384-eb112254-1021-4a3d-8abb-f92fcc3c140c.mp4

Thanks!

MatShir:
@FabienPapet small reminder so the PR can be ready for the auto-upgrade release :)

FabienPapet:
Should be good now

js/upgrade.js
```diff
@@ -816,9 +817,10 @@ $(document).ready(function() {
       } else if ($newChannel === "directory") {
         params.channel = "directory";
         params.directory_prestashop = $("select[name=directory_prestashop] option:selected").val();
-        var directory_num = $("input[name=directory_num]").val();
-        if (directory_num == "" || directory_num.indexOf(".") == -1) {
+        let directory_num = $("input[name=directory_num]").val();
+        if (directory_num == "" || !directory_num.match(input.versionNumberRegex)) {
           showConfigResult(input.translation.needToEnterDirectoryVersionNumber, "error");
+          console.log('test');
```
atomiix, line 823: I think you forgot this 😄
```suggestion
```

js/upgrade.js
```diff
@@ -11,6 +11,7 @@ if (typeof input === 'undefined') {
     ajaxUpgradeTabExists: true,
     currentIndex: 'page.php',
     tab: input.tab,
+    versionNumberRegex: '^(1\\.(6|7)\\.(\\d)\\.(\\d+)$)|^(8\\.(\\d).(\\d)$)',
```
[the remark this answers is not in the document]
FabienPapet: This will be converted anyway , see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#parameters

atomiix (dismissed):

matks (dismissed):

khouloudbelguith (commented):
Hi @FabienPapet,

1. I tried an upgrade from PS1.7.7.8 to PS8.0.0 + version = **8**

- I don't have any warning before launching the upgrade
- I have an error during the upgrade process 

image

2. I tried an upgrade from PS1.7.7.8 to PS8.0.0 + version = **8.0**

- I don't have any warning before launching the upgrade
- The upgrade process is ok
- I tried to log in into BO after upgrade => error 500

I attached a screen record

https://user-images.githubusercontent.com/16067358/189078630-08502ac7-e15f-40ff-a376-f3cfb7e00573.mp4

3. I tried an upgrade from PS1.7.6.9 to PS1.7.8.7 + version = **1.7.8**

- I don't have any warning before launching the upgrade
- The upgrade process is ok
- I tried to log in to BO & FO after the upgrade => An exception is displayed after enabling debug mode

I attached a screen record

https://user-images.githubusercontent.com/16067358/189090862-4e674404-ed22-468b-8d44-7ea4ef8c8f63.mp4

@FabienPapet, could you please check, I still have the same issues.

Thanks!

matthieu-rolland:
Hello @khouloudbelguith I updated the pull request, now you will have this message: 

Capture d’écran 2022-09-15 à 11 12 15

jolelievre (approved):

kpodemski (approved):

atomiix (approved):

khouloudbelguith (commented):
Hello @FabienPapet, @matthieu-rolland,

I checked with @HanaRebaiQA and @hibatallahAouadni  this PR.

**First issue**

1. Install a previous version of PrestaShop for example (PS1.6.1.24, PS1.7.8.7, PS1.7.7.8)
2. Install the module 1-click upgrade of this PR
3. Click on configure
4. Click on the button "More options"
5. Choose the local archive (do not copy-paste any zip in the folder /admin-dev/autoupgrade/download)
6. Click on Save
7. See error in the console and the page is not refreshed 
image

**With PR**

https://user-images.githubusercontent.com/16067358/191078183-31bed9dc-606a-4769-88af-3e4e9cdf284e.mp4

**Without PR**

https://user-images.githubusercontent.com/16067358/191078262-ae5da09e-8b69-4272-a3ce-89cc0d247ab9.mp4

**Second issue**

1. Install PS1.7.7.8
2. Install the 1-click upgrade of this PR & click on configure & choose "local archive"
3. Copy past the [zip](https://www.prestashop.com/fr/system/files/ps_releases/prestashop_1.7.8.7.zip?token=003dcae7f5) of 1.7.8.7 into /admin-dev/autoupgrade/download
4. Select the archive to use
5. Set an invalid number in the Input "Number of the version you want to upgrade to: *" for example "1.7.8 
image
6. Click on Save => The alert is well displayed
7. Set a valid number in the Input "Number of the version you want to upgrade to: *" for example "1.7.8.7 & click on Save
8. Do the upgrade
9. After the upgrade > Uninstall all problematic modules (ps_mbo, welcome, ps_mbo, ..)
10. Create a release from 8.0.x and copy paste the zip into /admin-dev/autoupgrade/download
11. Select the archive to use "8.0.0"
12. Set an invalid number in the Input "Number of the version you want to upgrade to: *" for example " 8.0
13. Click on Save => The alert is well displayed
14. Set a valid number in the Input "Number of the version you want to upgrade to: *" for example " 8.0.0 & click on Save
15. Do the upgrade
16. After the upgrade > go BO > See error 500
17. Enable debug mode > See exception 
image

**Without PR** => OK

https://user-images.githubusercontent.com/16067358/191079640-a6b941a3-4d35-4fc6-a41e-8eebd945b79d.mp4

Thanks @HanaRebaiQA @hibatallahAouadni for the check :heart: 

Thank you!

matthieu-rolland:
thank you @khouloudbelguith 

I fixed your first issue :+1: 

About the second one:

- I can't reproduce it following your steps, I guess something else went wrong in your case but I'm not sure what, we should definitely investigate
- I don't think it's linked to this PR (this PR only modifies JS, it can't break the upgrade process... and I didn't reproduce with my PR)

Do you think we can create an issue for your second point, with reproducible steps, then it would be fixed in another PR

js/upgrade.js
```diff
@@ -816,8 +819,8 @@ $(document).ready(function() {
       } else if ($newChannel === "directory") {
         params.channel = "directory";
         params.directory_prestashop = $("select[name=directory_prestashop] option:selected").val();
-        var directory_num = $("input[name=directory_num]").val();
-        if (directory_num == "" || directory_num.indexOf(".") == -1) {
+        let directory_num = $("input[name=directory_num]").val();
+        if (directory_num == "" || (directory_num !== undefined && !directory_num.match(versionNumberRegex))) {
```
atomiix, line 823: Same here?
```suggestion
        if (directory_num == "" || directory_num === undefined || !directory_num.match(versionNumberRegex)) {
```

Robin-Fischer-PS:
Hello @FabienPapet @matthieu-rolland @matks !

As seen with @khouloudbelguith , since the second issue cannot be reproduced by everyone and It seems that it's more subtle than we thought (yeah, 17 steps where not enough).

So we will create an issue for this second issue and investigate on it later. So we only need the first issue to be fixed and we will test it to check the fix, if it's OK, then merge.

js/upgrade.js
```diff
@@ -801,7 +804,7 @@ $(document).ready(function() {
         var archive_prestashop = $("select[name=archive_prestashop]").val();
         var archive_num = $("input[name=archive_num]").val();
         var archive_xml = $("select[name=archive_xml]").val();
-        if (archive_num == "") {
+        if (archive_num == "" || (archive_num !== undefined && !archive_num.match(versionNumberRegex))) {
```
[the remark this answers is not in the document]
matthieu-rolland: actually no, but I understand your confusion:

this is for the case where the archive num field is not present, in this case we don't want to try the match, because the value is undefined (before adding this check, js would break because we tried a match on an undefined)

matks (approved):

atomiix (approved):

HanaRebaiQA (approved):
Hello @FabienPapet & @matthieu-rolland 

I checked with @khouloudbelguith 


- Upgrade from PS16120 => PS16124 : OK 
- Upgrade from PS16124 => PS1778: OK 
- Upgrade from PS1778 => PS1787 : OK
- Upgrade from PS1787 => 8.0.0 : OK
- Rollback from PS8.0.0 => PS1787 : OK
- Upgrade from PS1778 => 8.0.0 : OK
- Rollback from PS8.0.0 => PS1778 : OK

- Check invalid versions : OK
- Check valid versions : OK 

So, QA :heavy_check_mark: 

Thank you!

matks:
Thank you everybody

PR #503, opened 2022-09-06
add sql queries for removing tabs

matthieu-rolland:
| Questions         | Answers
| Description?      | Theme and module catalog tab links have been removed in 8.0.0, autoupgrade should take that into account
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes [#29394](https://github.com/PrestaShop/PrestaShop/issues/29394)
| How to test?      | See how to test in issue [#29394](https://github.com/PrestaShop/PrestaShop/issues/29394)
| Possible impacts? |

### DIFF
--- a/upgrade/sql/8.0.0.sql
+++ b/upgrade/sql/8.0.0.sql
@@ -5,9 +5,13 @@ DROP TABLE IF EXISTS `PREFIX_referrer`;
 DROP TABLE IF EXISTS `PREFIX_referrer_cache`;
 DROP TABLE IF EXISTS `PREFIX_referrer_shop`;
 
-/* Remove page Referrers */
+/* Remove page Referrers, theme catalog and module catalog */
 ## Remove Tabs
 DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminReferrers';
+DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminThemesCatalog';
+DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminModulesCatalog';
+DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminAddonsCatalog';
+DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminParentModulesCatalog';
 DELETE FROM `PREFIX_tab_lang` WHERE `id_tab` NOT IN (SELECT id_tab FROM `PREFIX_tab`);
 ## Remove Roles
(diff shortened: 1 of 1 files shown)


upgrade/sql/8.0.0.sql
```diff
@@ -5,9 +5,13 @@ DROP TABLE IF EXISTS `PREFIX_referrer`;
 DROP TABLE IF EXISTS `PREFIX_referrer_cache`;
 DROP TABLE IF EXISTS `PREFIX_referrer_shop`;
 
-/* Remove page Referrers */
+/* Remove page Referrers, theme catalog and module catalog */
 ## Remove Tabs
 DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminReferrers';
+DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminThemesCatalog';
```
jolelievre, line 11: I'm afraid this is not enough and will likely generate bugs that the QA will report, we already remove a Tab in 8.0 (for the AdminReferrers controller) and as you can see it implies more than just cleaning the tab table:
- clean tab for the controller, along with its associated tab_lang translations (can be done via a generic query that removes orphan translations as you can see below)
- remove roles related to the controller
- optionally if the removed tab's parent now contains no children it should also be removed
- remove potential quick_access link redirecting to the removed Controller, and also clean its related quick_access_lang translations (again this can be done with a generic query removing orphan translations)

These are all bugs that were discovered while testing the previous PR.

jolelievre: This should all be done in a function `function removeControllerTab(string $legacyControllerName-: void` that can then be used inside any SQL upgrade script. It was suggested when @Progi1984 removed the AdminReferrers tab but couldn't be implemented by lack of time and need (only one removed tab at the time).

Now that there are three more tabs to remove the use of a function is not only nitpicking it's a matter of factorizing the effort of this work and also making sure nothing that should be removed is forgotten, thus reducing (in theory) the QA time later.

jolelievre (changes requested):
Thanks @matthieu-rolland 

but this should be done with a function that also needs to delete more than just the tab table.

upgrade/sql/8.0.0.sql
matthieu-rolland: ok :+1: 

would you have a link to the previous PR you're talking about ?

matks: I think it was a suggestion of... me 😋 
https://github.com/PrestaShop/autoupgrade/pull/466#pullrequestreview-990807009

PR #509, opened 2022-09-23
Add domain in all trans calls in twig templates

jolelievre:
| Questions         | Answers
| Description?      | Add domain in all trans calls in twig templates, the `Modules.Autoupgrade.Admin` was missing thus, the wordings could not be scanned and be included in the PrestaShop catalog
| Type?             | bug fix
| BC breaks?        | yes
| Deprecations?     | no
| Fixed ticket?     | ~
| How to test?      | Check that the module is still usable, but we can't check if the translation works fine since the catalog has not been updated yet
| Possible impacts? | Previously the module relied on internal PHP trad, now it will use the catalog from the core So for a while, the translations may not be available anymore

### DIFF
--- a/views/templates/block/activityLog.twig
+++ b/views/templates/block/activityLog.twig
@@ -1,7 +1,7 @@
 <div class="bootstrap" id="activityLogBlock" style="display:none">
     <div class="panel">
         <div class="panel-heading">
-            {{ 'Activity Log'|trans }}
+            {{ 'Activity Log'|trans({}, 'Modules.Autoupgrade.Admin') }}
         </div>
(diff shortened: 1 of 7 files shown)


atomiix (approved):

eternoendless:
We can still use legacy dictionaries thanks to the legacy fallback feature. That would allow us to have the module translated for 1.7.6+

eternoendless:
Or, we can decide that newer versions of the module support 1.7.8+ and simply use xlf files

matthieu-rolland (approved):

matthieu-rolland:
thanks @jolelievre !

PR #518, opened 2022-10-25
Requires the XML file when upgrading with archive

0x346e3730:
| Questions         | Answers
| Description?      | Requires the XML file when upgrading with archive
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/29932.
| How to test?      | Follow the steps in the issue
| Possible impacts? | Autoupgrading with an archive

### DIFF
--- a/views/templates/block/upgradeButtonBlock.twig
+++ b/views/templates/block/upgradeButtonBlock.twig
@@ -117,7 +117,7 @@
                                 <label for="archive_num">{{ 'Number of the version you want to upgrade to:'|trans({}, 'Modules.Autoupgrade.Admin') }} * </label>
                                 <input type="text" size="10" id="archive_num" name="archive_num" value="{{ archiveVersionNumber }}" placeholder="1.7.0.1" />
                                 <br>
-                                <label for="archive_xml">{{ 'XML file to use:'|trans({}, 'Modules.Autoupgrade.Admin') }}</label>
+                                <label for="archive_xml">{{ 'XML file to use:'|trans({}, 'Modules.Autoupgrade.Admin') }} *</label>
                                 <select id="archive_xml" name="archive_xml">
                                     <option value="">{{ 'Choose an XML file'|trans({}, 'Modules.Autoupgrade.Admin') }}</option>
                                     {% for file in xmlFiles %}
--- a/views/templates/block/upgradeButtonBlock.twig
+++ b/views/templates/block/upgradeButtonBlock.twig
@@ -93,10 +93,10 @@
                     {{ channelInfoBlock|raw }}
                     <div id="for-usePrivate">
                         <p><label for="private_release_link">{{ 'Link:'|trans({}, 'Modules.Autoupgrade.Admin') }} *</label>
-                            <input type="text" size="50" id="private_release_link" name="private_release_link" value="{{ privateChannel.releaseLink }}">
+                            <input type="text" size="50" id="private_release_link" name="private_release_link" value="{{ privateChannel.releaseLink }}" required>
                         </p>
                         <p><label for="private_release_md5">{{ 'Hash key:'|trans({}, 'Modules.Autoupgrade.Admin') }} *</label>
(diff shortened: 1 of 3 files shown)


views/templates/block/upgradeButtonBlock.twig
```diff
@@ -117,7 +117,7 @@
                                 <label for="archive_num">{{ 'Number of the version you want to upgrade to:'|trans({}, 'Modules.Autoupgrade.Admin') }} * </label>
                                 <input type="text" size="10" id="archive_num" name="archive_num" value="{{ archiveVersionNumber }}" placeholder="1.7.0.1" />
                                 <br>
-                                <label for="archive_xml">{{ 'XML file to use:'|trans({}, 'Modules.Autoupgrade.Admin') }}</label>
+                                <label for="archive_xml">{{ 'XML file to use:'|trans({}, 'Modules.Autoupgrade.Admin') }} *</label>
```
nicosomb, line 120: Maybe you should add a `required` attribute.

0x346e3730: I haven't because I looked how it was done for other fields and it's not done, the version number L118 for example

0x346e3730: done (also added an asterisk to the archive label because it's mandatory in the js code)

nicosomb: thank you

nicosomb (approved):

mflasquin (commented):
Would adding server side verification make sense?

0x346e3730:
> Would adding server side verification make sense?

This is an expert section in the back-office for the upgrade module, I don't think it would add much value here

matks (approved):
> This is an expert section in the back-office for the upgrade module, I don't think it would add much value here

Makes sense to me: this tool is being used by people to upgrade their shops. Even if someone can bypass the JS validation using crafted HTTP requests what will he gain? Use forbidden options to destroy his shop ? 😄

atomiix (approved):

HanaRebaiQA (commented):
Hello @0x346e3730 

I have checked this PR. The issue is fixed.

Here my checked list : 

- Check the upgrade from PS1787 to PS8.0 : OK
- Check the rollback from PS8.0 to PS1787 : OK
- Check the upgrade from PS1778 to PS1787 and then to PS8.0: OK
- Check the rollback from PS8.0 to PS1778 : OK
- Check the upgrade from PS1778 to PS8.0 : OK

image

 
But i found this issue : The upgrade button is enabled when want to upgrade from PS1778 to 8.0 without putting the xml
So, the upgrade is passed but it's upgraded to PS1787 (**Major version**)

image

https://user-images.githubusercontent.com/92912932/200618487-0861fbb1-02e0-4c2a-9a7c-10c2f5580183.mp4

https://user-images.githubusercontent.com/92912932/200619048-dc201213-e6f6-4b95-8b89-b1bc1bb053ea.mp4


What do you think about this behavior?

Thank You!

0x346e3730:
I think that this is the normal behavior that the button is clickable even if no XML is selected

kpodemski:
@0x346e3730 this looks like a good improvement, will you have time to finish this PR?

0x346e3730:
> @0x346e3730 this looks like a good improvement, do you have time to finish this PR? Or maybe it is ready?

For me it is finished, QA had a question for me and I answered it, adding back the waiting for QA label I guess to unblock this

florine2623 (approved):
Hello @0x346e3730 ,

After reading the comments on the PR, I think it is fixed.
There's no more error that prevents the upgrade to be done.

So as Krystian said, it could be considered as an improvement.

It is QA ✅ 
Thanks @HanaRebaiQA for the tests 🤗

matthieu-rolland:
thank you @0x346e3730 and @florine2623 !

PR #559, opened 2023-02-06
Fix issue with bad table prefix

Prestaworks:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | I changed the prefix in the SQL query, so the extra underscore was removed.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #31219 https://github.com/PrestaShop/PrestaShop/issues/31219 
| Sponsor company   | Prestaworks AB
| How to test?      | Follow instruction in issue

PR #571, opened 2023-03-02
Fix migration issues

jocel1:
| Questions         | Answers
| Description?      | Fix various bug in migration scripts
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      | Run an autoupgrade from a 1.7 DB to 8.0.1, especially with disable modules

### DIFF
--- a/upgrade/php/ps_1770_preset_tab_enabled.php
+++ b/upgrade/php/ps_1770_preset_tab_enabled.php
@@ -40,7 +40,7 @@ function ps_1770_preset_tab_enabled()
     );
     $moduleNames = [];
     foreach ($inactiveModules as $inactiveModule) {
-        $moduleNames[] = $inactiveModule['name'];
+        $moduleNames[] = '"'.$inactiveModule['name'].'"';
     }
     if (count($moduleNames) > 0) {
         $result &= Db::getInstance()->execute(
--- a/upgrade/php/ps_1770_preset_tab_enabled.php
+++ b/upgrade/php/ps_1770_preset_tab_enabled.php
@@ -40,7 +40,7 @@ function ps_1770_preset_tab_enabled()
     );
     $moduleNames = [];
     foreach ($inactiveModules as $inactiveModule) {
-        $moduleNames[] = '"'.$inactiveModule['name'].'"';
+        $moduleNames[] = '"' . $inactiveModule['name'] . '"';
     }
     if (count($moduleNames) > 0) {
         $result &= Db::getInstance()->execute(
--- a/upgrade/sql/1.7.7.0.sql
+++ b/upgrade/sql/1.7.7.0.sql
@@ -205,6 +205,7 @@ ALTER TABLE `PREFIX_order_state` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb
 ALTER TABLE `PREFIX_order_state_lang` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 ALTER TABLE `PREFIX_pack` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 ALTER TABLE `PREFIX_page` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
+/* php:execute_sql_if_table_exists('pagenotfound', 'ALTER TABLE `PREFIX_pagenotfound` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;'); */
 ALTER TABLE `PREFIX_page_type` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 ALTER TABLE `PREFIX_page_viewed` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 ALTER TABLE `PREFIX_product` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
@@ -296,6 +297,7 @@ ALTER TABLE `PREFIX_timezone` CHANGE `name` `name` varchar(32) CHARACTER SET utf
 ALTER TABLE `PREFIX_attribute_group` CHANGE `group_type` `group_type` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL;
 ALTER TABLE `PREFIX_search_word` CHANGE `word` `word` varchar(30) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL;
 ALTER TABLE `PREFIX_meta` CHANGE `page` `page` varchar(64) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL;
+/* php:execute_sql_if_table_exists('statssearch', 'ALTER TABLE `PREFIX_statssearch` CHANGE `keywords` `keywords` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL;'); */
 ALTER TABLE `PREFIX_stock` CHANGE `reference` `reference` varchar(64) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL;
 ALTER TABLE `PREFIX_stock` CHANGE `ean13` `ean13` varchar(13) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NULL;
 ALTER TABLE `PREFIX_stock` CHANGE `isbn` `isbn` varchar(32) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NULL;
--- a/upgrade/sql/1.7.7.0.sql
+++ b/upgrade/sql/1.7.7.0.sql
@@ -205,7 +205,6 @@ ALTER TABLE `PREFIX_order_state` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb
 ALTER TABLE `PREFIX_order_state_lang` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 ALTER TABLE `PREFIX_pack` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 ALTER TABLE `PREFIX_page` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
-ALTER TABLE `PREFIX_pagenotfound` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 ALTER TABLE `PREFIX_page_type` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 ALTER TABLE `PREFIX_page_viewed` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
 ALTER TABLE `PREFIX_product` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;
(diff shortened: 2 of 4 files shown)


upgrade/php/ps_remove_controller_tab.php
```diff
@@ -28,7 +28,7 @@ function ps_remove_controller_tab($className, $quickAccessUrls = [])
     // delete orphan tab langs
     Db::getInstance()->execute(
         sprintf(
-            'DELETE FROM `%stab_lang` WHERE `id_tab` NOT IN (SELECT id_tab FROM `%s_tab`)',
+            'DELETE FROM `%stab_lang` WHERE `id_tab` NOT IN (SELECT id_tab FROM `%stab`)',
```
nicosomb, line 31: About this bug, there is already an opened PR from @Prestaworks  https://github.com/PrestaShop/autoupgrade/pull/559

jocel1: Yes it's a full pack with all the currently detected issue using a DB dump from a pretashop 1.7.4.x :)

SharakPL:
Can you describe the bugs this PR tries to fix or point to the issues on PrestaShop/prestashop?

jocel1:
Well the bugs are describe in the description of each commit:

- on ps_1770_preset_tab_enabled if you have inactive module, it will create a SQL query with missing quote around the module name, which will make the migration fails
e.g.
```PrestaShopException^ {#25197
  #message: "SQLSTATE[42S22]: Column not found: 1054 Unknown column 'ps_categorytree' in 'where clause'"
  #code: 42
  #file: "./classes/db/DbPDO.php"
  #line: 150
  -previous: PDOException {#17082
    #message: "SQLSTATE[42S22]: Column not found: 1054 Unknown column 'ps_categorytree' in 'where clause'"
    #code: "42S22"
    #file: "./classes/db/DbPDO.php"
    #line: 148
    +errorInfo: array:3 [
      0 => "42S22"
      1 => 1054
      2 => "Unknown column 'ps_categorytree' in 'where clause'"
    ]
    trace: {
      ./classes/db/DbPDO.php:148 {
        DbPDOCore->_query($sql)^
        › try {
        ›     return $this->link->query($sql);
        › } catch (\PDOException $exception) {
        arguments: {
          $query: "UPDATE `dk_tab` SET `enabled` = 0 WHERE `module` IN (ps_categorytree,ps_checkpayment,ps_customeraccountlinks,ps_customersignin,ps_facetedsearch,ps_imageslider,ps_languageselector,ps_linklist,ps_mainmenu,ps_sharebuttons,ps_shoppingcart,ps_socialfollow,psgdpr,pscleaner)"
        }
      }
      ./classes/db/Db.php:377 { …}
      ./classes/db/Db.php:567 { …}
      ./modules/autoupgrade/upgrade/php/ps_1770_preset_tab_enabled.php:47 { …}
      ps_1770_preset_tab_enabled() {}
```
- on ps_remove_controller_tab.php:
```PrestaShopException^ {#25197
  #message: "SQLSTATE[42S02]: Base table or view not found: 1146 Table 'oneshop.dk__tab' doesn't exist"
  #code: 42
  #file: "./classes/db/DbPDO.php"
  #line: 150
  -previous: PDOException {#17082
    #message: "SQLSTATE[42S02]: Base table or view not found: 1146 Table 'oneshop.dk__tab' doesn't exist"
    #code: "42S02"
    #file: "./classes/db/DbPDO.php"
    #line: 148
    +errorInfo: array:3 [
      0 => "42S02"
      1 => 1146
      2 => "Table 'oneshop.dk__tab' doesn't exist"
    ]
    trace: {
      ./classes/db/DbPDO.php:148 {
        DbPDOCore->_query($sql)^
        › try {
        ›     return $this->link->query($sql);
        › } catch (\PDOException $exception) {
        arguments: {
          $query: "DELETE FROM `dk_tab_lang` WHERE `id_tab` NOT IN (SELECT id_tab FROM `dk__tab`)"
        }
      }
      ./classes/db/Db.php:377 { …}
      ./classes/db/Db.php:567 { …}
      ./modules/autoupgrade/upgrade/php/ps_remove_controller_tab.php:33 { …}
```
- on 8.0.0.sql:
```PrestaShopException^ {#25197
  #message: "SQLSTATE[42S22]: Column not found: 1054 Unknown column 'route_name' in 'field list'"
  #code: 42
  #file: "./classes/db/DbPDO.php"
  #line: 150
  -previous: PDOException {#17082
    #message: "SQLSTATE[42S22]: Column not found: 1054 Unknown column 'route_name' in 'field list'"
    #code: "42S22"
    #file: "./classes/db/DbPDO.php"
    #line: 148
    +errorInfo: array:3 [
      0 => "42S22"
      1 => 1054
      2 => "Unknown column 'route_name' in 'field list'"
    ]
    trace: {
      ./classes/db/DbPDO.php:148 {
        DbPDOCore->_query($sql)^
        › try {
        ›     return $this->link->query($sql);
        › } catch (\PDOException $exception) {
        arguments: {
          $query: "UPDATE `dk_tab` SET `active` = 1, `enabled` = 1, `wording` = 'Security', `wording_domain` = 'Admin.Navigation.Menu', `route_name` = 'admin_security' WHERE `class_name` = 'AdminSecurity'"
        }
      }
      ./classes/db/Db.php:377 { …}
      ./classes/db/Db.php:567 { …}
      ./modules/autoupgrade/upgrade/php/ps_800_add_security_tab.php:72 { …}
```
- on 1.7.7.0.sql:
- pagenotfound & statssearch are module tables, not core tables, so if at some point those modules are removed, the upgrade will fail

upgrade/sql/8.0.0.sql
```diff
@@ -192,6 +192,8 @@ ALTER TABLE `PREFIX_product_shop` MODIFY COLUMN `redirect_type` ENUM(
     '404', '410', '301-product', '302-product', '301-category', '302-category'
 ) NOT NULL DEFAULT '404';
 
+ALTER TABLE `PREFIX_tab` ADD route_name VARCHAR(256) DEFAULT NULL;
```
kpodemski, line 195: oh 🤯

jocel1: I actually try modifying the autoupgrade to run first the doctrine migration, but it doesn't work, it stop after the doctrine migration and never try to run the .sql part (I didn't have time to investigate why it's working like this)

upgrade/php/ps_1770_preset_tab_enabled.php
```diff
@@ -40,7 +40,7 @@ function ps_1770_preset_tab_enabled()
     );
     $moduleNames = [];
     foreach ($inactiveModules as $inactiveModule) {
-        $moduleNames[] = $inactiveModule['name'];
+        $moduleNames[] = '"'.$inactiveModule['name'].'"';
```
kpodemski, line 43:
```suggestion
        $moduleNames[] = '"' . $inactiveModule['name'] . '"';
```

cs fixer

kpodemski (approved):

upgrade/sql/1.7.7.0.sql
jolelievre, line 1: Why do you remove this, even in 8.0 these tables are still present so they need to be updated just like the others no?

upgrade/sql/8.0.0.sql
jolelievre: That's weird, I don't understand why you need to handle this manually if Doctrine already does it as you said 🤔 Wat is this import you are referring to?

Besides if we do need to force this manually why not I don't mind, however, this field was introduced in 1.7.7 https://github.com/PrestaShop/autoupgrade/pull/571/files so you'd need to do this in 1.7.7.0.sql file I guess

jolelievre: Oh by import you mean the tabs that are inserted in the `ps_800_add_security_tab` function maybe? I understand why this would fail then if doctrine scripts are executed after this SQL and PHP function, and it makes sense that you change this here right before the method is called But maybe out of safety this should also be done in 1.7.7.0.sql

In the future we'll need to get rid of the usage of this doctrine command, it shouldn't be used in production but only to help us build these SQL files. We'll need to go back to previous versions and run the command to integrate the changes back into these files (version by version) so we can then remove the usage of the command completely during the upgrade

upgrade/sql/1.7.7.0.sql
jocel1: because there are related to module which could have been removed. It's the responsability of the module (if present) to upgrade their own tables. In my case the customer has uninstalled those module, in the tables were not present anymore

upgrade/sql/8.0.0.sql
jocel1: @jolelievre I agree for the doctrine commands, it should be either full doctrine, or not at all, having a mix is not a good idea an actually the doctrine structure could be at the end different of what we have in the .sql
By import I mean the execution of the 8.0.0.sql file, which executes ps_800_add_security_tab function and needs this field. So the new field should have been definitely initially in the 1.7.7.0.sql file.

upgrade/sql/1.7.7.0.sql
jolelievre: Ok but as long as the module doesn't handle it we can't remove it from the upgrade module just like that or the DB will still be in a non consistent state for shops that still have the module and this could create bugs So instead we could let the responsibility here, but we need to check if the table still exists I guess maybe we need some PHP to check the existence because I believe `IF EXISTS` can't be used with `ALTER`

jolelievre (changes requested):
Hello,

this PR needs to be rebased, and the removal of alterations related to modules should be replaced with conditional queries based on the existence/absence of the targeted table (it will probably require some PHP function to do so).

upgrade/sql/1.7.7.0.sql
```diff
@@ -297,7 +296,6 @@ ALTER TABLE `PREFIX_timezone` CHANGE `name` `name` varchar(32) CHARACTER SET utf
 ALTER TABLE `PREFIX_attribute_group` CHANGE `group_type` `group_type` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL;
 ALTER TABLE `PREFIX_search_word` CHANGE `word` `word` varchar(30) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL;
 ALTER TABLE `PREFIX_meta` CHANGE `page` `page` varchar(64) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL;
-ALTER TABLE `PREFIX_statssearch` CHANGE `keywords` `keywords` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL;
```
jolelievre, line 300: We could add a PHP function for this something like:
```php
execute_sql_if_table_exists('statssearch', 'ALTER TABLE `PREFIX_statssearch` CHANGE `keywords` `keywords` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL;');
```

With a function that would look like this:
```php
function execute_sql_if_table_exists($tableToCheck, $sqlQuery): void
{
    $prefixedTable = _DB_PREFIX_ . $tableToCheck;
    // Check if table exists, if it doesn't return
    // If not execute the $sqlQuery (don't forget to replace placeholders like PREFIX
}
```

The `PrestaShopBundle\Install\SlqLoader` should be used in this version as it handles the replacement of placeholders and avoid multiple queries at the same time

jolelievre (changes requested):

boherm:
Thanks @jocel1!
After rebased your branch, I added a new commit to follow @jolelievre's comments. 😉

kpodemski:
let's wait for @jolelievre review as he requested some changes, it LGTM 👍🏻 thanks @jocel1 and @boherm 👍🏻

jolelievre (approved):

jolelievre:
Thanks @jocel1 and @boherm

sallemiines (approved):
Hello @jocel1 
Thanx for the PR 

I checked this PR on autoupgrade v4.16

Tested with an upgrade from 1.7.7.2 to 8.0.1 => OK


1-Click-Upgrade-1-Click-Upgrade-•-1772 (2)

Tested with an rollback from 8.0.1 to 1.7.7.2 => OK
1-Click-Upgrade-1-Click-Upgrade-•-1772 (3)

QA :heavy_check_mark: 
Thank you

nicosomb:
Thank you @boherm and @jocel1 !

matks:
And thank you @sallemiines for the extensive testing ;)

PR #599, opened 2023-07-19
Add key check

Hlavtox:
| Questions         | Answers
| Description?      | Adds warning about key generation in order not to crash the upgrade. Error shows up only if it fails, in order not to make the table kilometer long. :-) They key is only checked when upgrading to 8.1.0 or newer.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #https://github.com/PrestaShop/PrestaShop/issues/33286
| Sponsor company   | 
| How to test?      | See below

### How to test
- We can see that https://www.php.net/manual/en/function.openssl-pkey-new.php returns a false on failure, which is confirmed by the issue - https://github.com/PrestaShop/PrestaShop/issues/33286.
- So, to test it, put `return false;` to the beginning checkKeyGeneration method in  classes/UpgradeSelfCheck.php. You should see the error message in autoupgrade module afterwards.

### Related core PR
https://github.com/PrestaShop/PrestaShop/pull/33292

### Screenshot
Snímek obrazovky 2023-07-19 114543

### DIFF
--- a/classes/Twig/Block/UpgradeChecklist.php
+++ b/classes/Twig/Block/UpgradeChecklist.php
@@ -133,6 +133,7 @@ class UpgradeChecklist
             'checkPhpVersionCompatibility' => $this->selfCheck->isPhpVersionCompatible(),
             'checkApacheModRewrite' => $this->selfCheck->isApacheModRewriteEnabled(),
             'notLoadedPhpExtensions' => $this->selfCheck->getNotLoadedPhpExtensions(),
+            'checkKeyGeneration' => $this->selfCheck->checkKeyGeneration(),
             'checkMemoryLimit' => $this->selfCheck->isMemoryLimitValid(),
             'checkFileUploads' => $this->selfCheck->isPhpFileUploadsConfigurationEnabled(),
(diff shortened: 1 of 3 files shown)


matthieu-rolland (dismissed):
nice additional check, it's a good idea to display it only if it fails :+1:

M0rgan01 (dismissed):

kpodemski (approved):

kpodemski:
@Hlavtox could you propose a way to test it? 🤔

Hlavtox:
@kpodemski I extended the PR description, is it OK?

hibatallahAouadni:
@kpodemski do you see that a Dev should test both PRs https://github.com/PrestaShop/PrestaShop/pull/33292 and this one, or a QA may test it with a Developer's help? wdyt?

kpodemski (changes requested):
The check should be only for 8.1+

kpodemski (approved):

kpodemski (approved):

kpodemski:
I had to manually return false for the function as I don't know any broken environment and it is working fine :)

PR #604, opened 2023-07-30
Add empty values for MPN fields

Hlavtox:
| Questions         | Answers
| Description?      | When upgrading when PS versions 1.7.7 and lower, somebody forgot to set empty values after adding new columns for MPN field. This value is added, but is NULL. This was not a problem on legacy product page, but results in exception when opening products on new product page.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #https://github.com/PrestaShop/PrestaShop/issues/33493
| Sponsor company   | 
| How to test?      | Follow the steps in the issue and check that you can edit products just fine after upgrading.

### DIFF
--- a/upgrade/sql/1.7.7.0.sql
+++ b/upgrade/sql/1.7.7.0.sql
@@ -11,12 +11,17 @@ INSERT INTO `PREFIX_configuration` (`name`, `value`, `date_add`, `date_upd`) VAL
     ('PS_SEARCH_MAX_WORD_LENGTH', '15', NOW(), NOW())
 ;
 
-/* Add field MPN to tables */
+/* Add field MPN to tables and assign empty values */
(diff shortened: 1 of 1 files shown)


boherm (approved):

kpodemski (changes requested):
To be consulted with other committers.

kpodemski (approved):
Given it works for 1.7.7 upgrades only and it clearly produces fatal errors on the new product page, it seems reasonable to fix it on the upgrade module level, but the ultimate solution should be on the core side.

hibatallahAouadni (approved):
Hello @Hlavtox 

Thanks for your PR 🚀 
LGTM, QA ✅ 

image

Thanks!

PR #605, opened 2023-07-30
Normalize NULL data in database

Hlavtox:
| Questions         | Answers
| Description?      | Fixes many database incosistencies from the day when there could be NULL saved into many values.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #https://github.com/PrestaShop/PrestaShop/issues/33371, Fixes #https://github.com/PrestaShop/PrestaShop/issues/25996, Fixes #https://github.com/PrestaShop/PrestaShop/issues/22862, Fixes #https://github.com/PrestaShop/PrestaShop/issues/22386, Fixes #https://github.com/PrestaShop/PrestaShop/issues/22215
| Sponsor company   | 
| How to test?      | Run these queries in phpmyadmin and check that they work, you shouldn't really see a difference when coming from something recent. There will be some tiny changes, that's how I found https://github.com/PrestaShop/PrestaShop/issues/33492 and https://github.com/PrestaShop/PrestaShop/issues/33490. :-)

### Description
- Fixes many database incosistencies from the day when there could be NULL saved into many values in optional fields.
- There is NULL allowed for these fields in database, but there may be exception when user opens these entities.
- I spent whole day going through these columns one by one and checking what is the default value coming from the core for new values.
- Concerns only optional value, if somebody has NULL in something important like ID of something, I skipped it, that's his problem.
- We should consider setting a default value for these columns, like it is with other tables.
- Full list: [DB CHECK.xlsx](https://github.com/PrestaShop/autoupgrade/files/12209781/DB.CHECK.xlsx), my list perfectly aligns with the feedback from people - https://github.com/PrestaShop/PrestaShop/issues/22215#issuecomment-887522723


Misty-94:
@Hlavtox Hi, it fixes also https://github.com/PrestaShop/PrestaShop/issues/30301 ?

Prestaworks:
Database should not allow null if PHP do not.

Prestaworks:
was this not merged?

PR #608, opened 2023-08-01
Avoid deleting a popular payment module "stripejs"

nexustotalsolutions:
Hello,

I'm the contributor for the module "Stripejs" payment module and I've got lots of complains that the module does gets deleted while upgrading Prestashop. So I opened an issue to fix it but nothing was done after long time. https://github.com/PrestaShop/PrestaShop/issues/27561 So I decided to do it myself.

The code which I removed from this autopgrade module, is responsible to delete my module during the upgrade. 

Please consider this change at high priority because thousands of customers are constantly facing this serious issue. 
Best Regards,
Sitender

| Questions         | Answers
| Description?      | Bug fixed to stop deleting a popular Stripe payment module during PrestaShop upgrade.https://addons.prestashop.com/en/payment-card-wallet/17856-stripe-payment-pro-sca-ready.html
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #{[issue URL here](https://github.com/PrestaShop/PrestaShop/issues/27561)}
| How to test?      | Install "stripejs" => "Stripe Payment Pro" payment module. https://addons.prestashop.com/en/payment-card-wallet/17856-stripe-payment-pro-sca-ready.html and do upgrade Prestashop. It'll delete "stripejs" module. With changes, it'll not delete the module.

### DIFF
--- a/classes/TaskRunner/Upgrade/UpgradeModules.php
+++ b/classes/TaskRunner/Upgrade/UpgradeModules.php
@@ -29,9 +29,7 @@ namespace PrestaShop\Module\AutoUpgrade\TaskRunner\Upgrade;
 
 use PrestaShop\Module\AutoUpgrade\Parameters\UpgradeFileNames;
 use PrestaShop\Module\AutoUpgrade\TaskRunner\AbstractTask;
-use PrestaShop\Module\AutoUpgrade\UpgradeContainer;
 use PrestaShop\Module\AutoUpgrade\UpgradeException;
-use PrestaShop\Module\AutoUpgrade\UpgradeTools\FilesystemAdapter;
 
 /**
  * Upgrade all partners modules according to the installed prestashop version.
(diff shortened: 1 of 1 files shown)


Hlavtox (commented):
Can you also delete the whole `foreach ($modules_to_delete as $key => $module) {` block?

Hlavtox (approved):
I think we should approve this, the list should not be limited to single modules.

Instead, the autoupgrade module should get the whole module list, upgrade all modules it can and then disable all modules that do not match version compliancy.

kpodemski (approved):
I'm ok with this change as we have an option to disable non-native modules anyway.

Hlavtox (approved):

Hlavtox (approved):

Hlavtox:
@PrestaShop/qa-functional This can be merged without QA if you confirm. 

The code was hard deleting totally obsolete a list of modules - some old 1.4 modules + random problematic modules. This is no longer relevant. (@kpodemski previously approved before I fixed CS)

Robin-Fischer-PS:
Hi there,

It can be merged without manual QA test indeed ! Thanks @nexustotalsolutions for the PR, and @Hlavtox for the ping :)

PR #613, opened 2023-08-08
Update version regex in JavaScript

matthieu-rolland:
| Questions         | Answers
| Description?      | Version validation regex didn't take PS 9 into account
| Type?             | bug fix 
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | Partially Fixes https://github.com/PrestaShop/PrestaShop/issues/33578
| How to test?      | See https://github.com/PrestaShop/PrestaShop/issues/33578

### DIFF
--- a/js/upgrade.js
+++ b/js/upgrade.js
@@ -772,7 +772,7 @@ $(document).ready(function() {
     var params = {};
     var $newChannel = $("select[name=channel] option:selected").val();
     var $oldChannel = $("select[name=channel] option.current").val();
-    var versionNumberRegex = /^8\.\d+\.\d+$|^9\.\d+\.\d+$|^1\.(6|7)\.\d+\.\d+$/;
+    var versionNumberRegex = /^(8|9)\.\d+\.\d+$|^1\.(6|7)\.\d+\.\d+$/;
 
     $oldChannel = "";
 
--- a/js/upgrade.js
+++ b/js/upgrade.js
@@ -772,7 +772,7 @@ $(document).ready(function() {
     var params = {};
     var $newChannel = $("select[name=channel] option:selected").val();
     var $oldChannel = $("select[name=channel] option.current").val();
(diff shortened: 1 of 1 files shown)


matthieu-rolland:
In my opinion we want to validate only existing versions (up to 9.x.x...) but if we want to validate anything the regex would be:

`/^(8|9|\d{2,})\.\d+\.\d+$|^1\.(6|7)\.\d+\.\d+$/`

Hlavtox:
What about just allowing any version number?

^\d+(\.\d+)*$

> It accepts numbers separated by dots; it starts and ends with number; a number can have multiple digits; one number without dots is also accepted.

Hlavtox (approved):

matks (approved):

hibatallahAouadni (commented):
Hello @matthieu-rolland 

Indeed the blocked error message is gone and the upgarde is done.
But, the shop is not accessible ❌ 😞 please check the screen record below:

[Upgrade_613.webm](https://github.com/PrestaShop/autoupgrade/assets/23308427/1ea0247a-1047-4a4c-b06f-9c2fc0b1b53e)

Even when I tried to enable the debug mode, no success ❌ 

image

Please check and feedback.

Thanks!

MatShir:
@hibatallahAouadni I guess it is the expected behavior. The fix only allow to proceed to the upgrade. But Auto upgrade module is not ready yet for 9.0.0 upgrade. So I guess, we can merge it then @hibatallahAouadni wdyt?

boherm (approved):

hibatallahAouadni (approved):
Hello @matthieu-rolland 

The issue only happen when we create a release from develop branch and your PR doesn't impact the normal upgrade. LGTM, QA ✅ 

Thanks!

hibatallahAouadni:
FYI: @MatShir @matthieu-rolland the issue is created https://github.com/PrestaShop/PrestaShop/issues/33610 😉

PR #616, opened 2023-08-24
Remove 1.6 related code

Hlavtox:
| Questions         | Answers
| Description?      | Removes old 1.6 related code and upgrade scripts. Fixed return value when enableTheme failed - issue was hidden behind mixed return value.
| BC breaks?        | yes
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      | Run upgrade from some old version, like 1.7.6, should go without issues.


matks (commented):
@Hlavtox You made the CI cry 😆

darkWolf-PR:
What about to ... AT LEAST ... before removing 1.6 stuff, release a downloadable 4.14.3 version with this line:

https://github.com/PrestaShop/autoupgrade/blob/014ac0646cc3fde936d28c490eac7128a3e3bb48/classes/UpgradeSelfCheck.php#L393

changed to: 

`return true;`

So people, who actually needs to update from some older version can somehow do it? 

Because writing this: "The minimum PrestaShop compatibility version has been bumped to 1.7. If you upgrade from a version older than 1.7, please use a previous version of this module." is just plain stupid as noone is able to actually use older version because of that check.

U just have to love PS for the need to hack stuff that should work... One line that I had to find and after which I can run my update from 1.6.1.9 to 1.6.1.24.


Sorry to mess this PR, I would write it into issues, but there are no issues or I don´t see em.

matks:
> What about to ... AT LEAST ... before removing 1.6 stuff, release a downloadable 4.14.3 version with this line:
> 
> https://github.com/PrestaShop/autoupgrade/blob/014ac0646cc3fde936d28c490eac7128a3e3bb48/classes/UpgradeSelfCheck.php#L393
> 
> changed to:
> 
> `return true;`
> 
> So people, who actually needs to update from some older version can somehow do it?
> 
> Because writing this: "The minimum PrestaShop compatibility version has been bumped to 1.7. If you upgrade from a version older than 1.7, please use a previous version of this module." is just plain stupid as noone is able to actually use older version because of that check.
> 
> U just have to love PS for the need to hack stuff that should work... One line that I had to find and after which I can run my update from 1.6.1.9 to 1.6.1.24.
> 
> Sorry to mess this PR, I would write it into issues, but there are no issues or I don´t see em.

It's actually a good idea

matks:
@darkWolf-PR See https://github.com/PrestaShop/autoupgrade/pull/626

M0rgan01 (approved):

florine2623 (approved):
Hello @Hlavtox ,

Tested the following :

Upgrade went well from 1.7.6.4 to 1.7.8.0 ✅ 
Screenshot 2023-11-23 at 15 40 15


https://github.com/PrestaShop/autoupgrade/assets/16019289/5cc2c00c-2934-4438-8424-2b08d9cd06fb

Hlavtox:
Thank you @florine2623! :-)

PR #643, opened 2023-11-10
Show which file was trying to be closed when error on close

tswfi:
| Questions         | Answers
| Description?      | When backup zip file closing fails, show which zip file was trying to be closed
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      | try to fail zip close to see the new message. <br />Before it was just `ERROR - Could not close the Zip file properly. Check you are allowed to write on the disk and there is available space on it.` <br />now it is `ERROR - Could not close the Zip file: /path/to/admin-dev/autoupgrade/backup/auto-backupfiles_V8.0.5_20231110-142932-5b1cebf4.zip properly. Check you are allowed to write on the disk and there is available space on it.`

### DIFF
--- a/classes/ZipAction.php
+++ b/classes/ZipAction.php
@@ -110,8 +110,8 @@ class ZipAction
 
         if (!$zip->close()) {
             $this->logger->error($this->translator->trans(
-                'Could not close the Zip file properly. Check you are allowed to write on the disk and there is available space on it.',
-                [],
+                'Could not close the Zip file: %toFile% properly. Check you are allowed to write on the disk and there is available space on it.',
+                ['%toFile%' => $toFile],
                 'Modules.Autoupgrade.Admin'
             ));
 


tswfi:
This was actually interesting. The reason the close failed was that I had one file with `200` permissions `--w-------.`. ZipArchive tried to add it to the backup file but closing couldn't do read the file correctly so the close failed..

But anyways, it will be helpful to see the zip filename here :)

nicosomb (approved):

matthieu-rolland (approved):

boherm (approved):

AureRita:
Hi @tswfi 

Thank you for your PR, I tried to test your PR but curently I can't have the message even if I disable the right of the zip, as you can see : 
image

Could you give us an how to test that permite us to have this message ? 

Thank you

tswfi:
@AureRita try adding a file like "foobar.test" with only write permission somewhere that gets backed up. Its actually another bug (searching for the files to be backed up checks that it can see a file but it doesn't check that the file is readable). Then when the zip library tries to read the file into the zip it fails.

Also the restore step can fail if it tries to restore a file with only read permission and no write permission (for example some hoster have a "statistics" folder that gets awstats or similar generated every night and those files might be read only for the user).

AureRita:
Hi @tswfi 

Thank you for your information, yesterday I put the zip on read only, as you can see : 

image

[Untitled_ Nov 16, 2023 4_21 PM.webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/07cc74ff-a133-4c5d-9198-9bf58e77936c)

Thank you

nicosomb:
Thank you @tswfi !

PR #645, opened 2023-11-12
Add primary key to DB table accessory and fix duplicates

ShaiMagal:
| Questions         | Answers
| Description?      | Related to PR #54530 (https://github.com/PrestaShop/PrestaShop/pull/34530). Problem is, there is missing PRIMARY key, so, from older PrestaShop version there can be duplicate rows, so we must "repair" this duplicate rows and set PRIMARY key then.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #30022 (https://github.com/PrestaShop/PrestaShop/issues/30022)
| How to test?      |


Hlavtox (commented):
Although this will work perfectly, there could be a possible performance caveat with rebuilding the whole table. 🤔

I think it would be better to go for a php solution.

tleon (approved):

kpodemski:
@Hlavtox it's a valid solution

kpodemski (changes requested):
It looks good, but it should be part of 9.0.0.sql following the PR that is for the `develop` branch.

ShaiMagal:
@Hlavtox can you suggest php way of this? I am not familiar with this.

Hlavtox:
@ShaiMagal 9.0.0.sql please ;-)

Quetzacoalt91:
Closed in favor of https://github.com/PrestaShop/autoupgrade/pull/689

PR #648, opened 2023-11-12
Delete obsolete files from older versions in root folder

ShaiMagal:
| Questions         | Answers
| Description?      | Delete obsolete files from root folder - footer.php, header.php, images.inc.php.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      |

### DIFF
--- a/upgrade/php/ps_900_delete_obsolete_files_in_root_folder.php
+++ b/upgrade/php/ps_900_delete_obsolete_files_in_root_folder.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+function ps_900_delete_obsolete_files_in_root_folder()
+{
+    $obsolete_files = ['footer.php', 'header.php', 'images.inc.php'];
+    foreach ($obsolete_files as $file) {
+        $full_path_to_file = _PS_ROOT_DIR_ . '/' . $file;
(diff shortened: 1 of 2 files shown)


matthieu-rolland (approved):

Hlavtox (commented):
@ShaiMagal I have one remark with this... It would be better to put this fix into into the specific SQL file for version where these files disappeared.

But ➡️ these files disappeared between 1.6.1.24 and 1.7.0.0. ... and the develop branch of this module does not support upgrading from 1.6 directly anymore, the last release that did was https://github.com/PrestaShop/autoupgrade/releases/tag/v4.14.3.

Do we want to put this fix to an active branch of autoupgrade, to a "random" `9.0.0.sql` file (why not 8.0.0.sql?, why not to all of them?), when you won't encounter these files if you started on 1.7? 🤔

ShaiMagal:
@Hlavtox 
People already have, for example, PS version 8.1.2 with these files. So, giving it to 8.0.0.sql it makes no sense.

If we put it to 9.0.0, we can be sure whis files will be deleted for all people.

Of course its possible to add to earlier version as well (8.1.3+)...

But i dont know if/when this PR will be merged/allowed.

So I targeted do develop branch...

AureRita:
Hi @ShaiMagal 

Thank you for your PR, could you provide us an how to test or an issue that can give us some details on how we can validate your issue ? 

Thank you

ShaiMagal:
@AureRita

Create files footer.php, header.php, images.inc.php in root folder of PrestaShop.

Make update PS to version 9 (I don't know, if this is possible to do this yet?)

wdyt @Hlavtox ?

AureRita:
Hi @ShaiMagal 

Thank you for your details, currently we can only upgrade to 8.1.x, when we want to upgrade to develop, we have some issue on it. When we want to upgrade from 1.7.8.10 to 8.1.2, we have the files 'images.inc.php' wich is deleted, but I think this was before your PR. The 'header.php' and 'footer.php' are still here from an upgrade from 1.7.8.10 to 8.1.2. 

@PrestaShop/product-council WDYT, do we put this PR in blocked or anything else ?

matks (changes requested):
Hello guys I am inconfortable with this PR

I explain: I understand you want to take advantage of the upgrade process to clean some stuff. You're thinking "how can we get clean some shops of old obsolete files? mmm maybe we can use the fact that the upgrade runs scripts and insert a cleaning script inside the upgrade"

But I think it is not the Upgrade job to perform things like this although they are interesting. The idea to clean obsolete stuff is good, but I don't think it's at the right place.

For example if tomorrow you are in a house and there is no light, and you find out when you open the fridge it has a light inside, then it is NOT a good idea to keep the fridge open all the time because it provides you light 😋 . The real solution is to buy a light.

So about this cleaning script, I think it could be rather made available in a “Repair module” where a collection of queries and scripts would be available to run. Let's call it ps_repairmodule for the example.

And we could add a checkbox in the Upgrade module (that would be displayed only if ps_repairmodule is installed) “Do use the Repair Module to repair my shops after the upgrade” to make it automatic. But keep it as an option 😉 so people who dont have the module or dont want the cleaning scripts to be run are not bothered.

Cleaning your shops is a good thing. Upgrading your shop is a good thing. But mixing them? Ah 😄 I don't think it's a good choice and it could actually introduce nasty side-effects.

Hlavtox:
Yeah, it pretty much aligns what we talked about with @ShaiMagal 
in private. I think a proper solution would be to restore and improve [pscleaner](https://github.com/PrestaShop/pscleaner) module. Dominik offered to maintain it. :-)

We could all of these things there.

ShaiMagal:
@matks @Hlavtox ok, I am closing this PR.

I will continue with deleting this files "manually" with my script after every update , so for me it is not problem. I tried to help other people, not me. :)

PR #647, opened 2023-11-12
Clean old not existing modules in database for next major version PS 9.x

ShaiMagal:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | A lot of customers have in database tables "ps_modules" / "ps_hook_module" (and other similar tables) a lot of nonexistent modules on FTP. Then there will be this information in php error log - _PHP Deprecated:  The hook "XXX" is deprecated, please use "YYY" instead in module "ZZZ". in QQQ/classes/Hook.php on line 914_, so we should have resolved this discrepancy in the database to match the FTP content for major version PS 9.x, we would like to have this major version clean as much possible. For real examples check below. Most of problems(modules) are from old version PS 1.6 (it magically persist for unknown reasons)
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | https://www.openservis.cz/
| How to test?      | 

Real examples:
[12-Nov-2023 17:05:49 Europe/Prague] PHP Deprecated:  The hook "Header" is deprecated, please use "displayHeader" instead in module "blockbestsellers". in XXX/classes/Hook.php on line 914
[12-Nov-2023 17:05:49 Europe/Prague] PHP Deprecated:  The hook "Header" is deprecated, please use "displayHeader" instead in module "blockcontact". in XXX/classes/Hook.php on line 914
[12-Nov-2023 17:05:49 Europe/Prague] PHP Deprecated:  The hook "Header" is deprecated, please use "displayHeader" instead in module "blockcontactinfos". in XXX/classes/Hook.php on line 914
[12-Nov-2023 17:05:49 Europe/Prague] PHP Deprecated:  The hook "Header" is deprecated, please use "displayHeader" instead in module "blockmanufacturer". in XXX/classes/Hook.php on line 914
[12-Nov-2023 17:05:49 Europe/Prague] PHP Deprecated:  The hook "Header" is deprecated, please use "displayHeader" instead in module "blockmyaccount". in XXX/classes/Hook.php on line 914
[12-Nov-2023 17:05:49 Europe/Prague] PHP Deprecated:  The hook "Header" is deprecated, please use "displayHeader" instead in module "blockmyaccountfooter". in XXX/classes/Hook.php on line 914
[12-Nov-2023 17:05:49 Europe/Prague] PHP Deprecated:  The hook "Header" is deprecated, please use "displayHeader" instead in module "blocknewproducts". in XXX/classes/Hook.php on line 914
[...]

PR #650, opened 2023-11-13
Clean non existent addresses in DB table cart

ShaiMagal:
| Questions         | Answers
| Description?      | Time to time accidentaly and magically appears in DB table "cart" (column _id_address_invoice_/_id_address_delivery_) any record, what isn't exists in DB table "address" (column _id_address_). It show error 500 in section "Shopping Carts" and it's unusable becouse of error. **We should use next major version PS 9.0.0 like good starting point for cleaning old database inconsistency.**
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #34510 (and other similar issues)
| How to test?      | 

500Ka cart

### DIFF
--- a/upgrade/sql/9.0.0.sql
+++ b/upgrade/sql/9.0.0.sql
@@ -13,3 +13,6 @@ ALTER TABLE `PREFIX_feature_flag` ADD `type` VARCHAR(64) DEFAULT 'env,dotenv,db'
 
 /* Increase size of customized data - https://github.com/PrestaShop/PrestaShop/pull/31109 */
 ALTER TABLE `PREFIX_customized_data` MODIFY `value` varchar(1024) NOT NULL;
+
+DELETE FROM `PREFIX_cart` WHERE id_address_delivery NOT IN (SELECT id_address FROM `PREFIX_address`);
+DELETE FROM `PREFIX_cart` WHERE id_address_invoice NOT IN (SELECT id_address FROM `PREFIX_address`);
--- a/upgrade/sql/9.0.0.sql
+++ b/upgrade/sql/9.0.0.sql
@@ -14,5 +14,6 @@ ALTER TABLE `PREFIX_feature_flag` ADD `type` VARCHAR(64) DEFAULT 'env,dotenv,db'
 /* Increase size of customized data - https://github.com/PrestaShop/PrestaShop/pull/31109 */
 ALTER TABLE `PREFIX_customized_data` MODIFY `value` varchar(1024) NOT NULL;
 
-DELETE FROM `PREFIX_cart` WHERE id_address_delivery > 0 && id_address_delivery NOT IN (SELECT id_address FROM `PREFIX_address`);
-DELETE FROM `PREFIX_cart` WHERE id_address_invoice > 0 && id_address_invoice NOT IN (SELECT id_address FROM `PREFIX_address`);
+DELETE FROM `PREFIX_cart` WHERE `id_address_delivery` > 0 && id_address_delivery` NOT IN (SELECT `id_address` FROM `PREFIX_address`);
+DELETE FROM `PREFIX_cart` WHERE `id_address_invoice` > 0 && `id_address_invoice` NOT IN (SELECT `id_address` FROM `PREFIX_address`);
+DELETE FROM `PREFIX_cart_product` WHERE `id_cart` NOT IN (SELECT `id_cart` FROM `PREFIX_cart`);
--- a/upgrade/sql/9.0.0.sql
+++ b/upgrade/sql/9.0.0.sql
@@ -14,5 +14,5 @@ ALTER TABLE `PREFIX_feature_flag` ADD `type` VARCHAR(64) DEFAULT 'env,dotenv,db'
 /* Increase size of customized data - https://github.com/PrestaShop/PrestaShop/pull/31109 */
 ALTER TABLE `PREFIX_customized_data` MODIFY `value` varchar(1024) NOT NULL;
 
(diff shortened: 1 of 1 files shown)


upgrade/sql/9.0.0.sql
```diff
@@ -13,3 +13,7 @@ ALTER TABLE `PREFIX_feature_flag` ADD `type` VARCHAR(64) DEFAULT 'env,dotenv,db'
 
 /* Increase size of customized data - https://github.com/PrestaShop/PrestaShop/pull/31109 */
 ALTER TABLE `PREFIX_customized_data` MODIFY `value` varchar(1024) NOT NULL;
+
+DELETE FROM `PREFIX_cart` WHERE `id_address_delivery` > 0 && id_address_delivery` NOT IN (SELECT `id_address` FROM `PREFIX_address`);
```
tleon, line 17:
```suggestion
DELETE FROM `PREFIX_cart` WHERE `id_address_delivery` > 0 && `id_address_delivery` NOT IN (SELECT `id_address` FROM `PREFIX_address`);
```

I think you're missing a backquote here.

tleon (changes requested):

Hlavtox (changes requested):
As discussed privately, I think we should give this general maintenance a more indeph thought. My opinion is, that incosistencies like this happen on all versions. 👍

ShaiMagal:
@Hlavtox of course, but I think its necessary do two things:
1) fix this behaviour (fix problem itself in PS 9)
2) clean this problem for older versions when update to PS 9 (this PR).

Makes it sense?

upgrade/sql/9.0.0.sql
ShaiMagal: @tleon yea, sorry. Butter fingers :))

tleon (approved):

ShaiMagal:
@kpodemski So, we will skip this? Sad. I am trying to move PS forward from my REAL experiences from my customers.
pscleaner is independed (unsupported now) module, and we need to do this via update.
It's not good idea to tell people:
OK, use autoupgrade for upgrade and then use pscleaner for another cleaning.

I think, we will try make PS easier for end customers because then can our user base grow.

I can make pscleaner alive again, but...
If I understand it well, you suggest to make pscleaner alive again and give to this module PR #650  and PR #647 as well? Correct?

I think, it can be part of one time upgrade process for major version, than "periodicaly" pscleaner.

It can save any time for real customers.

What do others think? @tleon @M0rgan01  @Hlavtox @matks

M0rgan01 (commented):
If there is data which gives a 500 error, it must be cleaned, but I do not know if this is the role of the autoupgrade module.

matks:
> I think, it can be part of one time upgrade process for major version, than "periodicaly" pscleaner.

Hello @ShaiMagal I appreciate you try to make things a lot better but for me this is a hard no-go.

The Upgrade system and scripts is for ... upgrade 😁 . Not for fixing inconsistencies, not for optimizing users databases "by the way", not for fixing stuff "while we're at it". It looks like a seducing idea but altering the upgrade perimeter and goal is a very slippery slope. Trust us : we don't want to go there. It can open the doors to a lot of unwanted issues. And upgrade system is already fragile, not tested enough, it's not something where we want to try risky things.

We really prefer the safe way to have
- upgrade system for upgrade only
- pscleaner system for this kind of improvements

One tool, one goal, one purpose, one scope 😉 .

Also imagine if tomorrow you had `git` starting to clean your filesystem? Because they think that every time you run `git status` they could by the way change a little bit how your files are organized, make it better? Do you expect `git` to do this behind your back, without you knowing ? 😄 

Sorry for this disappointing answer. I will not approve this PR because I am really worried it introduces something that we will regret.

I consequently close this PR. Hope you won't get hard feelings, I'm not doing this against you but because I really think it's the best option 😉

ShaiMagal:
I am very sad with this decision to close this, but, up to you, I will still make this for my customers "manually", but... what other customers?

matks:
> I am very sad with this decision to close this, but, up to you, I will still make this for my customers "manually", but... what other customers? (I mean no openservis.cz customers - how they can solve this? A lot of them, doesn't know about this "problem")

I think you talk about a very good problem but this PR is the wrong solution.

If you hurt you arm, and go to see a doctor, the doctor tells you "I suggest we cut your arm: no arm, no wound". You can say the solution suggested is wrong without dismissing the problem, right?

> pscleaner is dead longer time and not supported. Why you would like leave this job for "non existent" pscleaner?
> 
> If you wish do this job pscleaner, why you cancel support for this module?

The PrestaShop project team is not here to solve every single problem of every user. Because this is an open source project: it's free. Because it's free, the project users cannot expect every single issue will be taken care of for them.

So what about other customers? They will have 2 choices:
1. Do they have technical skills? If yes they can attempt to do it by themselves, just like what @Hlavtox does everyday.
2. Do they have no technical skills but they have money? They can pay someone with technical skills to do it for them. It can mean ask for help a web agency or buy a module.

PrestaShop is an open source project. You can see it as a tool or a foundation. You can use it to build your shop but if there are missing parts _it is up to you to add them_ either yourself (option 1) or by paying someone to do it (option 2).

> If you wish do this job pscleaner, why you cancel support for this module? :) I am little confused about this decision, but I respect it of course, we are team, so there must be together target, how we could PrestaShop improve.

I suggest you read https://snarky.ca/the-social-contract-of-open-source/ or https://medium.com/swlh/free-as-in-puppy-5b7eb1bf3908 they are great explanations of what is the relationship between a project team and its users 😉

PR #651, opened 2023-11-16
Fix missing files from folders containing install

Hlavtox:
| Questions         | Answers
| Description?      | This fixes a mystery of missing files. See below.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/34385 (partially, don't close)
| Sponsor company   | 
| How to test?      | Upgrade from 1.7.6.9 to 8.1.2 and check that these files are no longer missing or non-updated.

### How I fixed this
- I extracted latest 8.1.2 code into `admin/autoupgrade/latest`.
- I opened up AdminSelfUpgrade and constructed the class for loading files to upgrade manually.
```
$fileUpgrader = new PrestaShop\Module\AutoUpgrade\TaskRunner\Upgrade\UpgradeFiles($this->upgradeContainer);
$fileUpgrader->warmUp();
```
- I dumped the result of `listFilesToUpgrade()` in that method and found out that the files I from the related issue are missing from the list due to an old forgotten condition `strpos($dir . DIRECTORY_SEPARATOR . $file, 'install') === false`.
- This condition is not needed, because all the exclusions are already handled by a much more sophisticated condition few lines above, `isFileSkipped`.
- Result - install and install-dev still skipped and other things with install in their paths are there.

Snímek obrazovky 2023-11-16 181137

### DIFF
--- a/classes/TaskRunner/Upgrade/UpgradeFiles.php
+++ b/classes/TaskRunner/Upgrade/UpgradeFiles.php
@@ -119,7 +119,7 @@ class UpgradeFiles extends AbstractTask
                 continue;
             }
             $list[] = str_replace($this->container->getProperty(UpgradeContainer::LATEST_PATH), '', $fullPath);
-            if (is_dir($fullPath) && strpos($dir . DIRECTORY_SEPARATOR . $file, 'install') === false) {
+            if (is_dir($fullPath)) {
                 $list = array_merge($list, $this->listFilesToUpgrade($fullPath));
             }
         }


ShaiMagal (approved):

tleon (approved):

M0rgan01 (approved):

Hlavtox:
@M0rgan01 This can be QAd normally, @AureRita can do it, he managed to reproduce the original issue. ;-)

boherm (approved):

matthieu-rolland (approved):

AureRita:
Hi @Hlavtox 

[Untitled_ Nov 23, 2023 4_58 PM.webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/f2c33816-4473-4002-8459-798be5a79315)

But in the Advanced Parameters > Information, in the List of changed files, there is only this two files, don't know if this issue should block this PR, I ask you at product council. 

If this one is not enought important, it's a QA Approved for me :heavy_check_mark: 

image


Thank you

Hlavtox:
@AureRita I am getting the same behavior on every upgrade.

It's because these two files are present on a clean prestashop - https://github.com/PrestaShop/PrestaShop/tree/develop/admin-dev/autoupgrade.

But the folder is deleted when uninstalling this module - https://github.com/PrestaShop/autoupgrade/blob/f36321043bc7c71ed4a6dab028cfaa84397590ba/autoupgrade.php#L112

I think a good solution would be to remove these two files from clean prestashop. This module cleaning after itself is OK. :-) You can create an issue about it if you want.

PR #652, opened 2023-11-16
Fix some files not being deleted

Hlavtox:
| Questions         | Answers
| Description?      | Solves a problem of some files not being deleted during upgrade. The common culprit was that these files had `/admin` in their paths.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/34385 (partially, don't close)
| Sponsor company   | 
| How to test?      | Upgrade from 1.7.6.9 to 8.1.2 and check that these files from the issue are gone.

### How I fixed it
- During my findings like in https://github.com/PrestaShop/autoupgrade/pull/651, I got to the step where it builds a list of files to get removed. It builds this list by comparing current version XML to a XML of the target version.
- Files that are on that list, are scheduled for deletion.
- However, I found out that because these files had universal /admin path in them, it was replacing it by our own admin folder. `$list_files_diff[$k] = str_replace('/' . 'admin', '/' . $admin_dir, $path);`
- However, it was replacing it not only in the beginning, but also in the middle of the paths, resulting in paths of some files to be changed to nonsense.

### Before
Snímek obrazovky 2023-11-16 192833

### After
after

### DIFF
--- a/classes/TaskRunner/Miscellaneous/CompareReleases.php
+++ b/classes/TaskRunner/Miscellaneous/CompareReleases.php
@@ -31,7 +31,7 @@ use PrestaShop\Module\AutoUpgrade\Parameters\UpgradeFileNames;
 use PrestaShop\Module\AutoUpgrade\TaskRunner\AbstractTask;
 
 /**
- * get the list of all modified and deleted files between current version
+ * This class gets the list of all modified and deleted files between current version
  * and target version (according to channel configuration).
  */
 class CompareReleases extends AbstractTask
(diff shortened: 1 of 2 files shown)


ShaiMagal (approved):

matthieu-rolland (approved):

ShaiMagal (approved):

AureRita:
Hi @Hlavtox 

[Untitled_ Nov 23, 2023 5_44 PM.webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/5a221592-759e-4599-9c4d-dcd0b56e0e4d)

But currently we can see that we can't go on the BO, we can go in FO. 

When we want to go again in BO to enable debug mode or anything else, we've this error : 
image



Thank you

Hlavtox:
@AureRita As discussed privately, caused by not filling a session table, fixed here - https://github.com/PrestaShop/autoupgrade/pull/655

A workaround is to clear cookies. :-) This will create a new session and the user will be able to log in. This PR is not the problem.

PR #655, opened 2023-11-23
Fill sessions with data after adding the column

Hlavtox:
| Questions         | Answers
| Description?      | The script doesn't fill new session columns with data, resulting in an error sometimes. This fills it.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #https://github.com/PrestaShop/autoupgrade/pull/652#issuecomment-1824717816
| Sponsor company   | 
| How to test?      | Upgrade from 1.7.6.9 to 8.1.2, directly after upgrade, open phpmyadmin and check that there won't be an empty `date_add` and `date_upd` in `ps_employee_session`.


ShaiMagal (approved):

Hlavtox:
@AureRita That's normal, you can be logged in from multiple browsers/IPs/whatever. Outdated sessions can be cleaned using the button in BO. :-) Thanks!

PR #656, opened 2023-11-23
Normalize NULL data in database

M0rgan01:
| Questions         | Answers
| Description?      | This PR repeats the work of the PR https://github.com/PrestaShop/autoupgrade/pull/605, adding a php function call
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #https://github.com/PrestaShop/PrestaShop/issues/33371, Fixes #https://github.com/PrestaShop/PrestaShop/issues/25996, Fixes #https://github.com/PrestaShop/PrestaShop/issues/22862, Fixes #https://github.com/PrestaShop/PrestaShop/issues/22386, Fixes #https://github.com/PrestaShop/PrestaShop/issues/22215
| Sponsor company   | -
| How to test?      | see below

To verify that the script has been executed, there are several solutions:

- if 8.1.3 is already released, perform an upgrade process from 8.0 to 8.1.3.
- if 8.1.3 is not released, copy the content of the changes from the 8.1.3 script to the 8.1.2 script and upgrade from 8.0.x to 8.1.2

There will be almost no change on a fresh install,

### DIFF
--- a/upgrade/php/update_null_values.php
+++ b/upgrade/php/update_null_values.php
@@ -1,0 +1,247 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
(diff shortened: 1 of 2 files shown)


Hlavtox:
We should also set the default values, or it can get broken again. 👍

M0rgan01:
@Hlavtox thank you, I modified accordingly

tleon (dismissed):

M0rgan01:
I'm passing the subject to draft, I have to review elements on default values

Hlavtox:
One more question - do we keep the columns NULLABLE or not? 👍

@jolelievre

M0rgan01:
@Hlavtox after reflection, we cannot modify all of the columns concerned on a core patch, it would have to be done on a major version. This would result in breaking changes.

Executing these requests does not solve the problem in any way, it is simply a mitigation of the problem while waiting for v9 and a potential (real) fix.

Hlavtox:
@M0rgan01 I tested the script and it works properly.

In order not to run the whole upgrade, I copied the method into IndexController and called it with $this->update_null_values();, all queries went OK.

However, I have one improvement I think we can do. Do you think you could add this before running the query? Maybe we will re-run it for v9 and I want to avoid some missing columns that we decide to remove in the meantime.

```php
            // Check if the table exists
            if (empty($db->executeS('SHOW TABLES LIKE "' . _DB_PREFIX_ . $tabName . '"'))) {
                continue;
            }

            // Check if the column exists
            if (empty($db->executeS('SHOW COLUMNS FROM `' . _DB_PREFIX_ . $tabName . "` WHERE Field = '" . $columnName . "'"))) {
                continue;
            }
```

M0rgan01:
@Hlavtox good idea, I added it, thanks

Hlavtox (approved):

Hlavtox:
Works perfectly.

Tried adding
```php
['webservice_account', 'dasdsadas', ''],
['dasdasd', 'dasdsadas', ''],
```
and they are skipped with no error. 👍

PR #662, opened 2024-01-06
Check if function get_magic_quotes_gpc() exists before use it

xsbr:
| Questions         | Answers
| Description?      | Check if function get_magic_quotes_gpc() exists before use it. This function was DEPRECATED in PHP 7.4 and REMOVED in PHP 8.0
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | yes
| Fixed ticket?     | Fixes #{[Prestashop/Prestashop#34988](https://github.com/PrestaShop/PrestaShop/issues/34988)}
| Sponsor company   | 
| How to test?      | Run upgrade and check if it ran correctly
--

### DIFF
--- a/classes/UpgradeTools/Translation.php
+++ b/classes/UpgradeTools/Translation.php
@@ -172,7 +172,7 @@ class Translation
         }
         fwrite($fd, "<?php\n\nglobal \$" . $var_name . ";\n\$" . $var_name . " = array();\n");
         foreach ($merge as $k => $v) {
-            if (get_magic_quotes_gpc()) {
+            if (function_exists('get_magic_quotes_gpc') && get_magic_quotes_gpc()) {
                 $v = stripslashes($v);
             }
             if ('mail' == $type) {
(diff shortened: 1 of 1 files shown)


Hlavtox (approved):

boherm (dismissed):

MatShir:
The problem does not seem to be replicable by the QA functional. @boherm or @Hlavtox, could you please provide some hints to test the solution?

boherm:
@MatShir As explain by @xsbr and the php doc (https://www.php.net/manual/en/function.get-magic-quotes-gpc.php) this function are deprecated in PHP 7.4 and REMOVED in PHP 8.0, and it's looks possible to execute this code when updating a file at his final state on PrestaShop upgrading. I think this could be tested in PHP 7.4 and PHP 8.0 with and without this PR.

Hlavtox:
@boherm @xsbr @MatShir Well, because get_magic_quotes_gpc apparently returns false since 5.4 and the minimum version is 5.6, we can just remove this part of a code and call it a day. :-)

Hlavtox (approved):

kpodemski (approved):

florine2623:
I put the label QA ✅ 
No tests are needed ^^

Hlavtox:
Thanks everyone!

PR #676, opened 2024-03-13
Move controller in proper folder to make it accessible on PS v9

Quetzacoalt91:
| Questions         | Answers
| Description?      | Fixes the loading of the upgrade module on PrestaShop v9
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/35540
| How to test?      | Open the configuration page of the module `autoupgrade`. The module should be accessible as on PS v8 and below.

image


Hlavtox (approved):

SharakPL (approved):

boherm (approved):

AureRita:
Hi @Quetzacoalt91 

Thank you for your PR, I tested you PR but currently I've this type of behavior : 

[recording(143).webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/4a9f4c45-d382-4522-b043-5cdc08e41489)

We can now access to the BO without syntax error, but we don't have the correct display of the 1-Click Upgrade. 
Moreover, we can see that we have an invalid token. It appears always when we want to configure the module. 

Waiting for your correction

Quetzacoalt91:
Indeed, pulling the latest `develop` brought the additional issues you found

Quetzacoalt91:
Isn't @ps-jarvis supposed to remove the "Waiting for author" label when the branch is updated?

Hlavtox (approved):

Hlavtox:
@Quetzacoalt91 The bots are broken... :-)

@AureRita You can re-test 👍

AureRita (approved):

Hlavtox:
@AureRita @Quetzacoalt91 The tab should be there guys, it's an issue if it's not. 🤔

AureRita:
in my opinion, yes, because it was here on all the previous version but we can take this issue on other PR.

Quetzacoalt91:
It depends if the issue happens on all versions because of this PR or only if the v9 is only impacted.

If all versions are concerned, I'll be happy to fix this in this PR.

MatShir:
On v5.0.1 and PS 8.1.3, the upgrade is well displayed in the menu. It may be a regression from 5.0.2

AureRita:
With your PR and 8.1.x, we have the autoupgrade on the side bar. 

Because this issue is not related to this PR, I think we can merge it :rocket:

PR #687, opened 2024-04-05
Fix tab visibility on PrestaShop 9

ga-devfront:
| Questions         | Answers
| Description?      | Fix tab visibility on PrestaShop 9
| Type?             | bug fix
| BC breaks?        |  no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/35726
| How to test?      | You can try to install the module on V8 or V9 all run as expected.

### DIFF
--- a/autoupgrade.php
+++ b/autoupgrade.php
@@ -80,12 +80,9 @@ class Autoupgrade extends Module
             $tab->class_name = 'AdminSelfUpgrade';
             $tab->module = 'autoupgrade';
 
-            // We have to check if 'DEFAULT' tab class name exist (introduced in V9) if not we use the old one 'AdminTools'
-            if (Tab::getIdFromClassName('DEFAULT')) {
-                $tab->id_parent = (int) Tab::getIdFromClassName('DEFAULT');
-            } else {
-                $tab->id_parent = (int) Tab::getIdFromClassName('AdminTools');
-            }
+            // We use 'DEFAULT' tab class to be compatible from v1.7 to v9 but we can that's appear in the "more" part of the menu
+            // 'DEFAULT' is a reference to deprecated tabs
+            $tab->id_parent = (int) Tab::getIdFromClassName('DEFAULT');
 
             foreach (Language::getLanguages(false) as $lang) {
                 $tab->name[(int) $lang['id_lang']] = '1-Click Upgrade';
--- a/autoupgrade.php
+++ b/autoupgrade.php
@@ -80,8 +80,7 @@ class Autoupgrade extends Module
             $tab->class_name = 'AdminSelfUpgrade';
             $tab->module = 'autoupgrade';
 
-            // We use 'DEFAULT' tab class to be compatible from v1.7 to v9 but we can that's appear in the "more" part of the menu
-            // 'DEFAULT' is a reference to deprecated tabs
+            // We use DEFAULT to add Upgrade tab as a standalone tab in the back office menu
             $tab->id_parent = (int) Tab::getIdFromClassName('DEFAULT');
 
             foreach (Language::getLanguages(false) as $lang) {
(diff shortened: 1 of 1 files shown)


M0rgan01 (dismissed):

kpodemski (changes requested):
blocking for investigation tomorrow

ga-devfront:
> blocking for investigation tomorrow

About what ?

kpodemski (commented):
Hello @ga-devfront 

The new tab system was introduced years ago, not in PrestaShop 9. This is why I blocked the PR yesterday: I find the title, commit name, and comment in the code confusing.

Could you point me to the PR from PrestaShop 9 introducing "DEFAULT" tab? Otherwise, let's rephrase everything that mentioned v9 as a reason why it didn't work 👍🏻

Quetzacoalt91:
I see what you mean @kpodemski. I think `DEFAULT` is the reference to the parent tab introduced in PrestaShop 1.7 when the sidebar has been reworked (https://github.com/PrestaShop/PrestaShop/commit/fef7eba557d6819400527b2800cce4e50ffd13ed?diff=split&w=1).

Sine PS 1.7.0.0, using AdminTools as parent triggered a fallback on Default. The method responsible of this will be removed in PrestaShop v9, explaining why the Upgrade module was not visible anymore: https://github.com/PrestaShop/PrestaShop/commit/178c4a482fac52d85404296f6e2124a8c60287b3

This breaking change can be documented as other very old modules may encounter the say issue.

autoupgrade.php
```diff
@@ -79,7 +79,14 @@ public function install()
...
             $tab->class_name = 'AdminSelfUpgrade';
             $tab->module = 'autoupgrade';
-            $tab->id_parent = (int) Tab::getIdFromClassName('AdminTools');
+
+            // We have to check if 'DEFAULT' tab class name exist (introduced in V9) if not we use the old one 'AdminTools'
+            if (Tab::getIdFromClassName('DEFAULT')) {
+                $tab->id_parent = (int) Tab::getIdFromClassName('DEFAULT');
+            } else {
+                $tab->id_parent = (int) Tab::getIdFromClassName('AdminTools');
+            }
```
Quetzacoalt91, line 88: If this version of the module is compliant with PS 1.7+ only, you may remove the `if` and only use the `DEFAULT` tab.

`DEFAULT` is a reference to deprecated tabs. It may be interesting to find a better place in the side bar to display the module in another PR.

ga-devfront: Thank you for your details. I will make the modification immediately. However if you have a better parent tab to put it in I would be happy to change it.

Quetzacoalt91: Let's stick to the fix here. Moving the menu entry to another place should be a dedicated topic :)

autoupgrade.php
```diff
@@ -79,7 +79,11 @@ public function install()
             $tab = new Tab();
             $tab->class_name = 'AdminSelfUpgrade';
             $tab->module = 'autoupgrade';
-            $tab->id_parent = (int) Tab::getIdFromClassName('AdminTools');
+
+            // We use 'DEFAULT' tab class to be compatible from v1.7 to v9 but we can that's appear in the "more" part of the menu
+            // 'DEFAULT' is a reference to deprecated tabs
```
kpodemski, line 84:
```suggestion
            // We use DEFAULT to add Upgrade tab as a standalone tab in the back office menu
```

kpodemski: @ga-devfront one comment + I changed title of the PR :)

ga-devfront: done

kpodemski (approved):

Hlavtox (approved):

AureRita (approved):
Hi @ga-devfront 

image

Tested on 8.0.5, 8.1.5 and develop 

Thank you

kpodemski:
thank you @ga-devfront

PR #689, opened 2024-04-16
Update missing SQL compare to a fresh install

M0rgan01:
| Questions         | Answers
| Description?      | Update missing SQL compare to a fresh install
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | Use https://github.com/PrestaShop/SeamlessUpgradeToolbox/tree/main/autoupgrade-scripts for test versions concerned


upgrade/sql/8.1.0.sql
```diff
@@ -58,6 +58,13 @@ ALTER TABLE `PREFIX_product_shop` MODIFY COLUMN `redirect_type` ENUM(
 UPDATE `PREFIX_product` SET `redirect_type` = 'default' WHERE `redirect_type` = '404' OR `redirect_type` = '' OR `redirect_type` IS NULL;
 UPDATE `PREFIX_product_shop` SET `redirect_type` = 'default' WHERE `redirect_type` = '404' OR `redirect_type` = '' OR `redirect_type` IS NULL;
 
+ALTER TABLE `PREFIX_product` MODIFY COLUMN `redirect_type` ENUM(
+   '404','410','301-product','302-product','301-category','302-category','200-displayed','404-displayed','410-displayed','default'
```
Hlavtox, line 62: Already present few lines above, you can modify the original one ;-)

M0rgan01: Thanks for the review :) I think we need to do this in 3 steps:
- Add the ENUMS while keeping the `''` so as not to lose the existing values.
- Update to `default` where `''` and ...
- Remove the unnecessary `''`

Hlavtox: Yes, exactly as you say. :-)

I kept the empty string in the enum, because products created pre 1.7.8.0 had it by default - https://github.com/PrestaShop/PrestaShop/blob/aab3f53c9c31d79a6c60d9744d2a175c4d6c5194/install-dev/data/db_structure.sql#L1647

And also, it was implemented in 8.1.0, so allowing that empty string was to avoid a BC break.

Hlavtox: Soooo, actually, I think we should do this in 9.0.0 and keep this like it is. 🤔

M0rgan01: Ok, well seen !

I changed the updating of data and columns in the v9 script, this way we catch up with potential insertions with empty string

upgrade/php/remove_duplicates_from_authorization_role.php
```diff
@@ -0,0 +1,55 @@
...
+        foreach ($elementsToRemoves as $elementToRemove) {
+            // We update the access table which may have used a duplicate role
+            Db::getInstance()->execute(
+                'UPDATE ' . _DB_PREFIX_ . "access SET id_authorization_role = '" . $duplicate['keep_ID'] . "' WHERE id_authorization_role = " . $elementToRemove['id_authorization_role']
+            );
+
+            // We remove the role
+            Db::getInstance()->delete(_DB_PREFIX_ . 'authorization_role', '`id_authorization_role` = ' . $elementToRemove['id_authorization_role']);
+        }
+    }
```
Quetzacoalt91, line 54: I think we can reduce the number of SQL requests and loops to run

```suggestion
    foreach ($duplicates as $duplicate) {
          // We update the access table which may have used a duplicate role
          Db::getInstance()->execute(
              'UPDATE ' . _DB_PREFIX_ . "access SET id_authorization_role = '" . $duplicate['keep_ID'] . "' WHERE slug = '" . $duplicate['slug'] 
          );

          // We remove the role
          Db::getInstance()->delete(_DB_PREFIX_ . 'authorization_role', "slug = '" . $duplicate['slug'] . "' AND id_authorization_role != " . $duplicate['keep_ID']);
    }
```

upgrade/sql/8.1.6-catchup.sql
```diff
@@ -0,0 +1,21 @@
+/* script intended for catching up with requests forgotten since 1.7 */
```
Quetzacoalt91, line 1: Is the version 8.1.6 to be released? otherwise it might be better to rename it as 8.1.5-catchup.sql

upgrade/php/remove_duplicates_from_authorization_role.php
M0rgan01: there is no slug in the access table, only an ID and an id_authorization_role

upgrade/sql/8.1.6-catchup.sql
M0rgan01: https://github.com/PrestaShop/PrestaShop/issues/35992

Quetzacoalt91 (approved):

Hlavtox (approved):

kpodemski (changes requested):
We will need to change 8.1.6 to 8.1.7.

M0rgan01:
@kpodemski Are you saying that because the merge and release of the module will come after the 8.1.6 release?

kpodemski:
@M0rgan01 I think we won't be able to release 8.1.6 with autoupgrade

Hlavtox:
@M0rgan01 Can you put the things to `8.1.7.sql` so we are consinstent? I did it also with my previous fixes. Just a normal version file.

M0rgan01:
@Hlavtox The idea is to have a specific catch-up file, to separate the "real" migrations from those that have been forgotten.

Quetzacoalt91 (dismissed):

upgrade/sql/9.0.0.sql
```diff
@@ -102,3 +102,19 @@ ALTER TABLE `PREFIX_order_detail` MODIFY COLUMN `product_ean13` VARCHAR(20);
...
+
+ALTER TABLE `PREFIX_product` MODIFY COLUMN `redirect_type` ENUM(
+    '404','410','301-product','302-product','301-category','302-category','200-displayed','404-displayed','410-displayed','default'
+    ) NOT NULL DEFAULT 'default';
+ALTER TABLE `PREFIX_product_shop` MODIFY COLUMN `redirect_type` ENUM(
+    '404','410','301-product','302-product','301-category','302-category','200-displayed','404-displayed','410-displayed','default'
+    ) NOT NULL DEFAULT 'default';
+
+ALTER TABLE `PREFIX_accessory` DROP KEY accessory_product;
+ALTER TABLE `PREFIX_accessory` ADD CONSTRAINT accessory_product PRIMARY KEY (`id_product_1`, `id_product_2`);
```
Quetzacoalt91, line 118: In https://github.com/PrestaShop/autoupgrade/pull/645, @ShaiMagal suggested additional SQL requests to remove any duplicated before creating the primary key. Do you think it should be moved here?

ga-devfront (dismissed):

Quetzacoalt91 (approved):

PR #690, opened 2024-04-17
Fix path to autoloader after moving controller

Quetzacoalt91:
| Questions         | Answers
| Description?      | Autoupgrade module fails to load following merge of https://github.com/PrestaShop/autoupgrade/pull/676. Autoloaders are not called automatically from PrestaShop in early versions of PS 1.7, and still require to be loaded manually. Since we move the controller, the path was incorrect.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Open the module configuration page on PS 1.7.3.4. You'll get an error about a class Translator undefined until you get this PR.

## Fixed issue

image


M0rgan01 (approved):

AureRita (approved):
Hi @Quetzacoalt91 

image

Tested on : 
1.7.1.0 (to see if it's really corrected )
develop ( to see if it don't do a regression on the new feature )

Thank you

boherm (approved):

PR #692, opened 2024-04-22
Attempt to replace home made files lookup with Finder

Quetzacoalt91:
| Questions         | Answers
| Description?      | This PR replaces our management of files to ignore with the Finder of Symfony
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Partially fixes https://github.com/PrestaShop/PrestaShop/issues/34048
| How to test?      | Run an upgrade until the end of the copy of files. Modules provided in the release should no have overwritten the modules currently installed (whatever their version is higher or lower). 

## Issue found with the current version
Capture d’écran du 2024-04-19 14-22-27

### DIFF
--- a/classes/TaskRunner/Upgrade/UpgradeFiles.php
+++ b/classes/TaskRunner/Upgrade/UpgradeFiles.php
@@ -31,6 +31,7 @@ use PrestaShop\Module\AutoUpgrade\Parameters\UpgradeFileNames;
 use PrestaShop\Module\AutoUpgrade\TaskRunner\AbstractTask;
 use PrestaShop\Module\AutoUpgrade\UpgradeContainer;
 use PrestaShop\Module\AutoUpgrade\UpgradeTools\FilesystemAdapter;
+use Symfony\Component\Finder\Finder;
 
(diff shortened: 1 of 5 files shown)


classes/TaskRunner/Upgrade/UpgradeFiles.php
```diff
@@ -92,43 +93,29 @@ public function run()
...
-                continue;
-            }
-            $list[] = str_replace($this->container->getProperty(UpgradeContainer::LATEST_PATH), '', $fullPath);
-            if (is_dir($fullPath)) {
-                $list = array_merge($list, $this->listFilesToUpgrade($fullPath));
-            }
-        }
+        $finder = new Finder();
+        $finder->in($pathToNewRelease)
+            ->exclude($this->container->getFileFilter()->getFilesToIgnoreOnUpgrade());
```
Quetzacoalt91, line 112: Unfortunately this list makes me ignore everyfolder that has `modules`, `themes/classic` etc even though it is found at the middle of the string.
Exclude does not handle regex patterns and using `notPath()` makes the results list empty.

PR #698, opened 2024-05-09
Remove orphan tabs following deletion of old controllers

Quetzacoalt91:
| Questions         | Answers
| Description?      | While upgrading to PrestaShop v9, remove tabs who lost their parent relation with the previous row deletion. this cleans the table and fixes an issue reported while opening the Stock page.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     |Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/35879
| How to test?      | Upgrade from PS 8 to 9, then open the Stock page. It should load without error.

### DIFF
--- a/upgrade/sql/9.0.0.sql
+++ b/upgrade/sql/9.0.0.sql
@@ -94,7 +94,9 @@ DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminStockInstantState';
 DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminStockCover';
 DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminSupplyOrders';
 DELETE FROM `PREFIX_tab` WHERE `class_name` = 'AdminStockConfiguration';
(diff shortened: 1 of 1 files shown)


Hlavtox (approved):

M0rgan01 (approved):

AureRita:
Hi @Quetzacoalt91 

Thank you for your PR, I tested it on : 

8.1.5 to 9.0.0, that works without any exception 

but when I try to upgrade from 8.0.5 to 9.0.0, that what I have : 

[recording(202).webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/ce0dc026-3a61-414a-b554-2f9c3bc3f171)

Waiting for your correction

Quetzacoalt91:
Seen with @AureRita: Because the stack trace relates to a module it is interesting to see if the error can be reproduced on the `dev` branch as well.

AureRita (approved):
Hi again, 

After a little review, I see that the issue is not due to your PR, so it's QA :heavy_check_mark:

PR #699, opened 2024-05-16
Align ignoring rules between copy and deletion of files + ignore native module only if they have they vendor/ folder

Quetzacoalt91:
| Questions         | Answers
| Description?      | This fixes an issue where modules files are unexpectly removed. It plugs XML diff to the method removing files to ignore, so the same rules during the copy of files are applied. <br><br>This also changes the way native modules are ignored. If one of them has never been modified (meaning their composer is still the one provided by the core), its content will be upgraded during the process.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/36184
| How to test?      | * Basic upgrades from 8.1.0 to 8.1.6 must upgrade files (delete + copy) in the `ps_googleanalytics` module during "Upgrade files" step.<br> * Installation of 8.1.0, deletion and reinstallation of module `ps_googleanalytics`, then upgrade to PS 8.1.6 should not delete or copy the module contents during "Upgrade files" step.

## Error currently found in the CI and fixed by this PR:

```
Fatal error: Uncaught Error: Class "PrestaShop\Module\Ps_Googleanalytics\Hooks\HookDisplayFooter" not found in /var/www/html/modules/ps_googleanalytics/ps_googleanalytics.php:132 Stack trace: #0 /var/www/html/classes/Hook.php(1043): Ps_Googleanalytics->hookDisplayBeforeBodyClosingTag(Array)
```

This is related to a module `ps_googleanalytics` provided in both packages 8.1.0 and 8.1.6. In PrestaShop releases, native module dependencies are merged in the core composer.
This can be confirmed by checking if a vendor/ folder is present in the module contents:
image

When installing PrestaShop 8.1.0 and upgrade immediately to 8.1.6, the new core dependencies are copied but:
* Some files were deleted if they were not found in the new release
* The new files of the module were not copied

These 2 issues are respectively fixed by each topic covered in the description.

### DIFF
--- a/classes/UpgradeTools/FileFilter.php
+++ b/classes/UpgradeTools/FileFilter.php
@@ -29,6 +29,7 @@ namespace PrestaShop\Module\AutoUpgrade\UpgradeTools;
 
 use DirectoryIterator;
 use PrestaShop\Module\AutoUpgrade\Parameters\UpgradeConfiguration;
+use SplFileInfo;
 
 class FileFilter
 {
@@ -160,9 +161,15 @@ class FileFilter
                 if (!$fileinfo->isDir() || $fileinfo->isDot()) {
                     continue;
                 }
-                if (in_array($fileinfo->getFilename(), $nativeModules)) {
-                    $this->excludeAbsoluteFilesFromUpgrade[] = '/modules/' . $fileinfo->getFilename();
+                if (!in_array($fileinfo->getFilename(), $nativeModules)) {
+                    continue;
                 }
+                if (!(new SplFileInfo($this->rootDir . '/modules/' . $fileinfo->getFilename() . '/vendor'))->isDir()) {
+                    // If a vendor folder is found in the module, this means it has been upgraded or manually installed
+                    // and can be ignored during the upgrade process
+                    continue;
+                }
+                $this->excludeAbsoluteFilesFromUpgrade[] = '/modules/' . $fileinfo->getFilename();
             }
         }
 
--- a/classes/TaskRunner/Miscellaneous/CheckFilesVersion.php
+++ b/classes/TaskRunner/Miscellaneous/CheckFilesVersion.php
@@ -29,6 +29,7 @@ namespace PrestaShop\Module\AutoUpgrade\TaskRunner\Miscellaneous;
 
 use PrestaShop\Module\AutoUpgrade\Parameters\UpgradeFileNames;
 use PrestaShop\Module\AutoUpgrade\TaskRunner\AbstractTask;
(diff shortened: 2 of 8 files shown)


M0rgan01 (approved):

classes/UpgradeTools/FileFilter.php
```diff
@@ -160,9 +161,15 @@ public function getFilesToIgnoreOnUpgrade()
...
                     continue;
                 }
-                if (in_array($fileinfo->getFilename(), $nativeModules)) {
-                    $this->excludeAbsoluteFilesFromUpgrade[] = '/modules/' . $fileinfo->getFilename();
+                if (!in_array($fileinfo->getFilename(), $nativeModules)) {
+                    continue;
+                }
+                if (!(new SplFileInfo($this->rootDir . '/modules/' . $fileinfo->getFilename() . '/vendor'))->isDir()) {
+                    // If a vendor folder is found in the module, this means it has been upgraded or manually installed
+                    // and can be ignored during the upgrade process
```
ga-devfront, line 169: Why ? If we don't have vendor, how the module can work properly ?

ga-devfront (approved):
All looks good for me nice refacto and improvement but I have just a question about /vendor file

classes/UpgradeTools/FileFilter.php
Quetzacoalt91: Module that are provided with a PrestaShop release have been added with composer, meaning their dependencies are not stored in their own vendor /folder, they are merged the core one.
It works because:
* The PrestaShop autoloader is already loaded and module files are found in it
* We check before trying to load the module autoloader, it may not exist https://github.com/PrestaShop/ps_googleanalytics/blob/618ade8f9a424ad31968f6367ddf3e0be8c5f879/ps_googleanalytics.php#L24-L27

ga-devfront: Okay I didn't understand that it was only for native modules. Thank you for the explanation.

AureRita (approved):
Hi @Quetzacoalt91 

Thank you for your PR, I tested it and it seems to works.

Tested on : 
1.7.8.11 to 8.1.6 
8.0.5 to 8.1.6 ( you corrected this PR : https://github.com/PrestaShop/PrestaShop/issues/36184 )
8.0.5 to 9.0.0 ( you corrected this PR :  https://github.com/PrestaShop/PrestaShop/issues/36153 )
8.1.5 to 8.1.16 
8.1.5 to 9.0.0
8.1.6 to 9.0.0

Thank you

PR #703, opened 2024-05-21
Add "Report Errors" option when upgrade failed

M0rgan01:
| Questions         | Answers
| Description?      | Added an option to send an error report if a store upgrade failed
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | -

image(1)


views/templates/block/activityLog.twig
```diff
@@ -19,6 +19,36 @@
             <div id="errorDuringUpgrade" class="col-xs-12" style="display:none;">
                 <h4>{{ 'Errors'|trans({}, 'Modules.Autoupgrade.Admin') }}</h4>
                 <div id="infoError" class="processing"></div>
+                <button id="reportErrorButton" class="btn btn-warning" style="margin-top: 10px;" data-toggle="modal" data-target="#errorModal">
```
Quetzacoalt91, line 22: Clearly not a fan of this `style` attribute, but it seems we have a Bootstrap so old we don't even have the spacing utilities available.

views/templates/block/activityLog.twig
```diff
@@ -19,6 +19,36 @@
...
+                            <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                                <span aria-hidden="true">&times;</span>
+                            </button>
+                        </div>
+                        <div class="modal-body">
+                            <input id="userEmail" type="text" class="form-control" placeholder="Email" style="display: block; margin-bottom: 10px;">
+                            <textarea id="errorDescription" rows="3" placeholder="{{ 'Additional details you may think relevant'|trans({}, 'Modules.Autoupgrade.Admin') }}"></textarea>
+                        </div>
+                        <div class="modal-footer">
+                            <button type="button" class="btn btn-secondary" data-dismiss="modal" style="text-transform: uppercase;">
```
Quetzacoalt91, line 43:
```suggestion
                            <button type="button" class="btn btn-secondary text-uppercase" data-dismiss="modal">
```

views/templates/block/activityLog.twig
```diff
@@ -19,6 +19,36 @@
...
                 <div id="infoError" class="processing"></div>
+                <button id="reportErrorButton" class="btn btn-warning" style="margin-top: 10px;" data-toggle="modal" data-target="#errorModal">
+                    {{ 'Report errors'|trans({}, 'Modules.Autoupgrade.Admin') }}
+                </button>
+            </div>
+
+            <div id="errorModal" class="modal fade" tabindex="-1" role="dialog" aria-hidden="true">
+                <div class="modal-dialog modal-dialog-centered" role="document">
+                    <div class="modal-content">
+                        <div class="modal-header" style="display: flex; justify-content: space-between; align-items: center;">
```
Quetzacoalt91, line 30: Was the default display broken?

views/templates/block/activityLog.twig
M0rgan01: I am the first to use utility classes for margins, but not available for this version

views/templates/block/activityLog.twig
M0rgan01: Sadly, not work in this version of bootstrap

views/templates/block/activityLog.twig
M0rgan01: After looking in detail, you need to put the close button before the title

Quetzacoalt91 (dismissed):

Hlavtox (dismissed):
I am not sure about this change guys. Implementing a specific (paid) error reporter into a native module?

We don't have any JS error reporting in the core itself, why in a module?

Isn't this a specific A LOT to some clients of yours?

M0rgan01:
This will not send JS errors, but the content of the "error" panel that occurred during an upgrade

Adding the tool will allow us to better understand errors only if the user wants to send a report. 

Also users of the modules will not pay anything.

Quetzacoalt91:
Hi @Hlavtox,

Now that we have a team dedicated to tackling all issues about the upgrade, we are indeed adding some error feedback form on the project. By providing a simple way to send the details to us with a click instead of a whole Github form, this should increase the number of feedback we receive, and then help prioritize the issues to work on.

As Morgan said this is a voluntary action, the error is sent to us only when the user confirms the modal. The visible inputs are optional, sending it as-is will provide basic but hopefully useful details about the error.

Hlavtox:
@Quetzacoalt91 @M0rgan01 Oh, okay, I thought that it will be an user implementation and the sentry ID will be configurable.

But that raises another questions:
- Why should a reporting to Prestashop SA be implemented in open source native module? The project was finally decoupled from SA things and now you want to add another?
- Who will have access to the data and errors reported? Will it be open to maintainer team?

Quetzacoalt91:
This tool has been proven useful on many other projects of the company, it makes sense to bring it on this project so we don't depend only on reports made on GitHub. Although I understand your point, our decisions must be cost efficient. That's why we use the existing workspace instead of creating a new one specifically for the upgrade module.

At the moment only a few people can reach the project and permissions can be customized in the future. I'm not the one who will able to say if someone may access the data or not, but it may an good occasion to try the export feature if we want to share some data about the module health.
image

M0rgan01:
I filtered "sensitive" data, like the admin path or the token. Also anonymization work was done on the logs in the PR https://github.com/PrestaShop/autoupgrade/pull/725. I also remove the WIP statue and the PR block, the one concerned questions about data collection, subject moved here https://github.com/PrestaShop/PrestaShop/discussions/36432

Quetzacoalt91 (approved):

Hlavtox:
@Quetzacoalt91 I understand what you are trying to achieve, but implementing in this in this way is breaking project neutrality.

It's the same as when I would want to implement reporting to my own custom server, from a native module.

> That's why we use the existing workspace instead of creating a new one specifically for the upgrade module.

Yes, but when your company goes bankrupt, what will happen with this account on Sentry? All data will be lost.

Quetzacoalt91:
If I understand what you mean, "implementing in this way" is related to the choice of Sentry.
When you think about it, the services currently powering the open-source repositories are already provided by PrestaShop SA. When it's not a Saas product directly, it would be the hosting for instance. We tend to reuse existing processes and solutions as they allow us to focus our knowledge on a minimal number of tools.

There are options on the Sentry interface to create GitHub issues when reports of the same problem are sent N times. 
image

With this proxy that can merge similar events to the same issue, we have an opportunity to ease the reporting of problem without spamming our GitHub issues. Unfortunately we've been unable to try it out so far, because the autoupgrade module would be the first project with issues on GitHub and the feature obviously needs to be used.
This should answer your last question as well, but I'd add this wouldn't be our main issue if this happens.

AureRita (approved):
Hi @M0rgan01 

[recording(252).webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/caf6b9c7-8e52-46a1-8cfd-5c317bfee04d)

Tested from : 
1.7.6.7 to 8.1.7
8.0.5 to 8.1.7
8.1.5 to 8.1.7
8.1.7 to 9.0.0
8.1.5 to 9.0.0
8.0.5 to 9.0.0

Thank you

PR #715, opened 2024-06-05
Add dynamic php requirements check for upgrade

M0rgan01:
| Questions         | Answers
| Description?      | Adds a new requirement, regarding  php compatibility of the target version in the upgrade process
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | see below

1. Try to update to a version with the correct PHP version. No alerts or errors should be visible

2. Try to update to a version requiring another version of PHP. An alert must be present specifying the compatibility range of the requested PHP versions.

4. Try to update to version 9, a warning should be present specifying that the PHP requirements could not be found


ps-jarvis:
This pull request seems to contain new translation strings. I have summarized them below to ease up review:

- `Admin.Actions`
    - [ ] `Go back`

(Note: this is an automated message, but answering it will reach a real human)

classes/UpgradeSelfCheck.php
```diff
@@ -318,7 +358,7 @@ public function isOkForUpgrade(): bool
             && ($this->isShopDeactivated() || $this->isLocalEnvironment())
             && $this->isCacheDisabled()
             && $this->isModuleVersionLatest()
-            && $this->isPhpVersionCompatible()
+            && $this->getPhpRequirementsState() != $this::PHP_REQUIREMENTS_INVALID
```
Quetzacoalt91, line 361:
```suggestion
            && $this->getPhpRequirementsState() !== $this::PHP_REQUIREMENTS_INVALID
```

classes/UpgradeSelfCheck.php
```diff
@@ -141,15 +135,70 @@ class UpgradeSelfCheck
...
+        ],
+        '1.7.8' => [
+            'php_min_version' => '5.6',
+            'php_max_version' => '7.4',
+        ],
+    ];
+
+    const PHP_REQUIREMENTS_INVALID = 0;
+    const PHP_REQUIREMENTS_VALID = 1;
+    const UNKNOWN_PHP_REQUIREMENTS = 2;
```
Quetzacoalt91, line 184:
```suggestion
    const PHP_REQUIREMENTS_UNKNOWN = 2;
```

This would allow PHPStan to better check the output of a method of this class.

classes/UpgradeSelfCheck.php
```diff
@@ -409,13 +440,55 @@ private function checkMaxExecutionTime(): int
...
+            }
         }
+        $currentPhpVersion = VersionUtils::getHumanReadableVersionOf(PHP_VERSION_ID);
+        $range['php_current_version'] = $currentPhpVersion;
+
+        return $this->phpCompatibilityRange = $range;
+    }
+
+    /**
+     * @return int
```
Quetzacoalt91, line 473:
```suggestion
     * @return self::PHP_REQUIREMENTS_*
```

classes/UpgradeSelfCheck.php
```diff
@@ -409,13 +440,55 @@ private function checkMaxExecutionTime(): int
...
+     */
+    public function getPhpRequirementsState()
+    {
+        $phpCompatibilityRange = $this->getPhpCompatibilityRange();
 
-        return (bool) ConfigurationTest::test_phpversion();
+        if (empty($phpCompatibilityRange)) {
+            return self::UNKNOWN_PHP_REQUIREMENTS;
```
Quetzacoalt91, line 480:
```suggestion
            return self::PHP_REQUIREMENTS_UNKNOWN;
```

classes/UpgradeSelfCheck.php
```diff
@@ -409,13 +440,55 @@ private function checkMaxExecutionTime(): int
...
+        } else {
```
Quetzacoalt91, line 481: No need of a `else` statement when you `return` in the if.

classes/DistributionApiService.php
```diff
@@ -0,0 +1,58 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ */
+
+namespace PrestaShop\Module\AutoUpgrade;
```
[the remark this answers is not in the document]
M0rgan01: I propose `services`, which may have other directories depending on the type of service if we start to have a certain number of them.

classes/UpgradeSelfCheck.php
```diff
@@ -409,13 +440,55 @@ private function checkMaxExecutionTime(): int
...
+            $versionMin = VersionUtils::getPhpVersionId($phpCompatibilityRange['php_min_version']);
+            $versionMax = VersionUtils::getPhpVersionId($phpCompatibilityRange['php_max_version']);
+            $actualVersion = VersionUtils::getPhpMajorMinorVersionId();
```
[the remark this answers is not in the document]
M0rgan01: Ok for the wording, however I do not use `$phpCompatibilityRange['php_current_version']` voluntarily, because it contains the patch versions, which poses a problem in the comparison in certain cases

classes/UpgradeSelfCheck.php
M0rgan01: maybe it's better to type the return of getPhpRequirementsState, right?

classes/UpgradeSelfCheck.php
M0rgan01: Pretty, I didn't know

classes/UpgradeSelfCheck.php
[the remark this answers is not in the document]
M0rgan01: For example if a store is in PHP 7.1.3, and the max is PHP 7.1, this will cause a problem when in reality it is good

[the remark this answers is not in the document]
M0rgan01: $phpCompatibilityRange['php_current_version'] is mainly present for the view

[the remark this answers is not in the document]
Quetzacoalt91: Yes, I agree with you, forget about the other change. That is another comment I initially wanted to put and forgot to undo the diff when I decided it wasn't relevant.

classes/Services/DistributionApiService.php
```diff
@@ -0,0 +1,58 @@
...
+class DistributionApiService
+{
+    public const API_URL = 'https://api.prestashop-project.org';
+
+    /**
+     * @throws DistributionApiException
+     */
+    public function getPhpVersionRequirements(string $targetVersion): array
+    {
+        $response = file_get_contents(self::API_URL . '/prestashop');
```
Quetzacoalt91, line 41: > Errors/Exceptions[ ¶](https://www.php.net/manual/en/function.file-get-contents.php#refsect1-function.file-get-contents-errors)
>
> An [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning) level error is generated if filename cannot be found, length is less than zero, or if seeking to the specified offset in the stream fails.
>
> When file_get_contents() is called on a directory, an [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning) level error is generated on Windows, and as of PHP 7.4 on other operating systems as well.

Do we need to make these warnings silent with `@file_get_contents`?

Quetzacoalt91 (approved):

classes/Services/DistributionApiService.php
M0rgan01: I think we can silence them

Quetzacoalt91 (dismissed):

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

[recording(238).webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/0cd40fa5-21fa-434a-904c-2f7e02f63024)

Tested from : 
1.7.1.0 to 1.7.8.11
1.7.8.11 to 8.1.6
8.0.5 to 8.1.6
8.1.5 to 8.1.6
8.0.5 to 9.0.0
8.1.5 to 9.0.0
8.1.6 to 9.0.0

Thank you

PR #716, opened 2024-06-06
Remove old controller AdminSelfUpgrade.php manually

Quetzacoalt91:
| Questions         | Answers
| Description?      | When upgrading the module to 5.0.2, the old controller we removed on Git is not deleted on the shop. It remains called first by the shop, and because it does not have some the changes on how to instanciate UpgradeSelfCheck.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/36317
| How to test?      | Upgrade from 5.0.1 to 5.0.3, accessing the configuration page must work.


M0rgan01 (approved):

kpodemski (approved):

AureRita (approved):
Hi @Quetzacoalt91 

[recording(228).webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/f59b6284-6705-4f10-90a4-dd7a83f53cc7)

Tested from : 
1.7.8.11 to 8.1.6 (major release) 
8.0.5 to 8.1.6 (major release)
8.1.5 to 8.1.6 (minor release) 
8.0.5 to 9.0.0 (local archive)
8.1.5 to 9.0.0 (local archive)
8.1.6 to 9.0.0 (local archive) 

FYI : When you install your PR with version 8.0.x or older without disable API distribution module, you'll have the version 5.0.2 for Autoupgrade 
We have the same behavior with the version 1.7.8.11 (Autoupgrade keep the version 5.0.2) as you can see : 

image

But we correctly upgrade the module. (I think it's an issue with the version supported for mbe) 

Thank you

PR #726, opened 2024-06-14
add new check between the xml and zip version for local upgrade

M0rgan01:
| Questions         | Answers
| Description?      | add new check between the xml and zip version for local upgrade
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | Use XML in another version. See alert

### DIFF
--- a/classes/TaskRunner/Miscellaneous/UpdateConfig.php
+++ b/classes/TaskRunner/Miscellaneous/UpdateConfig.php
@@ -95,12 +95,23 @@ class UpdateConfig extends AbstractTask
             }
 
             $xmlFile = $configurationData['archive_xml'];
-            if (!empty($xmlFile) && !file_exists($this->container->getProperty(UpgradeContainer::DOWNLOAD_PATH) . DIRECTORY_SEPARATOR . $xmlFile)) {
+            $fullXmlPath = $this->container->getProperty(UpgradeContainer::DOWNLOAD_PATH) . DIRECTORY_SEPARATOR . $xmlFile;
+            if (!empty($xmlFile) && !file_exists($fullXmlPath)) {
                 $this->error = true;
                 $this->logger->info($this->translator->trans('File %s does not exist. Unable to select that channel.', [$xmlFile]));
 
                 return false;
             }
+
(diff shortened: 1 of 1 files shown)


classes/TaskRunner/Miscellaneous/UpdateConfig.php
```diff
@@ -95,12 +95,23 @@ public function run()
...
+            if (!empty($xmlFile) && !file_exists($fullXmlPath)) {
                 $this->error = true;
                 $this->logger->info($this->translator->trans('File %s does not exist. Unable to select that channel.', [$xmlFile]));
 
                 return false;
             }
+
+            $xmlVersion = $this->getXmlVersion($fullXmlPath);
+
+            if ($xmlVersion != $targetVersion) {
```
Quetzacoalt91, line 108:
```suggestion
            if ($xmlVersion !== $targetVersion) {
```

ga-devfront (approved):
Same comment then @Quetzacoalt91 but for the rest I think everything is okay !

classes/TaskRunner/Miscellaneous/UpdateConfig.php
M0rgan01: extractPrestashopVersionFromZip return a string, same for getXmlVersion.

Quetzacoalt91: Than having the same type is a solid reason to have a strict check, and disallow any type casting.

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

[recording(241).webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/e943de67-422b-4fde-9515-cf9a272d245c)

Tested from : 
1.7.8.11 to 8.1.6
8.0.5 to 8.1.6
8.1.5 to 8.1.6
8.1.6 to 9.0.0
8.1.5 to 9.0.0
8.0.5 to 9.0.0

Thank you

PR #731, opened 2024-06-24
Fix 8.1.0 sql 

M0rgan01:
| Questions         | Answers
| Description?      | Fix authorized_application request
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | -

### DIFF
--- a/upgrade/sql/8.1.0.sql
+++ b/upgrade/sql/8.1.0.sql
@@ -88,7 +88,7 @@ ALTER TABLE `PREFIX_order_payment` ADD `id_employee` INT NULL AFTER `date_add`;
 CREATE TABLE `PREFIX_authorized_application`
 (
     id_authorized_application INT UNSIGNED AUTO_INCREMENT NOT NULL,
-    name                      VARCHAR(50) NOT NULL,
+    name                      VARCHAR(255) NOT NULL,
     description               LONGTEXT     NOT NULL,
     UNIQUE INDEX UNIQ_475B9BA55E237E06 (name),
     PRIMARY KEY (id_authorized_application)
--- a/upgrade/sql/8.1.1.sql
+++ b/upgrade/sql/8.1.1.sql
@@ -3,3 +3,5 @@ SET NAMES 'utf8mb4';
 
 /* We forgot to update redirect_type in 8.1.0. This updates it, if it's not done already */
 /* PHP:ps_811_update_redirect_type(); */;
+
+ALTER TABLE `PREFIX_authorized_application` CHANGE `name` `name` VARCHAR(50) NOT NULL;


upgrade/sql/8.1.8-catchup.sql
```diff
@@ -0,0 +1,3 @@
+/* script intended for catching invalid requests in previous scripts */
+
+ALTER TABLE `PREFIX_authorized_application` CHANGE `name` `name` VARCHAR(255) NOT NULL;
```
matks, line 3: Is this PR not doing what we said we would not do in https://github.com/PrestaShop/autoupgrade/pull/650#issuecomment-1914307091 ? 😄

M0rgan01: This fixes an error from a previous script, to be on par with a new installation. It's catching up

M0rgan01: the same way as was done on the 8.1.7-catchup.php script

M0rgan01: I am removing it because this table will be deleted in v9

jolelievre (approved):

upgrade/sql/8.1.8-catchup.sql
matks: If there was an issue in migration script from 8.0.0 to 8.1.0 we should fix the 8.1.0.sql migration script, not the 8.1.7 migration script.

Similarly if there was an issue in migration script from 1.7.8.11 to 1.7.8.12 we should fix the 1.7.8.12.sql migration script, not the 8.1.8 migration script, no? 😉 

I think it's similar to https://github.com/PrestaShop/autoupgrade/pull/650#issuecomment-1914307091 because you have the same thinking "well, people already did the 8.1 upgrade so if I want them to apply these changes we forgot, I'll slip them into next upgrade"

I think this is a right problem but a wrong solution 😉 . What about a different system?

An idea, to be explored: everytime we run a SQL migration script, let's store the info in MySQL. User John Doe runs 8.0.0.sql script -> we mark in a SQL table "he's run the 8.0.0.sql script". User John Doe runs 8.1.0.sql script -> we mark in a SQL table "he's run the 8.1.0.sql script". Something that Doctrine Migration does (marking which script was performed, and which was not).

Then we enable the possibility to create catch-up files 😉 and allow user to run them.

In that situation, if we find out we forgot some changes in 8.1.0 upgrade process, we put the forgotten changes into a 8.1.0-catchup1.sql file. If we find out we forgot some changes in 8.1.1 upgrade process, we put the forgotten changes into a 8.1.1-catchup1.sql file. And so on.

This way 😉 we keep the 8.1.7.sql script "clean" as it contains the changes to alter the database and data between 8.1.6 and 8.1.7 . And only that. No extra 😄 .

The catch-ups system I just suggested could also be modified to include scripts suggested by @ShaiMagal in https://github.com/PrestaShop/autoupgrade/pull/650

ga-devfront (dismissed):

AureRita:
Hi @M0rgan01 

Thank you for your PR, I tested it on 8.1.0 and it works as we can see here : 

image

But if we test it with next version like 8.1.3, it doesn't works as you can see : 

[recording(248).webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/8e580c9f-2074-4dd0-a068-74e40d416183)

With this last information I think I can't approve this PR, 

Waiting for your feedback

upgrade/sql/8.1.8-catchup.sql
M0rgan01: @matks 

We have already talked about migrations in the same spirit as that of doctrine, and this is a much better approach. However, this would require an overhaul of the current system, which is not currently planned in the current module. We therefore proposed a simple and quick solution to "catch up" on errors from previous scripts.

M0rgan01:
@AureRita In the video you use the dev branch.

@Hlavtox Indeed I removed the catch-up script, because this is a table from a beta feature (new API) which is not normally used in 8.1.0 and will be deleted in 9.0.0. Afterwards if it really worries you, I can put it back but for me its interest is very limited

Hlavtox:
@M0rgan01 If the tab is going to be removed, I think we can indeed call it a day. :-))

Quetzacoalt91 (approved):

AureRita:
@M0rgan01 In the video, at the start, without issue, I'm with the dev version but after I take your PR and I have the issue

AureRita (approved):
Hi @M0rgan01 

Thank you for your PR, I tested it and it seems to works as we expected.

Tested from : 
8.0.3 to 8.1.3 (for the validation)
1.7.8.11 to 8.1.7 (the other is for the non regression) 
8.0.5 to 8.1.7
8.1.5 to 8.1.7
8.1.7 to 9.0.0
8.1.5 to 9.0.0
8.0.5 to 9.0.0

Thank you

PR #733, opened 2024-06-25
Fix undefined method 'addPath' in TwigLoader

M0rgan01:
| Questions         | Answers
| Description?      | Fix undefined method 'addPath' in TwigLoader
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | In Prestashop 1.7.3.0 with php version < 7.1, install module with marketplace, replace the content with this PR and test configuration page. An alert should be present information that the PHP version is not compatible

### DIFF
--- a/controllers/admin/AdminSelfUpgradeController.php
+++ b/controllers/admin/AdminSelfUpgradeController.php
@@ -497,7 +497,10 @@ class AdminSelfUpgradeController extends ModuleAdminController
 
             try {
                 global $kernel;
-                $kernel->getContainer()->get('twig.loader')->addPath('../modules/autoupgrade/views/templates', 'ModuleAutoUpgrade');
+                $twigLoader = $kernel->getContainer()->get('twig.loader');
+                if (method_exists($twigLoader, 'addPath')) {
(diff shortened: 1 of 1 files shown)


Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

https://www.loom.com/share/8438499c69b34a09abef7dcd6ddba784?sid=ee351840-ccd5-436e-9fef-8e8b0ca466b8


FYI : When you want to install it or anything else from a version before php 7.2, you have this issue : 

image

But we can't be on 7.2 with this shop. 

image

From the UI . 


Tested from : 
1.7.3.0 to 1.7.8.11
1.7.8.11 to 8.1.7
8.0.5 to 8.1.7
8.1.5 to 8.1.7
8.1.7 to 9.0.0
8.1.5 to 9.0.0
8.0.5 to 9.0.0

Because the PR seems to works as expected, If you're ok with the problem asked about the php version, it's a QA ok 

Thank you

PR #734, opened 2024-06-25
Add condition when version_num is null for compute PHP compatibility range

M0rgan01:
| Questions         | Answers
| Description?      | Add condition when version_num is null for compute PHP compatibility range
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | Select "local archive" channel, save without selected files

### DIFF
--- a/classes/UpgradeSelfCheck.php
+++ b/classes/UpgradeSelfCheck.php
@@ -359,7 +359,7 @@ class UpgradeSelfCheck
             && ($this->isShopDeactivated() || $this->isLocalEnvironment())
             && $this->isCacheDisabled()
             && $this->isModuleVersionLatest()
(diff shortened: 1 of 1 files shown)


Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

[recording(246).webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/8578ec98-a339-4b6e-a844-44526a4661ce)

Tested from : 
8.0.2 to 8.1.7
8.1.5 to 8.1.7
8.0.2 to 9.0.0
8.1.5 to 9.0.0
8.1.7 to 9.0.0 

Thank you

PR #736, opened 2024-06-26
Fix minor channel on PS 8 and above

Quetzacoalt91:
| Questions         | Answers
| Description?      | When we switched to major versions with a single level of number, we did not update the way they are managed from the channel.xml file. As a result, we had to specify all minor versions instead of majors only. This PR fixes the management of major versions from PS 8.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/36430
| How to test?      | When running a shop on PS 8.0, the minor channel should propose upgrades to PS 8.1.7

### DIFF
--- a/classes/VersionUtils.php
+++ b/classes/VersionUtils.php
@@ -99,16 +99,9 @@ class VersionUtils
     /**
      * @param string $version
      *
-     * @return string|null
+     * @return array{'major': string,'minor': string,'patch': string}|null
      */
-    public static function getPrestashopMajorVersion($version)
-    {
-        preg_match('#(([0-1]{1}\.[0-9]+)|([0-9]+))(?:\.[0-9]+){2}#', $version, $matches);
-
-        return $matches[1];
-    }
-
-    public static function getPrestashopMinorVersion($version)
+    public static function splitPrestaShopVersion($version)
     {
         if (!is_string($version)) {
             throw new InvalidArgumentException('Version must be a string.');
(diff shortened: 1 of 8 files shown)


classes/VersionUtils.php
```diff
@@ -96,6 +96,17 @@ public static function isActualPHPVersionCompatible()
         return PHP_VERSION_ID >= self::MODULE_COMPATIBLE_PHP_VERSION;
     }
 
+    /**
+     * @param string $version
+     * 
+     * @return string|null
+     */
+    public static function getPrestashopMajorVersion($version)
```
Quetzacoalt91, line 104: @M0rgan01 The extracted method added is not totally aligned with the one below in terms of exceptions thrown.

There is likely some refacto to do as well, as we are using two different methods to extract separate pieces of the same variable.
We could have a single method returning `array{'minor' => string, 'major' => string}`

M0rgan01: Yes I see that, we could totally have one and the same method; We use the other to calculate compatibility ranges. I could merge them

M0rgan01 (approved):

AureRita (approved):
Hi @Quetzacoalt91 

[recording(249).webm](https://github.com/PrestaShop/autoupgrade/assets/100699445/069b2809-9eb4-474c-a176-3fbfc162dac4)


Tested from : 
- 1.7.8.11 -> Minor 1.7.8.11 
                -> Major 8.1;7
- 8.0.5      -> Minor 8.1.7 
                -> Major 8.1;7 
                -> To 9.0.0
- 8.1.5      -> Minor 8.1.7  
                -> Major 8.1;7 
                -> To 9.0.0
- 8.1.7 to 9.0.0


Thank you

PR #809, opened 2024-07-17
Make module Translator usable on Storybook

Quetzacoalt91:
| Questions         | Answers
| Description?      | Allow the translations to be used when using Storybook, allowing the templates to be multi languages while working on them. This PR also removes the use of PrestaShop constants in the `Translator` class, making it unit-testable and usable outside the BO context.
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Translations are still available on the module in the BO (non-regression tests), and you can now switch on another language from the top bar of Storybook.

[Capture vidéo du 2024-07-17 11-11-20.webm](https://github.com/user-attachments/assets/32da87b5-878d-45ca-a365-22ee03d22467)

### DIFF
--- a/autoupgrade.php
+++ b/autoupgrade.php
@@ -184,7 +184,10 @@ class Autoupgrade extends Module
     {
         require_once _PS_ROOT_DIR_ . '/modules/autoupgrade/classes/UpgradeTools/Translator.php';
 
-        $translator = new \PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator();
+        $translator = new \PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator(
+            _PS_ROOT_DIR_ . 'modules' . DIRECTORY_SEPARATOR . 'autoupgrade' . DIRECTORY_SEPARATOR . 'translations' . DIRECTORY_SEPARATOR,
+            \Context::getContext()->language->iso_code
+        );
 
         return $translator->trans($id, $parameters);
     }
--- a/storybook/config/services.yaml
+++ b/storybook/config/services.yaml
@@ -6,6 +6,12 @@ services:
         autowire: true
         autoconfigure: true
 
+    PrestaShop\Module\AutoUpgrade\:
+        resource: '../../classes'
+        exclude: '../../classes/**/index.php'
+
+    # makes classes in src/ available to be used as services
+    # this creates a service per class whose id is the fully-qualified class name
     App\:
         resource: '../src/'
         exclude:
@@ -13,3 +19,21 @@ services:
             - '../src/Entity/'
             - '../src/Kernel.php'
 
+    App\EventListener\LocaleListener:
+        tags:
+            - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest, priority: 50 }
+
+    PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator:
+        factory: ['App\Translation\TranslatorFactory', 'createTranslator']
+        arguments:
+            - '%kernel.project_dir%/../translations/'
+            - '@request_stack'
+
+    twig.extension.trans:
+        class: Symfony\Bridge\Twig\Extension\TranslationExtension
+        arguments: ['@App\Translation\TranslatorBridge']
+        tags:
+            - { name: twig.extension }
+
+    # add more service definitions when explicit configuration is needed
+    # please note that last definitions always *replace* previous ones
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -6,6 +6,7 @@ $config
     ->setUsingCache(true)
     ->getFinder()
     ->in(__DIR__)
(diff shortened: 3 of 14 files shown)


storybook/composer.json
```diff
@@ -39,7 +40,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "App\\Tests\\": "tests/"
+            "PrestaShop\\Module\\AutoUpgrade\\App\\Tests\\": "tests/"
```
Quetzacoalt91, line 43: To rollback

autoupgrade.php
```diff
@@ -184,7 +184,10 @@ public function trans($id, array $parameters = [], $domain = null, $locale = nul
     {
         require_once _PS_ROOT_DIR_ . '/modules/autoupgrade/classes/UpgradeTools/Translator.php';
 
-        $translator = new \PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator();
+        $translator = new \PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator(
+            _PS_ROOT_DIR_ . '/modules/autoupgrade/translations/',
```
ga-devfront, line 188:
```suggestion
            _PS_ROOT_DIR_ . DIRECTORY_SEPARATOR . 'modules' . DIRECTORY_SEPARATOR . 'autoupgrade' . DIRECTORY_SEPARATOR . 'translations' . DIRECTORY_SEPARATOR,
```
you sue DIRECTORY_SEPARATOR in `classes/UpgradeContainer.php` but not her. I think it's a good thing to use it.

ga-devfront (dismissed):
All seams good, just one comment to improve code style.

storybook/config/services.yaml
```diff
@@ -6,10 +6,30 @@ services:
...
         resource: '../src/'
         exclude:
             - '../src/DependencyInjection/'
             - '../src/Entity/'
             - '../src/Kernel.php'
 
+    PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator:
+        arguments:
+            $translationsFilesPath: '%kernel.project_dir%/../translations/'
+            $locale: '@=service("request_stack").getCurrentRequest()?.getPayload().all("args")["_locale"]'
```
M0rgan01, line 25: Using this type of expression is often not a good practice, it is better to inject the request stack to do this work in the constructor for several reasons:
- complexity and reduced readability
- performance (they need to be evaluated every time the service is instantiated)
- debugging
- mock in tests

tests/unit/UpgradeContainer/FilesystemAdapterTest.php
```diff
@@ -286,7 +286,7 @@ public function testTempFolderIsNotAPrestashopReleaseAfterChanges()
 
     protected function fillFolderWithPsAssets($folder)
     {
-        mkdir($folder);
+        mkdir($folder, 0777, true);
```
M0rgan01, line 289: why we have to do this ?

storybook/config/services.yaml
```diff
@@ -6,10 +6,30 @@ services:
...
+
+    twig.extension.trans:
+        class: Symfony\Bridge\Twig\Extension\TranslationExtension
+        public: false
```
M0rgan01, line 29: default service visibility is false by default

storybook/config/services.yaml
Quetzacoalt91: I'm not in my confort zone here. Even though I agree with you, I had to define in the service container the parameters of the our module `Translator` class to allow Storybook to boot (issue with the autowiring I guess).

I initially had my own construction of the Translation in the TranslatorBridge, but Symfony did no care about it.

I think sending the whole request is out of the table for our Translator in the module folder.

tests/unit/UpgradeContainer/FilesystemAdapterTest.php
Quetzacoalt91: The test was failing for some reason. Let me revert this.

storybook/config/services.yaml
M0rgan01: So it may be a better approach to go through a factory https://symfony.com/doc/current/service_container/factories.html

Quetzacoalt91: Switched on a combo Factory / EventListener

M0rgan01 (approved):

Quetzacoalt91:
Issue reported from QA after rebase. One variable needs to be set.

image

M0rgan01 (approved):

AureRita (approved):
Hi @Quetzacoalt91 

Thank you for your PR, I tested it and it seems to works as we expected : 



Tested from : 
1.7.8.8 to 8.1.7 
8.0.5 to 8.1.7 
8.1.5 to 8.1.7 
8.1.7 to 9.0.0
8.1.5 to 9.0.0
8.0.5 to 9.0.0

Thank you

PR #813, opened 2024-07-19
enchancement: Change the logger to throw exception and silent mkdir

ga-devfront:
| Questions         | Answers
| Description?      | Change the logger to throw exception and silent `mkdir` to throw understandable error. Related to this issue https://github.com/PrestaShop/PrestaShop/issues/36530 because in this case we don't know wich folder cause the issue. It's not a fix of the problem but and improvement to understand it.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | --
| How to test?      | You can set chmod 000 on /%admin_folder_name%/autoupgrade to see the right error then you access AutoUpgrade page.

### DIFF
--- a/classes/UpgradeContainer.php
+++ b/classes/UpgradeContainer.php
@@ -595,7 +595,6 @@ class UpgradeContainer
         }
 
         $this->workspace = new Workspace(
-            $this->getLogger(),
             $this->getTranslator(),
             $paths
         );
--- a/classes/Workspace.php
+++ b/classes/Workspace.php
@@ -27,17 +27,11 @@
 
 namespace PrestaShop\Module\AutoUpgrade;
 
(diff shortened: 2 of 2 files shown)


Quetzacoalt91 (approved):

M0rgan01 (approved):

florine2623 (approved):
Hello @ga-devfront ,

Tested the following :
`PS 1.7.8.11` to 8.1.7, then change to wrong permissions on the folder then rollback
`PS 8.0.5` to any version with the wrong permissions
`PR 8.1.5` to 9.0.0, then refresh the page as it won't upgrade

Here's the result, it works as expected with and without debug mode :

Screenshot 2024-08-02 at 17 32 26

Screenshot 2024-08-05 at 10 54 37

It is QA ✅ 
Thanks!

PR #814, opened 2024-07-19
Improve the upgrade configuration system

M0rgan01:
| Questions         | Answers
| Description?      | Updating the upgrade configuration system
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | The different configurations work the same way as before. It is now possible to add configurations to a CLI upgrade

### DIFF
--- a/classes/Parameters/UpgradeConfiguration.php
+++ b/classes/Parameters/UpgradeConfiguration.php
@@ -27,7 +27,9 @@
 
 namespace PrestaShop\Module\AutoUpgrade\Parameters;
 
+use Configuration;
 use Doctrine\Common\Collections\ArrayCollection;
+use Shop;
 
 /**
  * Contains the module configuration (form params).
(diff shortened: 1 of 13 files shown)


classes/Parameters/UpgradeConfiguration.php
```diff
@@ -160,6 +184,20 @@ public function shouldUpdateRTLFiles(): bool
...
     }
 
+    public static function shouldDisabledOverride(): bool
+    {
+        return (bool) Configuration::get('PS_DISABLE_OVERRIDES');
+    }
+
+    public static function updateDisabledOverride(bool $value, ?int $shopId = null): void
+    {
+        if ($shopId) {
```
kpodemski, line 194: It probably doesn't make sense to rely on `$shopId` in this module, you work on one PrestaShop instance, not on a single store.

M0rgan01: @kpodemski 

I took this configuration which was already present at the base. I tested to make sure I understand why this configuration was added. 

It is possible to set up this configuration in a particular shop in the advanced settings. 

When adding this configuration in the autoupgrade module, this feature with the context of a shop was also added, to keep the same value on this configuration between the 2 pages.

classes/Parameters/UpgradeConfiguration.php
```diff
@@ -160,6 +190,28 @@ public function shouldUpdateRTLFiles(): bool
         return (bool) $this->get('PS_AUTOUP_UPDATE_RTL_FILES');
     }
 
+    public static function shouldDisabledOverride(): bool
```
Quetzacoalt91, line 193: Let's find another name for this method, to replace the "should" and fix the verb in the past tense.

* `isOverrideAllowed()`
* `canOverrideBeUsed()`
* `isOverridingEnabled()`

classes/Parameters/UpgradeConfiguration.php
Quetzacoalt91: Disabling overrides makes sense to try the shop after upgrade. Unfortunately the global value can be overriden if a specific shop has its own configuration value.

The current implementation covers this case.

classes/Task/Miscellaneous/UpdateConfig.php
```diff
@@ -131,11 +132,16 @@ public function run(): int
...
             $config['directory.version_num'] = $configurationData['directory_num'];
         }
-        if (isset($configurationData['skip_backup'])) {
-            $config['skip_backup'] = $configurationData['skip_backup'];
-        }
-        if (isset($configurationData['PS_AUTOUP_CHANGE_DEFAULT_THEME'])) {
-            $config['PS_AUTOUP_CHANGE_DEFAULT_THEME'] = $configurationData['PS_AUTOUP_CHANGE_DEFAULT_THEME'];
+
+        foreach (UpgradeConfiguration::UPGRADE_CONST_KEYS as $key) {
+            if (isset($configurationData[$key])) {
```
Quetzacoalt91, line 137:
```suggestion
            if (!isset($configurationData[$key])) {
                continue;
            }
```

Quetzacoalt91 (dismissed):

Quetzacoalt91 (approved):

AureRita (dismissed):

PR #816, opened 2024-07-19
Make sure only the errors by merchants are sent to Sentry to avoid noise from the rest happening on the BO

Quetzacoalt91:
| Questions         | Answers
| Description?      | From the first tests, it appears anything reported on the backoffice when on the upgrade page can be sent to our Sentry project. This PR filters the errors that only triggered from the feedback modal and add useful informations for triage and debug.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Try manually sending an error to Sentry (or with `Sentry.captureMessage()`). It won't be taken in account.

* Add the concerned release of the module
* Add attachments: Contents of logs panel + Contents of error panel

### DIFF
--- a/js/sentry.js
+++ b/js/sentry.js
@@ -28,17 +28,17 @@ Sentry.init({
     release: input.autoupgrade.version,
     beforeSend(event, hint) {
         // Only the one we handle via the feedback modal must be sent.
-        if (!event.tag?.source || event.tag.source !== 'feedbackModal') {
+        if (!event.tags?.source || event.tags.source !== 'feedbackModal') {
             return null;
         }
         event.request.url = maskSensitiveInfoInUrl(event.request.url, input.adminUrl);
 
-        hint.attachments = [{ filename: "log.txt", data: readLogPanel() }];
+        hint.attachments = [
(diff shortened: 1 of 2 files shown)


ga-devfront (dismissed):

js/sentry.js
```diff
@@ -25,10 +25,18 @@
 
 Sentry.init({
     dsn: "https://eae192966a8d79509154c65c317a7e5d@o298402.ingest.us.sentry.io/4507254110552064",
+    release: document.getElementById('shop_version').innerText,
```
Quetzacoalt91, line 28: Should be the module version, not the PrestaShop one

M0rgan01 (approved):

AureRita (approved):
Hi @Quetzacoalt91 

Thank you for your PR, I tested it and it seems to works as expected

Tested from : 
1.7.8.8 to 1.7.8.11
1.7.8.8 to 8.1.7
8.0.5 to 8.1.7
8.1.5 to 8.1.7
8.1.7 to 9.0.0
8.1.5 to 9.0.0
8.0.5 to 9.0.0

Thank you

PR #832, opened 2024-07-31
Improvement for new v9 context

M0rgan01:
| Questions         | Answers
| Description?      | Quick fix on a controller display on V9 after an upgrade. A new context mechanism is present on v9, and on old controllers the constructor is called before the listenners to avoid context overload (https://github.com/PrestaShop/PrestaShop/blob/f55bf596ba0a12e22846d706ac0834be8ebf0e45/src/PrestaShopBundle/Routing/LegacyRouterChecker.php#L117). After an upgrade we disconnect the user from the session, and the employee context is null. Before 9 we were redirected before arriving at the controller, but in v9 we call the constructor and the init method to initialize the context on LegacyRouterChecker, and we have an error because the init method calls Cookie::create(context- >employee->id, context->language->iso_code)
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #https://github.com/PrestaShop/PrestaShop/issues/36642
| Sponsor company   | -
| How to test?      | -

### DIFF
--- a/controllers/admin/AdminSelfUpgradeController.php
+++ b/controllers/admin/AdminSelfUpgradeController.php
@@ -295,6 +295,12 @@ class AdminSelfUpgradeController extends ModuleAdminController
             parent::init();
         }
 
+        // V9 context security
+        // After an upgrade we disconnect the user from the session, and the employee context is null.
+        if (!$this->context->employee->id) {
+            return;
+        }
+
         // For later use, let's set up prodRootDir and adminDir
         // This way it will be easier to upgrade a different path if needed
         $this->prodRootDir = _PS_ROOT_DIR_;


ga-devfront (approved):

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

[recording(268).webm](https://github.com/user-attachments/assets/8291e8e6-3249-470f-ac57-47d1e52a78f9)


Tested from : 
8.0.5 to 8.1.7 
8.1.5 to 8.1.7 
8.1.7 to 9.0.0
8.1.5 to 9.0.0
8.0.5 to 9.0.0

Thank you

PR #847, opened 2024-08-07
Update build of storybook container for permission issues

Quetzacoalt91:
| Questions         | Answers
| Description?      | QA checks have reported issues while starting Storybook on Docker, because of files with bad user or permission. This PR changes the way containers are created, by requesting the user ID, and downloading all dependencies at startup.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | From a fresh clone of this repository with the branch, starting Storybook with Docker by foillowing the readme will avoid issues with files created as root.

### DIFF
--- a/storybook/README.md
+++ b/storybook/README.md
@@ -50,33 +50,15 @@ Once started, you can access Storybook at: http://localhost:6006/
 
 ## Build and start environment with docker
 
-### Install Project Dependencies with Docker
-
-**Install PHP Dependencies**
-
-Run the following command to install PHP dependencies via Docker:
+A Compose file is provided in the `storybook/` folder. It is responsible in installing the dependencies before running PHP and Storybook.
+To start Storybook via Docker, use the following command:
 
 ```shell
-$ docker run -u 1000:1000 --rm --interactive --tty -v "$PWD"/../:/app -w /app/storybook composer install
+$ cd storybook/
+$ THE_UID="$(id -u)" THE_GID="$(id -g)" docker compose up
 ```
 
-**Install Node dependencies**
-
-Run the following command to install Node.js dependencies via Docker:
-
-```shell
-$ docker compose run storybook-js npm install
-```
-
-### Start Docker environment
-
-**Start Symfony Server and Storybook**
-
-To start the Symfony server and Storybook via Docker, use the following command:
-
-```shell
-$ docker compose up
-```
+Providing user and group ids allows the files created during the build to be set with the proper permissions.
 
 Once started, you can access Storybook at: http://localhost:6006/
 
--- a/storybook/README.md
+++ b/storybook/README.md
@@ -50,11 +50,11 @@ Once started, you can access Storybook at: http://localhost:6006/
 
 ## Build and start environment with docker
 
-A Compose file is provided in the `storybook/` folder. It is responsible in installing the dependencies before running PHP and Storybook.
+A Docker Compose file is provided in the `storybook/` folder. It is responsible in installing the dependencies before running PHP and Storybook.
 To start Storybook via Docker, use the following command:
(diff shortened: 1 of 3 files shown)


storybook/README.md
```diff
@@ -50,33 +50,15 @@ Once started, you can access Storybook at: http://localhost:6006/
 
 ## Build and start environment with docker
 
-### Install Project Dependencies with Docker
-
-**Install PHP Dependencies**
-
-Run the following command to install PHP dependencies via Docker:
+A Compose file is provided in the `storybook/` folder. It is responsible in installing the dependencies before running PHP and Storybook.
```
M0rgan01, line 53:
```suggestion
A Docker Compose file is provided in the `storybook/` folder. It is responsible in installing the dependencies before running PHP and Storybook.
```

storybook/README.md
```diff
@@ -50,33 +50,15 @@ Once started, you can access Storybook at: http://localhost:6006/
...
+To start Storybook via Docker, use the following command:
 
 ```shell
-$ docker run -u 1000:1000 --rm --interactive --tty -v "$PWD"/../:/app -w /app/storybook composer install
+$ cd storybook/
```
M0rgan01, line 57:
```suggestion
```

We should follow the context of the README, the commands are launched from the directory thereof. As indicated above: # From the storybook/folder

storybook/docker-compose.yml
```diff
@@ -1,20 +1,41 @@
...
   storybook-js:
     build:
       context: .
       dockerfile: ./.docker/Dockerfile
-    command: npm run storybook
+    command: bash -c "npm install && npm run storybook"
+    user: "${THE_UID}:${THE_GID}"
     environment:
       IS_IN_DOCKER: true
     volumes:
```
M0rgan01, line 41: Each time we start the container, the npm install and composer install commands on the 2 composer.json will be executed, I find it very tedious to just start a container, which should be quick after having installed once dependencies.

storybook/README.md
Quetzacoalt91:
```suggestion
# From the storybook/ folder
```

storybook/docker-compose.yml
M0rgan01: It's not just about composer, the JS part is also reinstalled. If a template is modified/created, there is no need to reinstall the dependencies, because this is already configured in storybook, which directly points to the template directory. The impact is I think 10 to 20 seconds depending on the network. It's not much but if we can optimize it's better

M0rgan01: Afterwards it's just a small project for storybook, I'm not going to be stuck for that

M0rgan01 (approved):

AureRita:
Hi @Quetzacoalt91 

Thank you for your PR, I tested it and I have this message when I want to launch the command on the Readme to use it with docker : 

image

I didn't have this message before, 

Waiting for your feedback

Quetzacoalt91:
Looks like something goes wrong while downloading JS dependencies, they seem incomplete. Then, there are still permission issues with the user npm folder. :thinking:

tblivet (approved):
👌

ga-devfront (changes requested):
Don't work for me I end with a NPM error :
npm error code EACCES
npm error syscall mkdir

ga-devfront (commented):
After some tests it can run if :
- I delete node_modules, vendor, var
- add --build at the end of the command

Quetzacoalt91:
The permission issue remains on MacOS. Let not waste more time on this issue as we can still use storybook if we accept having the node_modules folder created by `root`.

PR #867, opened 2024-09-03
Improve requirements for v7.0.0

M0rgan01:
| Questions         | Answers
| Description?      | Certain requirements disappeared with v7, I also have a function to recover the wording to unify it between the CLI and the web version. Withdrawal of the requirement "checkMissingFiles"
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | Non-regression tests on the checklist

### DIFF
--- a/classes/UpgradeSelfCheck.php
+++ b/classes/UpgradeSelfCheck.php
@@ -316,7 +316,7 @@ class UpgradeSelfCheck
                 );
 
             case self::MODULE_VERSION_IS_OUT_OF_DATE:
-                return $this->translator->trans('Your current version of the module is out of date.');
+                return $this->translator->trans('Your current version of the module is outdated.');
 
             case self::PHP_REQUIREMENTS_UNKNOWN:
                 return $this->translator->trans('We were unable to check your PHP compatibility with the destination PrestaShop version.');
--- a/classes/UpgradeSelfCheck.php
+++ b/classes/UpgradeSelfCheck.php
@@ -622,6 +622,11 @@ class UpgradeSelfCheck
         return (bool) ConfigurationTest::test_upload();
     }
 
+    public function isPhpSessionsValid(): bool
+    {
+        return in_array(session_status(), [PHP_SESSION_ACTIVE, PHP_SESSION_NONE], true);
+    }
+
     /**
      * @return array<string>
      */
--- a/classes/UpgradeSelfCheck.php
+++ b/classes/UpgradeSelfCheck.php
@@ -32,115 +32,66 @@ use ConfigurationTest;
 use Exception;
 use PrestaShop\Module\AutoUpgrade\Exceptions\DistributionApiException;
 use PrestaShop\Module\AutoUpgrade\Services\DistributionApiService;
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator;
+use PrestaShop\Module\AutoUpgrade\Xml\ChecksumCompare;
 use Shop;
 
 class UpgradeSelfCheck
 {
-    /**
-     * @var bool
-     */
+    /** @var bool */
     private $fOpenOrCurlEnabled;
-
-    /**
-     * @var bool
-     */
+    /** @var bool */
     private $zipEnabled;
-
-    /**
-     * @var bool
-     */
+    /** @var bool */
     private $rootDirectoryWritable;
-
-    /**
-     * @var bool
-     */
+    /** @var bool */
     private $adminAutoUpgradeDirectoryWritable;
-
-    /**
-     * @var string
-     */
+    /** @var string */
     private $adminAutoUpgradeDirectoryWritableReport = '';
-
-    /**
-     * @var bool
-     */
+    /** @var bool */
     private $shopDeactivated;
-
-    /**
-     * @var bool
-     */
+    /** @var bool */
     private $localEnvironement;
-
-    /**
-     * @var bool
-     */
+    /** @var bool */
     private $cacheDisabled;
-
-    /**
-     * @var bool
-     */
+    /** @var bool */
     private $safeModeDisabled;
-
-    /**
-     * @var bool|mixed
-     */
+    /** @var bool|mixed */
     private $moduleVersionIsLatest;
-
-    /**
-     * @var string
-     */
-    private $rootWritableReport;
-
-    /**
-     * @var int
-     */
+    /** @var int */
     private $maxExecutionTime;
-
-    /**
-     * @var array<string, string>
-     */
+    /** @var array<string, string> */
     private $phpCompatibilityRange;
-
-    /**
-     * @var string
-     */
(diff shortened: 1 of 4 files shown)


classes/UpgradeSelfCheck.php
```diff
@@ -181,24 +132,133 @@ class UpgradeSelfCheck
...
+            self::MAX_EXECUTION_TIME_VALUE_INCORRECT => $this->getMaxExecutionTime() > 0 && $this->getMaxExecutionTime() < 30,
+            self::APACHE_MOD_REWRITE_DISABLED => !$this->isApacheModRewriteEnabled(),
+            self::NOT_LOADED_PHP_EXTENSIONS_LIST_NOT_EMPTY => !empty($this->getNotLoadedPhpExtensions()),
+            self::NOT_EXIST_PHP_FUNCTIONS_LIST_NOT_EMPTY => !empty($this->getNotExistsPhpFunctions()),
+            self::MEMORY_LIMIT_INVALID => !$this->isMemoryLimitValid(),
+            self::PHP_FILE_UPLOADS_CONFIGURATION_DISABLED => !$this->isPhpFileUploadsConfigurationEnabled(),
+            self::KEY_GENERATION_INVALID => !$this->checkKeyGeneration(),
+            self::NOT_WRITING_DIRECTORY_LIST_NOT_EMPTY => !empty($this->getNotWritingDirectories()),
+            self::SHOP_VERSION_NOT_MATCHING_VERSION_IN_DATABASE => !$this->isShopVersionMatchingVersionInDatabase(),
+        ];
```
Quetzacoalt91, line 208: If we're supposed to return only display errors, this method could return only the detected issues. Something OK would not appear.

And if you only return the issues, you could generate at this time the translated message? By doing so you would call trans only when necessary.

Quetzacoalt91 (dismissed):

Quetzacoalt91 (approved):

M0rgan01:
@Hlavtox Yes it is voluntary. This verification is already done by PrestaShop itself since PrestaShop 1.7, so we are dropping it. There will be problems before you even get to the module screen

Quetzacoalt91 (dismissed):

ga-devfront (approved):

classes/UpgradeSelfCheck.php
```diff
@@ -181,24 +132,209 @@ class UpgradeSelfCheck
...
+
+    /**
+     * @throws Exception
+     *
+     * @return array<int, bool>
+     */
+    public function getErrors(): array
+    {
+        $errors = [
+            self::PHP_REQUIREMENTS_INVALID => $this->getPhpRequirementsState() == self::PHP_REQUIREMENTS_INVALID,
```
Quetzacoalt91, line 191:
```suggestion
            self::PHP_REQUIREMENTS_INVALID => $this->getPhpRequirementsState() === self::PHP_REQUIREMENTS_INVALID,
```

classes/UpgradeSelfCheck.php
```diff
@@ -181,24 +132,209 @@ class UpgradeSelfCheck
...
+    }
+
+    /**
+     * @return array<int, bool>
+     */
+    public function getWarnings(): array
+    {
+        $warnings = [
+            self::MODULE_VERSION_IS_OUT_OF_DATE => !$this->isModuleVersionLatest(),
+            self::PHP_REQUIREMENTS_UNKNOWN => !$this->getPhpRequirementsState() == self::PHP_REQUIREMENTS_UNKNOWN,
```
Quetzacoalt91, line 220:
```suggestion
            self::PHP_REQUIREMENTS_UNKNOWN => $this->getPhpRequirementsState() === self::PHP_REQUIREMENTS_UNKNOWN,
```

Quetzacoalt91 (dismissed):

Quetzacoalt91 (approved):

Hlavtox (changes requested):
@M0rgan01 Please check this - https://github.com/PrestaShop/autoupgrade/pull/561

I specifically implemented the session check inside this module, because the core check was faulty.

M0rgan01:
@Hlavtox While waiting to investigate the subject of PHP sessions, I am removing it from the PR so as not to block it

Quetzacoalt91:
> @M0rgan01 Please see this - #561
> 
> I specifically implemented the session check inside this module, because the core check was faulty. I made the change to fix this issue - [PrestaShop/PrestaShop#30658](https://github.com/PrestaShop/PrestaShop/issues/30658).

@Hlavtox considering all the versions currently supported by the autoupgrade module require sessions to work properly, is this check still relevant in the module before an update?
I guess that if sessions don't work properly, the installation would have detected it, or in the worst case an error would have been thrown while using the backoffice or something?

Hlavtox:
@Quetzacoalt91 Well, the problem was that even though sessions worked, the check present in the core did not.

So, if you want to completely skip this check, we can remove it, because if he go to the screen, the sessions work.
But, if you want to use the CORE check, it won't work for pre 8.0 cores, I fixed the issue for 8.0.x - https://github.com/PrestaShop/PrestaShop/pull/31241

AureRita (approved):
Hi @M0rgan01 

image

Tested from : 
1.7.7.8 to 8.1.7 
8.0.5 to 8.1.7 
8.1.7 to 9.0.0
8.1.5 to 9.0.0
8.0.5 to 9.0.0

Thank you

classes/UpgradeSelfCheck.php
```diff
@@ -181,24 +132,213 @@ class UpgradeSelfCheck
...
+                    'The following PHP extension%s need%s to be installed:',
+                    [$phpExtensionsCount > 1 ? 's' : '', $phpExtensionsCount === 1 ? 's' : '']
+                );
+
+            case self::NOT_EXIST_PHP_FUNCTIONS_LIST_NOT_EMPTY:
+                $phpFunctionsCount = count($this->getNotExistsPhpFunctions());
+
+                return $this->translator->trans(
+                    'The following PHP function%s need%s to be allowed:',
+                    [$phpFunctionsCount > 1 ? 's' : '', $phpFunctionsCount === 1 ? 's' : '']
```
Hlavtox, line 294: This doesn't work for other languages. :-)

Quetzacoalt91:
I've heard of this feature, but I wasn't sure it can be used. Seeing the opened PR seems to confirm there is no such thing on the Core? :/
We'll check again, thanks for the feedback in the meantime.

Hlavtox:
@Quetzacoalt91 Symfony supports it out of the box, and it worked when I tried it, but it would require some minor tweaks. :-)

It's the proper way to do plurals in all languages.
Example:
- English:
  - 0 items
  - 1 item
  - 2 items
  - 3 items
  - 4 items etc.
- Czech:
  - 0 položek
  - 1 položka
  - 2 položky
  - 3 položky
  - 4 položky
  - 5 položek
  - 6 položek etc.

PR #877, opened 2024-09-06
Drop server performance configuration

M0rgan01:
| Questions         | Answers
| Description?      | Drop server performance configuration
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | -

### DIFF
--- a/classes/Analytics.php
+++ b/classes/Analytics.php
@@ -118,7 +118,6 @@ class Analytics
                     'upgrade_channel' => $this->upgradeConfiguration->getChannel(),
                     'backup_files_and_databases' => $this->upgradeConfiguration->shouldBackupFilesAndDatabase(),
                     'backup_images' => $this->upgradeConfiguration->shouldBackupImages(),
-                    'server_performance' => $this->upgradeConfiguration->getPerformanceLevel(),
                     'disable_non_native_modules' => $this->upgradeConfiguration->shouldDeactivateCustomModules(),
                     'upgrade_default_theme' => $this->upgradeConfiguration->shouldUpdateDefaultTheme(),
                     'switch_to_default_theme' => $this->upgradeConfiguration->shouldSwitchToDefaultTheme(),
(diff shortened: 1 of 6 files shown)


Quetzacoalt91 (approved):

kpodemski:
Are you planning to allow developers to override these values? Time is precious.

M0rgan01:
@kpodemski According to our metrics this configuration was very little modified. We aim to simplify the update process so that users are as comfortable as possible.

Time is precious -> We also study the subject of performance :)

kpodemski:
@M0rgan01 

> According to our metrics this configuration was very little modified. We aim to simplify the update process so that users are as comfortable as possible.

Of course, and that's a good approach. Two thoughts:
- nontechnical users probably won't end up doing major upgrades,
- it is an agency or technical person who does the upgrade quite often - I don't want to modify the module's files to make things quicker because I run on `max_execution=-1` and `memory_limit=-1`, etc., so it would be great to have it configurable, even by having a code like this:

$nbFiles = Configuration::get('PS_AUTOUPGRADE_NB_FILES, null, null, null, 400);

so I could tinker with perf settings :D

Hlavtox:
I don't think the loop size/time is to concern, if the upgrade takes 2 minutes o 2 minutes 5 seconds is irrelevant.

What is concerning is `'maxBackupFileSize' => [15728640, 31457280, 62914560], // bytes`, this changes the behavior of the backup. Not that I ever used this. :-)))

AureRita (approved):
Hi @M0rgan01 

[recording(287).webm](https://github.com/user-attachments/assets/1117cadf-947c-4ffd-a2bf-a4f20293a95c)


Tested from : 
1.7.8.9 to 8.1.7 
8.0.5 to 8.1.7 
8.1.7 to 9.0.0
8.1.5 to 9.0.0
8.0.5 to 9.0.0

Thank you

PR #878, opened 2024-09-09
Drop 'update default theme' configuration

M0rgan01:
| Questions         | Answers
| Description?      | Drop 'update default theme' configuration. We will systematically update the theme
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | The theme update configuration is no longer present, on the WEB or on the CLI. The theme is well updated


kpodemski:
Hello @M0rgan01 

Could you guys share the reasoning for this change? It is very common for users to use Classic and modify Classic. We all know it is not recommended, but it is what it is.

Is it because you plan to improve this setting or the plan is to always update the default theme?

M0rgan01:
Hi @kpodemski ! The theme will be updated in order to avoid that the theme is broken during a major update. The other configuration "PS_AUTOUP_CHANGE_DEFAULT_THEME" will be modified to retrieve a list of themes, and will switch to this one during the update

kpodemski:
While I understand the reasoning, you could approach this differently. You could require people to upgrade their theme only if they do a major upgrade, and you could allow the theme to remain unchanged during patch and minor upgrades. I'm certain that we will hear complaints from people if we don't make any compromises.

cc @gericfo

Hlavtox:
@M0rgan01 

> and will switch to this one during the update

I hope you don't mean to switch to default theme automatically? 🤔 That would be a bad behavior.

M0rgan01:
@Hlavtox You can choose the theme you want to switch to.

Hlavtox:
@M0rgan01 Do I have an option to stay on the current theme, because I tested it on the new version and made it compatible?

M0rgan01:
@kpodemski @Hlavtox 

We will suggest that you stay on the current theme, or switch to one of the themes present in the release. 

If you had custom themes, you will still be able to use them after the release.

If you had directly modified the classic theme, which is bad practice for customizing themes, the modifications will be deleted with the upgrade.

kpodemski:
@M0rgan01  

> If you had directly modified the classic theme, which is bad practice for customizing themes, the modifications will be deleted with the upgrade.

That's a risky choice not to let user decide, especially for the patch or minor upgrade, we will see how this goes.

M0rgan01:
We agree with you, but if we look at the metrics, 87% of users do not touch this option, and update their classic theme following an update. 

Users must follow the parent/child feature for themes https://devdocs.prestashop-project.org/8/themes/reference/template-inheritance/parent-child-feature/.

If we see that this change poses a problem later, it is entirely possible to go back

Hlavtox:
@M0rgan01 The autoupgrade module must not touch a third party theme.

> If you had custom themes, you will still be able to use them after the release.

Yes, but it must not switch to classic theme, because it completely messes up the shop - enables modules, disables modules, hooks modules, fucks up layout configuration. No go.

M0rgan01:
@Hlavtox 

> The autoupgrade module must not touch a third party theme. 

It will not touch the third-party module. 

There will be the possibility of remaining on the current theme so for me there is no problem

Hlavtox:
@M0rgan01 OK, that's good and OK! :-)

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

image


Tested from : 
1.7.7.8 to 8.1.7 
8.0.5 to 8.1.7 
8.1.5 to 8.1.7 
8.1.7 to 9.0.0
8.1.5 to 9.0.0
8.0.5 to 9.0.0

Thank you

PR #887, opened 2024-09-11
Backup Optim : Primary Key instead of Limit Offset

GautierBoudeweel:
| Questions         | Answers
| Description?      | Please be specific when describing this change. What did you change? Why?
| Type?             | bug fix / improvement / new feature / refactor
| BC breaks?        | yes / no
| Deprecations?     | yes / no
| Fixed ticket?     | Fixes #{issue URL here}, Fixes #{another issue URL here}
| How to test?      | Indicate how to verify that this change works as expected.

### DIFF
--- a/classes/Task/Upgrade/BackupDb.php
+++ b/classes/Task/Upgrade/BackupDb.php
@@ -121,7 +121,7 @@ class BackupDb extends AbstractTask
                 continue;
             }
 
-            $primary_key = $this->getPrimaryKey($table));
+            $primary_key = $this->getPrimaryKey($table);
 
             if ($written == 0 || $written > $this->container->getUpgradeConfiguration()->getMaxSizeToWritePerCall()) {
                 // increment dbStep will increment filename each time here
--- a/classes/Task/Upgrade/BackupDb.php
+++ b/classes/Task/Upgrade/BackupDb.php
@@ -254,6 +254,12 @@ class BackupDb extends AbstractTask
                             ++$i;
                         }
                         $time_elapsed = time() - $start_time;
+                    } else if ($primary_key) {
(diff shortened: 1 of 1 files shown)


classes/Task/Upgrade/BackupDb.php
```diff
@@ -219,7 +221,11 @@ public function run(): int
             if (!in_array($table, $ignore_stats_table)) {
                 do {
                     $backup_loop_limit = $this->container->getState()->getBackupLoopLimit();
-                    $data = $this->container->getDb()->executeS('SELECT * FROM `' . $table . '` LIMIT ' . (int) $backup_loop_limit . ',200', false, false);
+                    if ($primary_key) {
+                        $data = $this->container->getDb()->executeS('SELECT * FROM `' . $table . '` WHERE `' . $primary_key . '` BETWEEN ' . (int) $backup_loop_limit . ' AND ' . (int) $backup_loop_limit + 200 , false, false);
```
Quetzacoalt91, line 225: Note that in this implementation, if there is a gap in the rows that is bigger than a batch size, nothing would be returned by this request and the sync of the backup would be considered as done.

Quetzacoalt91: That's a great idea, and this should be stored in a cache. We would avoid reloading the same request at each HTTP request.

GautierBoudeweel:
There is another issue with composite primary key, like in ps_cart_rule_combination
We could ignore the primary_key optimization for these tables with composite indexes, but it is precisely on these tables that we find the largest volumes and that the need for optimization is the greatest

ga-devfront:
I'm going to close this one backup SQL optimization already done on this one https://github.com/PrestaShop/autoupgrade/pull/894.

PR #895, opened 2024-09-16
Add: Update page step version choice

ga-devfront:
| Questions         | Answers
| Description?      | Add controllers for new upgrades pages, change twig, update storybook, update vite assets management.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | You can try to access with basic url and add one of this :
- `&new-ui=1&route=home-page`
- `&new-ui=1&route=update-page-version-choice`
- `&new-ui=1&route=update-page-update-options`
- `&new-ui=1&route=update-page-backup`
- `&new-ui=1&route=update-page-update`
- `&new-ui=1&route=update-page-post-update`


ajax-upgradetab.php
```diff
@@ -49,7 +51,17 @@
...
+
+if (!empty($action)) {
+    $controller = new SingleTask($container);
+    $controller->setOptions(['action' => $action]);
+    $controller->run();
+    echo $controller->getJsonResponse();
+} else {
+    $response = (new Router($container))->handle(Request::createFromGlobals());
+    if ($response instanceof \Symfony\Component\HttpFoundation\Response) {
+        $response->send();
```
Quetzacoalt91, line 64: Isn't an exit missing here to avoid the `echo`? or a `else` ?

classes/Twig/UpdateSteps.php
```diff
@@ -0,0 +1,81 @@
...
+    const STATE_CURRENT = 'current';
+    const STATE_DONE = 'done';
+
+    /** @var Translator */
+    private $translator;
+
+    /** @var array<string, array<string,string>> */
+    private $steps;
+
+    public function __construct(Translator $translator)
```
Quetzacoalt91, line 25: Could be nice to have unit tests for this file.

classes/Twig/UpdateSteps.php
```diff
@@ -0,0 +1,81 @@
...
+
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator;
+
+class UpdateSteps
+{
+    const VERSION_CHOICE = 'version-choice';
+    const UPDATE_OPTIONS = 'update-options';
+    const BACKUP = 'backup';
+    const UPDATE = 'update';
+    const POST_UPDATE = 'post-update';
```
Quetzacoalt91: To restrict the possible values that can be sent to the method, I advise to prefix your constants with a same word. For instance:

```suggestion
    const STEP_VERSION_CHOICE = 'version-choice';
    const STEP_UPDATE_OPTIONS = 'update-options';
    const STEP_BACKUP = 'backup';
    const STEP_UPDATE = 'update';
    const STEP_POST_UPDATE = 'post-update';
```

classes/Twig/UpdateSteps.php
```diff
@@ -0,0 +1,81 @@
...
+            ],
+            self::POST_UPDATE => [
+                'title' => $this->translator->trans('Post-update'),
+            ],
+        ];
+    }
+
+    /**
+     * @return array<int, array<string, string>>
+     */
```
Quetzacoalt91: If all your possible values are the constants prefixed with the same word: 

```suggestion
    /**
     * @param self::STEP_* $currentStep
     * @return array<int, array<string, string>>
     */
```

Then PHPStan will be able to report you a value that is not allowed.

classes/Twig/UpdateSteps.php
```diff
@@ -0,0 +1,81 @@
...
+    const POST_UPDATE = 'post-update';
+
+    const STATE_NORMAL = 'normal';
+    const STATE_CURRENT = 'current';
+    const STATE_DONE = 'done';
+
+    /** @var Translator */
+    private $translator;
+
+    /** @var array<string, array<string,string>> */
```
Quetzacoalt91: I'm not sure this can work, it is worth a try.

```suggestion
    /** @var array<self::STEP_*, array<string,string>> */
```

controllers/admin/self-managed/HomePageController.php
```diff
@@ -23,13 +25,27 @@
...
+    const CURRENT_PAGE = 'home';
 
-export default {
-  component: RollbackForm,
-  args: {
-    availableBackups: ["backup1", "backup2"],
-  },
-};
+    /**
+     * @return array
```
Quetzacoalt91, line 38: PHPStan says nothing about it?

controllers/admin/self-managed/UpdatePageUpdateOptionsController.php
```diff
@@ -0,0 +1,75 @@
...
+    protected function getParams(Request $request): array
+    {
+        $updateSteps = new UpdateSteps($this->upgradeContainer->getTranslator());
+
+        return [
+            'step' => [
+                'code' => $this::CURRENT_STEP,
+                'title' => $updateSteps->getStepTitle($this::CURRENT_STEP),
+            ],
+            'steps' => $updateSteps->getSteps($this::CURRENT_STEP),
```
Quetzacoalt91: Don't you think this whole block could be return by a metid in `UpdateSteps` class?

controllers/admin/self-managed/UpdatePageVersionChoiceController.php
```diff
@@ -0,0 +1,90 @@
...
+
+        return [
+            'step' => [
+                'code' => $this::CURRENT_STEP,
+                'title' => $updateSteps->getStepTitle($this::CURRENT_STEP),
+            ],
+            'steps' => $updateSteps->getSteps($this::CURRENT_STEP),
+            'upToDate' => true /* TODO */ ,
+            'noLocalArchive' => !$this->upgradeContainer->getLocalArchiveRepository()->hasLocalArchive(),
+            // TODO: How to find images based on shop URL ?
```
Quetzacoalt91:
```suggestion
```

This should be good now

views/templates/steps/home.html.twig
```diff
@@ -1,11 +1,9 @@
-{% extends '@ModuleAutoUpgrade/layouts/page.html.twig' %}
-
-{% block stepper %}{% endblock %}
+{% extends '@ModuleAutoUpgrade/layouts/step-content.html.twig' %}
 
 {% block page_class %}welcome-page{% endblock %}
 
 {% block title %}
-    <h1>{{ 'Welcome to PrestaShop Update Assistant'|trans({}) }}</h1>
+    <h1 class="h2">{{ 'Welcome to PrestaShop Update Assistant'|trans({}) }}</h1>
```
Quetzacoalt91: Woot?

Did you mean:
```suggestion
    <h2>{{ 'Welcome to PrestaShop Update Assistant'|trans({}) }}</h2>
```

? Or maybe updating the theme?

controllers/admin/self-managed/HomePageController.php
ga-devfront: no he say nothing (see the CI)

views/templates/steps/home.html.twig
ga-devfront: @tblivet

tblivet: In terms of semantics, it's better to use an `<h1>` because we shouldn't use an `<h2>` if there is no `<h1>`. The designer changed the style, I think, because previously the `<h1>` was styled like an `<h1>`, but now it's styled like an `<h2>`. For me, the code is fine, but I agree it can be a bit confusing.

Quetzacoalt91 (approved):

ga-devfront:
I know some issues with twig templates (format), we don't fix it on this PR cause the PR is already big. But we are going to fix it into another PR after this one is merged.

AureRita (approved):
Hi @ga-devfront 

[recording(295).webm](https://github.com/user-attachments/assets/60448b1b-a520-42fe-aa55-bb003e6bf0ec)

Tested from : 
1.7.7.8 to 8.1.7 
8.0.5 to 8.1.7 
8.1.7 to 9.0.0
8.0.5 to 9.0.0

Thank you

PR #917, opened 2024-09-26
[NEW UI] Start TS

ga-devfront:
| Questions         | Answers
| Description?      | Initialize the TS for hydration pages (home page is fully fonctionnal now !)
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | You need to build the TS (see readme) and try to access to home page and navigate to update page. You can try next en before from your browser too.

### DIFF
--- a/_dev/src/ts/api/RequestHandler.ts
+++ b/_dev/src/ts/api/RequestHandler.ts
@@ -8,9 +8,7 @@ export default class RequestHandler {
   }
 
   public post(route: string, data = new FormData(), fromPopState?: boolean) {
-    if (data) {
-      data.append('dir', window.AutoUpgrade.variables.admin_dir);
-    }
+    data.append('dir', window.AutoUpgrade.variables.admin_dir);
 
     baseApi
       .post('', data, {
--- a/_dev/src/ts/api/RequestHandler.ts
+++ b/_dev/src/ts/api/RequestHandler.ts
@@ -3,9 +3,13 @@ import { ApiResponse } from '../types/apiTypes';
 import Hydration from '../utils/Hydration';
 
 export default class RequestHandler {
+  constructor() {
+    window.AutoUpgrade.classes.RequestHandler = this;
+  }
+
   public post(route: string, data = new FormData(), fromPopState?: boolean) {
     if (data) {
-      data.append('dir', window.AutoUpgrade.admin_dir);
+      data.append('dir', window.AutoUpgrade.variables.admin_dir);
     }
 
     baseApi
--- a/_dev/src/ts/api/RequestHandler.ts
+++ b/_dev/src/ts/api/RequestHandler.ts
@@ -2,13 +2,9 @@ import baseApi from './baseApi';
 import { ApiResponse } from '../types/apiTypes';
 import Hydration from '../utils/Hydration';
 
-export default class RequestHandler {
-  constructor() {
-    window.AutoUpgrade.classes.RequestHandler = this;
-  }
-
+class RequestHandler {
   public post(route: string, data = new FormData(), fromPopState?: boolean) {
-    data.append('dir', window.AutoUpgrade.variables.admin_dir);
+    data.append('dir', window.AutoUpgradeVariables.admin_dir);
 
     baseApi
       .post('', data, {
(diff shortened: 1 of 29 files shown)


_dev/src/ts/pages/Home.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  private handleSubmit = (event: Event) => {
+    event.preventDefault();
+    if (this.form) {
+      const formData = new FormData(this.form);
+
+      const route = this.form.dataset.route;
+
+      if (route) {
+        new RequestHandler().post(route, formData);
```
M0rgan01: We could put the request handler in a static variable or in the global context, so as not to have to re-instantiate it each time

_dev/src/ts/api/RequestHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ApiResponse } from '../types/apiTypes';
+import Hydration from '../utils/Hydration';
+
+export default class RequestHandler {
+  constructor() {
+    window.AutoUpgrade.classes.RequestHandler = this;
+  }
+
+  public post(route: string, data = new FormData(), fromPopState?: boolean) {
+    if (data) {
```
Quetzacoalt91: I guess this test is always true, no need to check it.

_dev/src/ts/pages/Home.ts
```diff
@@ -0,0 +1,56 @@
...
+  submitButton: HTMLButtonElement | null;
+
+  constructor() {
+    super();
+    this.form = document.forms.namedItem('next_page');
+    this.submitButton = this.getSubmitButton();
+  }
+
+  public mount = () => {
+    if (this.form) {
```
Quetzacoalt91: If the form is not found, can we consider this as normal? Or shall we trigger an exception?

_dev/src/ts/pages/Home.ts
```diff
@@ -0,0 +1,56 @@
...
+      this.submitButton?.removeAttribute('disabled');
+    } else {
+      this.submitButton?.setAttribute('disabled', 'true');
+    }
+  };
+
+  private handleSubmit = (event: Event) => {
+    event.preventDefault();
+    if (this.form) {
+      const formData = new FormData(this.form);
```
Quetzacoalt91: Move the initialization inside the if, otherwise you may define it for nothing.

_dev/src/ts/pages/Home.ts
```diff
@@ -0,0 +1,56 @@
...
+
+      const route = this.form.dataset.route;
+
+      if (route && window.AutoUpgrade.classes.RequestHandler) {
+        window.AutoUpgrade.classes.RequestHandler.post(route, formData);
```
Quetzacoalt91: If you're running a file of the autoupgrade project, is this relevant to always check everything is defined?

And if we regularly call this method, what about having an helper to reach it faster?
Something like `Api.post(...);`?

_dev/src/ts/routing/RouteHandler.ts
```diff
@@ -0,0 +1,42 @@
+export default class RouteHandler {
+  constructor() {
+    window.AutoUpgrade.classes.RouteHandler = this;
```
Quetzacoalt91: Is this a good practice? Shouldn't it be better for the caller of this class to be responsible in defining it in the global domain?

ajax-upgradetab.php
```diff
@@ -37,6 +37,8 @@
  * Calling it from the module/autoupgrade folder will have unwanted consequences on the upgrade and your shop.
  */
 require_once realpath(dirname(__FILE__) . '/../../modules/autoupgrade') . '/ajax-upgradetabconfig.php';
+// TODO: We need to remove this in the futur and add Twig into the module dependencies.
```
Quetzacoalt91:
```suggestion
// TODO: We need to remove this in the future and add Twig into the module dependencies.
```

Quetzacoalt91 (dismissed):

M0rgan01 (dismissed):

tblivet (approved):
👌

PR #932, opened 2024-10-01
Make redirection easier to call

Quetzacoalt91:
| Questions         | Answers
| Description?      | On the new UI, we need to redirect the user to another page in some cases. We had a first solution where we must provide the source in order to have the proper destination. We now have a method in the root controller allowing us to redirect to an route.
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Reach the upgrade page directly, it should redirect to the backup page.

### DIFF
--- a/classes/Router/Router.php
+++ b/classes/Router/Router.php
@@ -123,6 +123,6 @@ class Router
 
         $method = $route['method'];
 
-        return (new $route['controller']($this->upgradeContainer))->$method($request);
+        return (new $route['controller']($this->upgradeContainer, $request))->$method($request);
     }
 }
--- a/classes/Router/Router.php
+++ b/classes/Router/Router.php
@@ -123,6 +123,6 @@ class Router
 
         $method = $route['method'];
 
-        return (new $route['controller']($this->upgradeContainer, $request))->$method($request);
+        return (new $route['controller']($this->upgradeContainer, $request))->$method();
     }
 }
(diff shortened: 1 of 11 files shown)


M0rgan01 (dismissed):

classes/Router/Router.php
```diff
@@ -119,6 +119,6 @@ public function handle(Request $request)
 
         $method = $route['method'];
 
-        return (new $route['controller']($this->upgradeContainer))->$method($request);
```
ga-devfront, line 126: you prefer to pass $request to the constructor but if you check you already pass it to $method. I think we have to choose one, we don't need to pass 2 time the same variable to the Class.

ga-devfront (commented):
Just one small comment, but everything else looks good. I'm a little bit disappointed with the tests but after several read it looks good. Thank you !

classes/Router/Router.php
Quetzacoalt91: There is a topic to discuss here indeed. It is redundant to have it both in the constructor and the method, so what do we prefer:

I think it is nice to have the simplest possible method `$this->redirectTo($route)`, but this requires to have the request defined at the class level. On the other hand, I don't know if there is a potential gain to have the $request at the method level (I doubt it but could a same controller be called several times in different Request instances?).


I would be in favor of moving it in the constructor, but that's something we should all agree on.

ga-devfront (dismissed):

M0rgan01 (dismissed):

ga-devfront (dismissed):

M0rgan01 (approved):

PR #947, opened 2024-10-09
Improve UpdateConfig and ConfigurationValidator

M0rgan01:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Refactoring of the UpdateConfig task, adding cases in the ConfigurationValidator.
| Type?             | improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | Taking the configuration into account works as before, on the CLI and on the WEB version

PR #951, opened 2024-10-11
[NEW UI] finish front version choice

ga-devfront:
| Questions         | Answers
| Description?      | Finish the frontend of the new ui page update version choice
| Type?             |  new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | build the assets, dump autoload composer and try to access page with parameter `&new-ui=1&page=update-page-version-choice`


classes/Parameters/ConfigurationValidator.php
```diff
@@ -41,11 +41,19 @@ class ConfigurationValidator
...
             switch ($key) {
                 case 'channel':
                     $this->validateChannel($value);
                     break;
+                case 'archive_zip':
+                    $this->validateArchiveZip($value, $isLocal);
+                    break;
+                case 'archive_xml':
+                    $this->validateArchiveXml($value, $isLocal);
+                    break;
```
M0rgan01: A PR was extracted for this https://github.com/PrestaShop/autoupgrade/pull/947

controllers/admin/self-managed/UpdatePageVersionChoiceController.php
```diff
@@ -118,29 +125,44 @@ protected function getParams(): array
                     'xml' => $archiveRepository->getXmlLocalArchive(),
                 ],
                 'next_release' => [
-                    'version' => $this->upgradeContainer->getUpgrader()->getDestinationVersion(),
+                    'version' => $onlineDestination ? $onlineDestination->getVersion() : '0.0.0',
```
M0rgan01: Why this change? How do you determine the zip version?

_dev/src/ts/pages/HomePage.ts
```diff
@@ -2,55 +2,64 @@ import PageAbstract from './PageAbstract';
...
-  form: HTMLFormElement;
-  submitButton: HTMLButtonElement;
-
   constructor() {
     super();
-    const form = document.forms.namedItem('next_page');
-    if (form) {
-      this.form = form;
-    } else {
+    if (!this.form) {
```
Quetzacoalt91, line 7: Where is this form coming from?

_dev/src/ts/pages/HomePage.ts
```diff
@@ -2,55 +2,64 @@ import PageAbstract from './PageAbstract';
...
+    if (routeToSubmit) {
       const formData = new FormData(this.form);
-      api.post(route, formData);
+      api.post(routeToSubmit, formData);
     }
   };
+
+  private get form(): HTMLFormElement | null {
+    return document.forms.namedItem('next_page');
+  }
```
Quetzacoalt91, line 52: To avoid bringing back all the `if (this.form)` checks, what about always returning the form or triggering an exception here?

```suggestion
  private get form(): HTMLFormElement | null {
    return document.forms.namedItem('next_page') || throw new Error("The form wasn't found inside DOM. HomePage can't be initiated properly");
  }
```

You constructor can be then safely removed.

_dev/src/ts/pages/HomePage.ts
```diff
@@ -2,55 +2,64 @@ import PageAbstract from './PageAbstract';
...
+  private get formIsValid(): boolean {
+    return this.form ? this.form.checkValidity() : false;
+  }
+
+  private get submitButton(): HTMLButtonElement | undefined {
+    return this.form
+      ? (Array.from(this.form.elements).find(
+          (element) => element instanceof HTMLButtonElement && element.type === 'submit'
+        ) as HTMLButtonElement | undefined)
+      : undefined;
```
Quetzacoalt91, line 63: Same here, if possible I would trigger an exception in case we can't find the button.

classes/Commands/UpdateCommand.php
```diff
@@ -79,7 +79,7 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
 
             // in the case of commands containing the update status, it is not necessary to update the configuration
             // also we do not want to repeat the update of the config in the recursive commands
-            if ($input->getOption('data') === null) {
+            if (!$input->hasOption('data')) {
```
M0rgan01: here the condition to change, we now look if the command has an option, and not if the option is null

Quetzacoalt91: To revert

classes/UpgradeContainer.php
```diff
@@ -265,7 +265,7 @@ public function getProperty(string $property): string
             case self::LOGS_PATH:
                 return $this->autoupgradeWorkDir . DIRECTORY_SEPARATOR . 'logs';
             case self::ARCHIVE_FILENAME:
-                return $this->getUpgradeConfiguration()->getArchiveZip();
+                return $this->getUpgradeConfiguration()->getArchiveZip() ? $this->getUpgradeConfiguration()->getArchiveZip() : UpgradeConfiguration::DEFAULT_FILENAME;
```
Quetzacoalt91: Can we find another place for setting the default value? I'm not convinced the container is the best place for that.

We can find the use of default value in the `Upgrader` class, it may be enough?

M0rgan01: This will cause a problem on `case self::ARCHIVE_FILEPATH`

Quetzacoalt91: What will cause the issue? the current state or the requested change? It may be possible to handle this error even by moving the default value elsewhere.

M0rgan01: the requested change will pose a problem. You ask if the default value in upgrader can be enough, here we are going to have a problem on the `case self::ARCHIVE_FILEPATH` which itself refers to the `case self::ARCHIVE_FILENAME`.

M0rgan01: > It may be possible to handle this error even by moving the default value elsewhere.

We could put this elsewhere, but what's the point here?

Quetzacoalt91: * This is not the responsibility of the container to have the business logic, its purpose is to provide shortcuts to other parts of the codebase.
* It looks confusing to have different methods with the same same (Archive file name) but different behaviors. UpgradeContainer and UpgradeConfiguration do not work the same way.

ga-devfront: I think that the naming of this "default" value is wrong, it is the name of the online archive. We are mixing the two concepts which are not appropriate here, we should review this whole part.

M0rgan01: > This is not the responsibility of the container to have the business logic, its purpose is to provide shortcuts to other parts of the codebase. 

However, there is business logic in several places in this class

> It looks confusing to have different methods with the same same

here I don't see 2 methods with the same name

M0rgan01: @ga-devfront Yes, that's the name of the online archive. We can change the name of the constant if it improves understanding.

M0rgan01: If the location is a problem, then we would need an "intelligent" getter on the UpgradeConfiguration side, which would have this logic, and which would be called here. What do you think?

M0rgan01: It will be necessary to keep getArchiveZip() unchanged for the needs of the UI

ga-devfront: yes but the condition her is not right => $this->getUpgradeConfiguration()->getArchiveZip() ? $this->getUpgradeConfiguration()->getArchiveZip() : UpgradeConfiguration::DEFAULT_FILENAME
we need to choose the archive depending on the channel not if getArchiveZip is set or not.

Quetzacoalt91: > However, there is business logic in several places in this class

My bad if this was introduced by me or if I let it go through recently. But as we tend to make this module better, let's avoid making the same mistakes.

> If the location is a problem, then we would need an "intelligent" getter on the UpgradeConfiguration side, which would have this logic, and which would be called here. What do you think?

This is indeed what I suggested. `UpgradeConfiguration` is a candidate, `Upgrader` could fit this need as well.
And yeah there is a scenario where if the merchants configures the archive channel, then goes back to the online channel: the value of this property would be wrong because it still returns the value of the archive zip file for the archive channel, even though we went back to the online option.
This code would likely changed to fit the proper name based on the selected channel, and this is at that moment you see this logic growing in complexity (even though it remains quite simple compared to the rest of the module). That's why it should be moved elsewhere.


About
>here I don't see 2 methods with the same name

`UpgradeContainer::getProperty(self::ARCHIVE_FILENAME)` and `UpgradeConfiguration::getArchiveZip()` have almost the same meaning, but their behavior are different. One returns the actual source file to use during the update process, the other one returns the file name configured for a specific channel. It seemed to be clear for someone who recently worked on this part, but from my point of view it was confusing before the difference was explained to me.

M0rgan01: > And yeah there is a scenario where if the merchants configures the archive channel, then goes back to the online channel: the value of this property would be wrong because it still returns the value of the archive zip file for the archive channel, even though we went back to the online option.

just condition this according to the channel,  and it will work as expected.

> UpgradeContainer::getProperty(self::ARCHIVE_FILENAME) and UpgradeConfiguration::getArchiveZip() have almost the same meaning, but their behavior are different. One returns the actual source file to use during the update process, the other one returns the file name configured for a specific channel. It seemed to be clear for someone who recently worked on this part, but from my point of view it was confusing before the difference was explained to me.

There will always be the `UpgradeConfiguration::getArchiveZip()` function, which will return the actual configured value, and the `UpgradeContainer::getProperty(self::ARCHIVE_FILENAME)` function, which will return the name of the archive depending on the channel. Adding a function in UpgradeConfiguration will not necessarily help in understanding for me. `getArchiveZip` should probably be renamed to `getLocalArchiveZip`

M0rgan01 (dismissed):

controllers/admin/self-managed/HomePageController.php
```diff
@@ -47,21 +48,15 @@ public function submit(): JsonResponse
...
-            ]);
+            return AjaxResponseBuilder::nextRouteResponse(Routes::UPDATE_PAGE_VERSION_CHOICE);
         }
 
         // if is not update is restore
         if ($this->getParams()['empty_backup']) {
-            return new JsonResponse([
-                'error' => 'You can\'t acced this route because you haven\'t backup.',
-            ], 401);
+            return AjaxResponseBuilder::errorResponse('You can\'t acced this route because you haven\'t backup.', 401);
```
Quetzacoalt91, line 56:
```suggestion
            return AjaxResponseBuilder::errorResponse('You can\'t access this route because you have no backups.', 401);
```

Quetzacoalt91 (dismissed):

Quetzacoalt91 (dismissed):

Quetzacoalt91:
* The wording is incorrect, we should not have the parentheses
image

AureRita (dismissed):
Hi @ga-devfront 

[recording(312).webm](https://github.com/user-attachments/assets/9cc37452-8a11-4646-888b-56d9adb9ed05)


Tested from : 
1.7.8.9 to 8.2 
8.0.5 to 8.1.5 
8.1.5 to 8.2 
8.1.5 to 9.0.0
8.0.5 to 9.0.0

We can se that if we don't have matched two zip and xml we'll have the same behavior that previously. We already see that together and 'll fix it on another PR 

Thank you

Quetzacoalt91 (approved):

PR #956, opened 2024-10-16
Split SQL query execution

M0rgan01:
| Questions         | Answers
| Description?      | Split SQL query execution. This allows us to avoid timeouts and to have a finer calculation of the progress percentage.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | It is now possible to see in the logs the number of database queries launched. Each request is split and requires as many HTTP requests


classes/UpgradeTools/CoreUpgrader/CoreUpgrader.php
```diff
@@ -474,7 +450,8 @@ protected function runPhpQuery(string $upgrade_file, string $query): void
         if ($this->hasPhpError($phpRes)) {
             $this->logPhpError($upgrade_file, $query, $phpRes);
         } else {
-            $this->logger->debug('<div class="upgradeDbOk">[OK] PHP ' . $upgrade_file . ' : ' . $query . '</div>');
+            $this->logger->debug('Migration file: ' . $upgrade_file);
+            $this->logger->debug('Query: ' . $query);
```
Quetzacoalt91: I'm not sure repeating the migration file in a dedicated line for each request is a good idea, I guess it generates too much duplicates?

M0rgan01: I'm going to merge the two

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

Thank you for your PR, I tested it and it seems to works as you can see (we see the queries ask for the database)  : 

[recording(319).webm](https://github.com/user-attachments/assets/93669106-3758-4901-8677-50e3c7a8e097)

Tested from : 
1.7.8.8 to 8.2 
8.0.4 to 8.2
8.2 to 9.0.0
8.0.4 to 9.0.0

I See that we have a ProgressPercentage different from the backlog part and the install Part but this one is not because off the PR and it's because it's another part.

Thank you

PR #972, opened 2024-10-25
Implement common methods for controllers to make sure basics are set

Quetzacoalt91:
| Questions         | Answers
| Description?      | Change the way some data are set on the controllers level, to trigger errors when a mandatory one is forgotten. This PR also enables the update options page.
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| Sponsor company   | /
| How to test?      | Loading the existing pages still works, and the page displaying the update options is reachable.

### DIFF
--- a/controllers/admin/self-managed/UpdatePagePostUpdateController.php
+++ b/controllers/admin/self-managed/UpdatePagePostUpdateController.php
@@ -28,28 +28,20 @@
 namespace PrestaShop\Module\AutoUpgrade\Controller;
 
 use PrestaShop\Module\AutoUpgrade\Twig\UpdateSteps;
-use Twig\Error\LoaderError;
-use Twig\Error\RuntimeError;
-use Twig\Error\SyntaxError;
 
-class UpdatePagePostUpdateController extends AbstractPageController
+class UpdatePagePostUpdateController extends AbstractPageWithStepController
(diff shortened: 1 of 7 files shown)


controllers/admin/self-managed/UpdatePagePostUpdateController.php
```diff
@@ -28,28 +28,20 @@
...
-        return $this->getTwig()->render(
-            '@ModuleAutoUpgrade/steps/post-update.html.twig',
-            $this->getParams()
-        );
+        return 'update';
+    }
+
+    protected function getStepTemplate(): string
+    {
+        return UpdateSteps::STEP_POST_UPDATE;
```
ga-devfront: Others steps page return `self::CURRENT_STEP` I think you have to do the same her.

ga-devfront (changes requested):

controllers/admin/self-managed/UpdatePagePostUpdateController.php
Quetzacoalt91: You're right, this has been updated.

ga-devfront (approved):

ps-jarvis:
PR merged, well done!

PR #974, opened 2024-10-28
Improve php script error checking

M0rgan01:
| Questions         | Answers
| Description?      | Implementation of an error checking system in the php query scripts
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | Updates work the same way as before

### DIFF
--- a/classes/DbWrapper.php
+++ b/classes/DbWrapper.php
@@ -1,0 +1,149 @@
+<?php
+
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License 3.0 (AFL-3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ */
+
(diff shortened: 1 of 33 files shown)


classes/DbWrapper.php
```diff
@@ -0,0 +1,149 @@
...
+        $result = Db::getInstance()->getRow($sql, $use_cache);
+        self::validateDBQuery();
+
+        return $result;
+    }
+
+    /**
+     * @throws UpdateDatabaseException
+     */
+    public static function validateDBQuery(): void
```
Quetzacoalt91: What about:

```suggestion
    public static function assertQuerySuccess(): void
```

I think it is worth mentioning when check the resultat instead of the request itself.

classes/UpgradeTools/CoreUpgrader/CoreUpgrader.php
```diff
@@ -488,19 +495,10 @@ private function hasPhpError($phpRes): bool
...
-     * @return void
-     */
-    private function logPhpError(string $upgrade_file, string $query, $phpRes): void
+    private function logPhpError(string $upgrade_file, string $query, string $message): void
     {
         $this->logger->error(
-            '[ERROR] PHP ' . $upgrade_file . ' ' . $query . "\n" .
-            (empty($phpRes['error']) ? '' : $phpRes['error'] . "\n") .
-            (empty($phpRes['msg']) ? '' : ' - ' . $phpRes['msg'] . "\n")
+            '[ERROR] PHP ' . $upgrade_file . ' ' . $query . "\n" . $message
```
Quetzacoalt91: I would take the opportinity to remove the [ERROR] from the original string. The logger is likely able to add it

upgrade/php/ps_1770_update_charset.php
```diff
@@ -57,26 +62,26 @@ function ps_1770_update_charset()
 
     if ($adminFilterTableExists) {
         if ($adminFilterFilterIdExists) {
-            $result &= Db::getInstance()->execute(
+            $result &= DbWrapper::execute(
```
Quetzacoalt91: To keep $result as a boolean, the check should be

```suggestion
            $result = $result && DbWrapper::execute(
```

Otherwise, PHP makes a [bitwise operation](https://www.php.net/manual/en/language.operators.assignment.php#language.operators.assignment.bitwise) which lead to an integer

classes/DbWrapper.php
M0rgan01: I don't expect to have an exception on an assertion, but rather a boolean? This would require having to manage exceptions on all calls

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

Thank you for your PR, I tested it and it seems to works 

Tested from : 
1.7.7.8 to 8.2 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

Thank you

PR #975, opened 2024-10-28
[NEW UI] add error version choice

ga-devfront:
| Questions         | Answers
| Description?      | Refactor validator to display errors on new ui
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | Try local channel on new ui with zip or xml with no version or with different version.


classes/Parameters/ConfigurationValidator.php
```diff
@@ -28,79 +28,89 @@
...
 
-    /**
-     * @throws UnexpectedValueException
-     */
-    private function validateChannel(string $channel): void
+    private function validateChannel(string $channel): ?string
     {
         if ($channel !== Upgrader::CHANNEL_LOCAL && $channel !== Upgrader::CHANNEL_ONLINE) {
-            throw new UnexpectedValueException('Unknown channel ' . $channel);
+            return $this->translator->trans('Unknown channel %s', [$channel]);
```
M0rgan01, line 86: The errors present here are not normally raised in common scenarios. I wonder about the relevance of finding ourselves with translations here. Errors will not show up on the UI unless a development error was made

classes/Parameters/LocalChannelConfigurationValidator.php
```diff
@@ -0,0 +1,71 @@
...
+        $fullXmlPath = $this->downloadPath . DIRECTORY_SEPARATOR . $xmlFile;
+        if (!empty($xmlFile) && !file_exists($fullXmlPath)) {
+            return ['archive_xml' => $this->translator->trans('File %s does not exist. Unable to select that channel.', [$xmlFile])];
+        }
+
+        try {
+            $xmlVersion = (new PrestashopVersionService($this->zipAction))->extractPrestashopVersionFromXml($fullXmlPath);
+        } catch (Exception $exception) {
+            return ['archive_xml' => $this->translator->trans('We couldn\'t find a PrestaShop version in the XML file that was uploaded in your local archive. Please try again')];
+        }
```
M0rgan01: `new PrestashopVersionService` We could avoid instantiating the service multiple times and put it in dependency injection. This is also instantiated in `UpdateConfig` line 89

classes/Task/Miscellaneous/UpdateConfig.php
```diff
@@ -77,54 +74,33 @@ public function run(): int
...
 
-                return ExitCode::FAIL;
-            }
+        if ($isLocal && empty($error)) {
+            $this->container->getLocalChannelConfigurationValidator()->validate($config);
+        }
 
+        if ($isLocal) {
+            $file = $config['archive_zip'];
+            $fullFilePath = $this->container->getProperty(UpgradeContainer::DOWNLOAD_PATH) . DIRECTORY_SEPARATOR . $file;
```
M0rgan01, line 93: why continue processing if there is an error?

classes/Parameters/LocalChannelConfigurationValidator.php
```diff
@@ -0,0 +1,71 @@
...
+
+        if ($xmlVersion !== $targetVersion) {
+            return ['global' => $this->translator->trans('The PrestaShop version in your archive doesn’t match the one in XML file. Please fix this issue and try again.')];
```
M0rgan01: Why “global”? Why not version or “prestashop_version”?

controllers/admin/self-managed/UpdatePageVersionChoiceController.php
```diff
@@ -206,17 +207,40 @@ private function getRequirements(): array
...
+        }
+
+        if (empty($error)) {
+            if ($isLocal) {
+                $file = $requestConfig['archive_zip'];
+                $fullFilePath = $this->upgradeContainer->getProperty(UpgradeContainer::DOWNLOAD_PATH) . DIRECTORY_SEPARATOR . $file;
+                $config['archive_version_num'] = (new PrestashopVersionService($this->upgradeContainer->getZipAction()))->extractPrestashopVersionFromZip($fullFilePath);
+            }
+
+            $config = $this->upgradeContainer->getUpgradeConfiguration();
```
M0rgan01: Config is reassigned afterwards? the `if ($isLocal)` is therefore not useful, right?

classes/Parameters/ConfigurationValidator.php
M0rgan01: Maybe it has the scenario where the person uses the json configuration file before going to the UI? Is this taken into account?

ga-devfront: Yes if you check in UpdatePageVersionChoiceController we make both validator. If we got any error on this validator it raised an error to the front, then we need to translate it.

classes/Parameters/LocalChannelConfigurationValidator.php
ga-devfront: it's more about the frontend part we don't know how to name it because he is not attached to an unique input.

M0rgan01: this error can also be present on the CLI version, and I find that the "global" key is not very meaningful for this error

ga-devfront: yes but in CLI you never used the key.

classes/Task/Miscellaneous/UpdateConfig.php
```diff
@@ -77,54 +73,33 @@ public function run(): int
...
+        }
 
-            $xmlFile = $config['archive_xml'];
-            $fullXmlPath = $this->container->getProperty(UpgradeContainer::DOWNLOAD_PATH) . DIRECTORY_SEPARATOR . $xmlFile;
-            if (!empty($xmlFile) && !file_exists($fullXmlPath)) {
-                $this->setErrorFlag();
-                $this->logger->info($this->translator->trans('File %s does not exist. Unable to select that channel.', [$xmlFile]));
+        if (!empty($error)) {
+            $this->setErrorFlag();
+            $this->logger->info(reset($error));
```
M0rgan01: With the new error management we could use the "error" log level

classes/Parameters/LocalChannelConfigurationValidator.php
M0rgan01: For the moment yes, but the key still has no meaning in relation to the error it throws

ga-devfront: and what do you think about `form` ?

M0rgan01: I don't understand x) The errors for zip have "archive_zip", those for xml have "archive_xml", so why the error for version matching can't be "version" or "file_version"?

M0rgan01: why `global` and `form` ?

ga-devfront: cause we don't have any field named 'version' or 'file_version'.

M0rgan01: A validator shouldn’t contain responses for specific front-end requirements, right? It's fine to return keys or error codes, but front-end technical needs shouldn’t alter the validator's behavior

classes/Parameters/LocalChannelConfigurationValidator.php
```diff
@@ -0,0 +1,146 @@
...
+                'message' => $this->translator->trans('We couldn\'t find a PrestaShop version in the .zip file that was uploaded in your local archive. Please try again.'),
+                'target' => 'archive_zip',
+            ];
+        }
+
+        return null;
+    }
+
+    /**
+     * @return array<string, string>|null
```
Quetzacoalt91: Not sure the type is right

```suggestion
     * @return array{'message': string, 'target'?: string}|null
```

classes/Parameters/LocalChannelConfigurationValidator.php
```diff
@@ -0,0 +1,146 @@
...
+                'message' => $this->translator->trans('We couldn\'t find a PrestaShop version in the XML file that was uploaded in your local archive. Please try again.'),
+                'target' => 'archive_xml',
+            ];
+        }
+
+        return null;
+    }
+
+    /**
+     * @return array<string, string>|null
```
Quetzacoalt91:
```suggestion
     * @return array{'message': string, 'target'?: string}|null
```

classes/Parameters/LocalChannelConfigurationValidator.php
```diff
@@ -0,0 +1,146 @@
...
+            if ($versionErrors) {
+                $errors[] = $versionErrors;
+            }
+        }
+
+        return $errors;
+    }
+
+    /**
+     * @return array<string, string>|null
```
Quetzacoalt91:
```suggestion
     * @return array{'message': string, 'target'?: string}|null
```

classes/Parameters/LocalChannelConfigurationValidator.php
```diff
@@ -0,0 +1,146 @@
...
+    {
+        $this->translator = $translator;
+        $this->downloadPath = $downloadPath;
+        $this->versionService = $versionService;
+    }
+
+    /**
+     * @param array<string, mixed> $array
+     *
+     * @return array<array<string, string>>
```
Quetzacoalt91:
```suggestion
     * @return array<array{'message': string, 'target'?: string}>
```

classes/Parameters/ConfigurationValidator.php
```diff
@@ -28,79 +28,91 @@
...
+        $this->translator = $translator;
+    }
+
     /**
      * @param array<string, mixed> $array
      *
-     * @return void
-     *
-     * @throws UnexpectedValueException
+     * @return array<array<string, string>>
```
Quetzacoalt91:
```suggestion
     * @return array<array{'message': string, 'target': string}|null>
```

controllers/admin/self-managed/UpdatePageVersionChoiceController.php
```diff
@@ -206,17 +206,51 @@ private function getRequirements(): array
...
+            if ($isLocal) {
+                $file = $requestConfig['archive_zip'];
+                $fullFilePath = $this->upgradeContainer->getProperty(UpgradeContainer::DOWNLOAD_PATH) . DIRECTORY_SEPARATOR . $file;
+                $requestConfig['archive_version_num'] = $this->upgradeContainer->getPrestashopVersionService()->extractPrestashopVersionFromZip($fullFilePath);
+            }
+
+            $config = $this->upgradeContainer->getUpgradeConfiguration();
+            $config->merge($requestConfig);
+
+            (new UpgradeConfigurationStorage($this->upgradeContainer->getProperty(UpgradeContainer::WORKSPACE_PATH) . DIRECTORY_SEPARATOR))->save($config, UpgradeFileNames::CONFIG_FILENAME);
```
Quetzacoalt91:
```suggestion
            $this->upgradeContainer->getUpgradeConfigurationStorage()->save($config, UpgradeFileNames::CONFIG_FILENAME);
```

tests/unit/Parameters/ConfigurationValidatorTest.php
```diff
@@ -28,90 +28,128 @@
...
-        $this->expectException(UnexpectedValueException::class);
-        $this->expectExceptionMessage('Unknown channel toto');
+        $channel = 'toto';
 
-        $validator->validate(['channel' => 'toto']);
+        $result = $this->validator->validate(['channel' => $channel]);
+        $this->assertEquals([
+            'message' => 'Unknown channel ' . $channel,
+            'target' => 'channel',
+        ], $result[0]);
```
Quetzacoalt91: If that is relevant, you can test the whole array returned

```suggestion
        $this->assertEquals([
            [
                'message' => 'Unknown channel toto',
                'target' => 'channel',
            ]
        ], $result);
```

* You make sure there is no other unexpected errors
* You avoid unnecessary logic in the test (https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html)

classes/Parameters/LocalChannelConfigurationValidator.php
```diff
@@ -0,0 +1,146 @@
+<?php
+
+namespace PrestaShop\Module\AutoUpgrade\Parameters;
+
+use Exception;
+use PrestaShop\Module\AutoUpgrade\Services\PrestashopVersionService;
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator;
+
+class LocalChannelConfigurationValidator
```
M0rgan01: Be careful not to forget the comment for the license

Quetzacoalt91 (approved):

M0rgan01 (approved):

AureRita (approved):
Hi @ga-devfront 

[recording(321).webm](https://github.com/user-attachments/assets/91e1fea5-271f-49e6-be87-22ec67618776)

Tested from : 
1.7.8.8 to 8.2 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

Thank you

PR #976, opened 2024-10-28
[NEW UI] Update options page

Quetzacoalt91:
| Questions         | Answers
| Description?      | Handle update option with a brand new page
| BC breaks?        | Yes, the config var `PS_AUTOUP_KEEP_EMAIL` is now `PS_AUTOUP_REGEN_EMAIL`
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Update options can be updated via the new UI. Refreshing a page after an option is switched on or off must keep its value.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Here is an example of the different fields that can be found in it:
   "archive_num": "8.0.0", // Release number, specific to the archive channel
   "PS_AUTOUP_CUSTOM_MOD_DESACT": 1, // Disable non-native modules
   "PS_AUTOUP_CHANGE_DEFAULT_THEME": 0, // Keep the current theme
-  "PS_AUTOUP_KEEP_MAILS": 0, // Retain customized email templates
+  "PS_AUTOUP_REGEN_EMAIL": 1, // Retain customized email templates
   "PS_AUTOUP_BACKUP": 0, // Do not create a store backup
   "PS_AUTOUP_KEEP_IMAGES": 1, // Retain images
   "PS_DISABLE_OVERRIDES": 1 // Disable all overrides
--- a/_dev/src/ts/pages/UpdatePageUpdateOptions.ts
+++ b/_dev/src/ts/pages/UpdatePageUpdateOptions.ts
@@ -3,28 +3,19 @@ import api from '../api/RequestHandler';
 
 export default class UpdatePageUpdateOptions extends UpdatePage {
   protected stepCode = 'update-options';
-  protected form: HTMLFormElement;
-  protected formFields: Element[] = [];
-
-  constructor() {
-    super();
-    this.form = this.assertAndGetForm();
-  }
 
   public mount() {
     this.initStepper();
-    this.initFormFields(this.form);
-    this.form.addEventListener('submit', this.onFormSubmit);
+    this.form.addEventListener('submit', this.onSubmit);
+    this.form.addEventListener('change', this.onChange);
   }
 
   public beforeDestroy() {
-    this.form.removeEventListener('submit', this.onFormSubmit);
-    this.formFields.forEach((element) => {
-      element.removeEventListener('change', this.onInputChange);
-    });
+    this.form.removeEventListener('submit', this.onSubmit);
+    this.form.removeEventListener('change', this.onChange);
   }
 
-  private assertAndGetForm(): HTMLFormElement {
+  private get form(): HTMLFormElement {
     const form = document.forms.namedItem('update-options-page-form');
     if (!form) {
       throw new Error('Form not found');
@@ -39,19 +30,7 @@ export default class UpdatePageUpdateOptions extends UpdatePage {
     return form;
   }
 
-  private initFormFields(form: HTMLFormElement): void {
-    Array.from(form.elements).forEach((element) => {
-      if (element.nodeName !== 'INPUT') {
-        return;
-      }
-
-      this.formFields.push(element);
-
-      element.addEventListener('change', this.onInputChange);
-    });
-  }
-
-  private onInputChange = (ev: Event) => {
+  private readonly onChange = (ev: Event) => {
     const optionInput = ev.target as HTMLInputElement;
 
     optionInput.setAttribute('disabled', 'true');
@@ -64,7 +43,7 @@ export default class UpdatePageUpdateOptions extends UpdatePage {
     optionInput.removeAttribute('disabled');
   };
 
-  private onFormSubmit = (event: Event) => {
+  private readonly onSubmit = (event: Event) => {
     event.preventDefault();
 
     api.post(this.form.dataset.routeToSubmit!, new FormData(this.form));
--- a/_dev/src/ts/pages/UpdatePageUpdateOptions.ts
+++ b/_dev/src/ts/pages/UpdatePageUpdateOptions.ts
@@ -41,7 +41,7 @@ export default class UpdatePageUpdateOptions extends UpdatePage {
 
   private initFormFields(form: HTMLFormElement): void {
     Array.from(form.elements).forEach((element) => {
-      if (element.nodeName !== "INPUT") {
+      if (element.nodeName !== 'INPUT') {
         return;
       }
 
@@ -52,17 +52,17 @@ export default class UpdatePageUpdateOptions extends UpdatePage {
   }
 
   private onInputChange = (ev: Event) => {
-    const optionInput = (ev.target as Element);
+    const optionInput = ev.target as HTMLInputElement;
 
     optionInput.setAttribute('disabled', 'true');
 
     const data = new FormData();
     data.append('name', optionInput.name);
-    data.append('value', optionInput.checked);
+    data.append('value', JSON.stringify(optionInput.checked));
     api.post(this.form.dataset.routeToSave!, data);
 
     optionInput.removeAttribute('disabled');
-  }
+  };
 
   private onFormSubmit = (event: Event) => {
     event.preventDefault();
--- a/_dev/src/ts/pages/UpdatePageUpdateOptions.ts
+++ b/_dev/src/ts/pages/UpdatePageUpdateOptions.ts
@@ -30,7 +30,7 @@ export default class UpdatePageUpdateOptions extends UpdatePage {
     return form;
   }
 
-  private readonly onChange = (ev: Event) => {
+  private readonly onChange = async (ev: Event) => {
     const optionInput = ev.target as HTMLInputElement;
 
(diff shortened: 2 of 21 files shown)


Quetzacoalt91:
Requires the 2 following PRs to be merged:
* https://github.com/PrestaShop/autoupgrade/pull/975
* https://github.com/PrestaShop/autoupgrade/pull/980

views/templates/steps/update-options.html.twig
```diff
@@ -7,38 +7,51 @@
...
+    data-route-to-submit="{{ form_route_to_submit }}"
+    id="update-options-page-form"
+    name="update-options-page-form"
+  >
     {% include "@ModuleAutoUpgrade/components/render-switch.html.twig" with {
-      id: "PS_AUTOUP_CUSTOM_MOD_DESACT",
-      name: "PS_AUTOUP_CUSTOM_MOD_DESACT",
+      id: form_fields.deactive_non_native_modules.field,
+      name: form_fields.deactive_non_native_modules.field,
       title: "Deactivate non-native modules",
```
Quetzacoalt91:
```suggestion
      title: 'Deactivate non-native modules'|trans({}),
```

_dev/src/ts/pages/UpdatePageUpdateOptions.ts
```diff
@@ -1,13 +1,51 @@
...
+
+  private readonly onChange = async (ev: Event) => {
+    const optionInput = ev.target as HTMLInputElement;
+
+    optionInput.setAttribute('disabled', 'true');
+
+    const data = new FormData();
+    data.append('name', optionInput.name);
+    data.append('value', JSON.stringify(optionInput.checked));
+    await api.post(this.form.dataset.routeToSave!, data);
```
Quetzacoalt91: @ga-devfront I've kept this implementation after trying switching on the whole form in FormData. Unfortunately the options are only sent if they are checked, which does not fit the ConfigurationValidator behavior where only updated data is expected.

README.md
```diff
@@ -99,7 +99,7 @@ Here is an example of the different fields that can be found in it:
   "archive_num": "8.0.0", // Release number, specific to the archive channel
   "PS_AUTOUP_CUSTOM_MOD_DESACT": 1, // Disable non-native modules
   "PS_AUTOUP_CHANGE_DEFAULT_THEME": 0, // Keep the current theme
-  "PS_AUTOUP_KEEP_MAILS": 0, // Retain customized email templates
+  "PS_AUTOUP_REGEN_EMAIL": 1, // Retain customized email templates
   "PS_AUTOUP_BACKUP": 0, // Do not create a store backup
   "PS_AUTOUP_KEEP_IMAGES": 1, // Retain images
   "PS_DISABLE_OVERRIDES": 1 // Disable all overrides
```
M0rgan01, line 105: the comments were minor but helped understand the configurations. Maybe adding an explanation on the README or a link to the devdoc would help

classes/Analytics.php
```diff
@@ -124,7 +124,7 @@ public function getProperties($type): array
                     'upgrade_channel' => $this->upgradeConfiguration->getChannel(),
                     'disable_non_native_modules' => $this->upgradeConfiguration->shouldDeactivateCustomModules(),
                     'switch_to_default_theme' => $this->upgradeConfiguration->shouldSwitchToDefaultTheme(),
-                    'keep_customized_email_templates' => $this->upgradeConfiguration->shouldKeepMails(),
+                    'keep_customized_email_templates' => !$this->upgradeConfiguration->shouldRegenerateMailTemplates(),
```
M0rgan01: maybe we could also change the tag on segment? We could manage this by version right?

controllers/admin/self-managed/UpdatePageUpdateOptionsController.php
```diff
@@ -49,23 +54,81 @@ protected function displayRouteInUrl(): ?string
...
+        $name = $this->request->request->get('name');
+
+        $config = [
+            $name => $this->request->request->get('value'),
+        ];
+
+        $error = $this->upgradeContainer->getConfigurationValidator()->validate($config);
+
+        if (empty($error)) {
+            if ($name === 'PS_DISABLE_OVERRIDES') {
```
M0rgan01: We could wait for the PR https://github.com/PrestaShop/autoupgrade/pull/993 and thus use the new constants, same thing in getParams

controllers/admin/self-managed/UpdatePageUpdateOptionsController.php
```diff
@@ -49,23 +54,81 @@ protected function displayRouteInUrl(): ?string
...
+        $upgradeConfiguration = $this->upgradeContainer->getUpgradeConfiguration();
+        $upgradeConfigurationStorage = $this->upgradeContainer->getUpgradeConfigurationStorage();
+
+        $name = $this->request->request->get('name');
+
+        $config = [
+            $name => $this->request->request->get('value'),
+        ];
+
+        $error = $this->upgradeContainer->getConfigurationValidator()->validate($config);
```
Quetzacoalt91: Rename as `$errors`

Quetzacoalt91: Data structure is incorrect for template and needs to be reworked before being sent. Reuse logic from `controllers/admin/self-managed/UpdatePageVersionChoiceController.php::save()`

_dev/src/ts/pages/UpdatePageUpdateOptions.ts
ga-devfront: you can do it like this to get all the form with all chebox true or false =>
```
const checkboxStates = Array.from(this.form.elements)
  .filter(el => el.type === 'checkbox')
  .reduce((acc, checkbox) => {
    acc[checkbox.name] = checkbox.checked;
    return acc;
  }, {});
  ```

File: controllers/admin/self-managed/UpdatePageUpdateOptionsController.php (hunk 5 above)
Comment (ga-devfront) [reply controllers/admin/self-managed/UpdatePageUpdateOptionsController.php]: +1

File: classes/Analytics.php (hunk 4 above)
Comment (Quetzacoalt91) [reply classes/Analytics.php]: I needed to check with the data tcheam first, but yeah we can do that knowing the data will be reset on the date of the v7 release.

File: README.md (hunk 3 above)
Comment (Quetzacoalt91) [reply README.md:105]: Good idea, we can go further than a unique example.
I've created a task for this matter so it can go in a dedicated PR: SUE-248

**Review by M0rgan01 — DISMISSED** [review]

**Comment by Quetzacoalt91** [comment]
Blocked until we get an additional acceptance criteria from Geric.

File: classes/Parameters/ConfigurationValidator.php [hunk 7]
```diff
@@ -107,9 +107,12 @@ private function validateArchiveXml(string $xml, bool $isLocal): ?string
...
-    private function validateBool(string $boolValue, string $key): ?string
+    /**
+     * @param string|bool $boolValue
+     */
+    private function validateBool($boolValue, string $key): ?string
     {
-        if ($boolValue === '' || filter_var($boolValue, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) === null) {
+        if (!is_bool($boolValue) && ($boolValue === '' || filter_var($boolValue, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) === null)) {
             return $this->translator->trans('Value must be a boolean for %s', [$key]);
         }
```
M0rgan01, line 116: I don't think it's a good practice to add mixed typing, especially if one day we want to upgrade phpstan level. This function was intended to validate the boolean in the form of a string, perhaps we could do this line 71, or in another function

classes/Parameters/UpgradeConfiguration.php
```diff
@@ -242,7 +242,9 @@ public static function updatePSDisableOverrides(bool $value): void
     public function merge(array $array = []): void
     {
         foreach ($array as $key => $value) {
-            $this->set($key, $value);
+            if (in_array($key, self::UPGRADE_CONST_KEYS)) {
+                $this->set($key, $value);
+            }
         }
     }
```
M0rgan01: If the input array contains something other than the accepted configurations, then an error should probably be raised, or the element in question should be logged.

classes/Parameters/ConfigurationValidator.php
Quetzacoalt91: Let's remember these methods return an error when something is wrong. If we go with another method, this new one would return a boolean to let us know if further checks are needed or not. This makes the contents of the class more confusing.

Having this in the switch / case list adds unnecessary complexity as well. We have calls to a single method for each case and I like the way it currently is. Simple and efficient.

The `mixed` case does not apply here as we specify two types in the php doc block. Even if the freshly introduce lvl 10 of PHPStan, this definition pass:
image

classes/Parameters/UpgradeConfiguration.php
Quetzacoalt91: I've reverted this change so we can move forward.

classes/Parameters/ConfigurationValidator.php
M0rgan01: > Having this in the switch / case list adds unnecessary complexity as well.

I don't see why this would add complexity, we would just have the check before calling the method, which could be renamed to explain that we are validating a boolean "string"

PHPstan may not say anything, but this remains a practice that does not follow modern PHP practices. This would not have been possible on languages ​​with (really) strong typing, and I think it's better.

I think that on the contrary, a parameter with multiple types can make the function more difficult to understand, because it can be difficult to know how the function will behave depending on the type passed.

M0rgan01: But I don't want to block the PR for this detail, we have to move it forward :+1:

ga-devfront (dismissed):
image

Quetzacoalt91:
Following latest conversation with @gericfo:
* The current behavior on the switches is acceptable in case of errors,
* Some acceptance criteria will be added on the related user story, and will be done on another pull-request.

Quetzacoalt91:
Had to rebase after the merge of the latest pull-requests

Here we go again

ga-devfront (approved):

AureRita (approved):
Hi @Quetzacoalt91 

[recording(333).webm](https://github.com/user-attachments/assets/2a10132d-7549-4079-89eb-a7f58eeceb25)


Tested from : 
1.7.8.8 to 8.2 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

Thank you

PR #978, opened 2024-10-29
Add CI for Twig

tblivet:
| Questions         | Answers
| Description?      | Add CI check for Twig
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| How to test?      | Check in "Actions" and "Twig Check" if the CI is green !

### DIFF
--- a/.github/workflows/twig-integration.yml
+++ b/.github/workflows/twig-integration.yml
@@ -1,0 +1,34 @@
+name: Twig Check
+
+on: [push, pull_request]
+
+jobs:
+  twig_cs_check:
+    name: Twig Check
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checkout the repository to access the code
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      # Set up PHP with Composer
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.2'  # Use the appropriate PHP version for your project
+          tools: composer
+
+      # Install project dependencies (including Twig CS Fixer)
+      - name: Install dependencies
+        run: composer install --prefer-dist --no-progress --no-suggest
+        working-directory: ./tests/twig-check
+
+      # Run the Twig CS Fixer check command
+      - name: Run Twig CS Fixer check
+        run: |
+          composer run twig || {
+            echo "::error::Twig CS Fixer check failed. Run 'composer run twig:fix' to fix the issues.";
+            exit 1;
+          }
+        working-directory: ./tests/twig-check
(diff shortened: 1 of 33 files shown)


.github/workflows/twig-integration.yml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+      # Checkout the repository to access the code
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      # Set up PHP with Composer
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.2'  # Use the appropriate PHP version for your project
```
M0rgan01, line 19: Is this the comments of an AI?

tests/twig-check/composer.json
```diff
@@ -0,0 +1,14 @@
...
+  "config": {
+    "platform": {
+      "php": "8.2"
+    }
+  },
+  "scripts": {
+    "twig": "vendor/bin/twig-cs-fixer lint \"../../views/templates\"",
+    "twig:fix": "vendor/bin/twig-cs-fixer fix \"../../views/templates\""
+  }
+}
```
M0rgan01, line 14: It's starting to have a lot of composer.json... We have 4 including the storybook one. We don't have a way to use the main composer, with the dependency in requires-dev?

.github/workflows/twig-integration.yml
tblivet: No, it's a comment from the docs I wrote, but you're right, I can remove it. 👍

tests/twig-check/composer.json
tblivet: I tried using the main composer.json with the dependencies in require-dev, but I couldn’t get it to work because it requires a more recent PHP version. Do we have any other options?

M0rgan01: I think we could use docker. You must create an image, use the PHAR given by the project

Ex: 

```
FROM php:8.2-cli

RUN wget -c https://github.com/VincentLanglet/Twig-CS-Fixer/releases/download/3.1.0/twig-cs-fixer.phar \
    -O /usr/local/bin/twig-cs-fixer

RUN chmod +x /usr/local/bin/twig-cs-fixer

ENTRYPOINT ["twig-cs-fixer"]
CMD ["twig"]

```
`docker build -t twig-cs-fixer-image .`
`docker run --rm twig-cs-fixer-image`

Quetzacoalt91: If there is no way to have an additional tool in our existing dependency list, I would rather not try to force its inclusion in the project because it would create one more way to run a specific binary. :/

tblivet: Ok, I close this PR !

PR #979, opened 2024-10-29
Fix translations

Hlavtox:
| Questions         | Answers
| Description?      | This is not translatable into other languages, creates nonsense strings. :-))) Proper plurals must be done by ICU. Unfortunately, not working in PrestaShop.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Crowdin
| Sponsor company   | 
| How to test?      | No need to test.

### DIFF
--- a/classes/UpgradeSelfCheck.php
+++ b/classes/UpgradeSelfCheck.php
@@ -250,13 +250,9 @@ class UpgradeSelfCheck
                 return $this->translator->trans('Apache mod_rewrite needs to be enabled.');
 
             case self::NOT_LOADED_PHP_EXTENSIONS_LIST_NOT_EMPTY:
-                $phpExtensionsCount = count($this->getNotLoadedPhpExtensions());
-
                 return $this->translator->trans('The following PHP extensions need to be installed:');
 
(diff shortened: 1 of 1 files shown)


Quetzacoalt91 (approved):

Quetzacoalt91:
Let's keep it simple indeed

Hlavtox:
@Quetzacoalt91 A partially-working solution would be to create two different strings, if there is 1 or more. That would be translatable.

But still not an ideal solution, it would work in English and Czech, but maybe other languages would have different plural forms for 2,3,4,5 etc. items.

Quetzacoalt91:
Yeah I've dealt with this topic on other projects by providing the different versions in a single string, each version separated by a pipe.
For instance: `I have no apples|I have 1 apple|I have %d apples`.

But it was possible if the translator was able to find the proper part to use, depending on the target language and the parameters, and this is not the case here. (yet). :/

M0rgan01 (approved):

ps-jarvis:
PR merged, well done!

PR #984, opened 2024-10-30
[NEW-UI] optimise frontend

ga-devfront:
| Questions         | Answers
| Description?      | Refactor TS and add JS doc
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | Try if nothing change on new UI

### DIFF
--- a/_dev/src/ts/api/RequestHandler.ts
+++ b/_dev/src/ts/api/RequestHandler.ts
@@ -5,14 +5,29 @@ import Hydration from '../utils/Hydration';
 export class RequestHandler {
   private currentRequestAbortController: AbortController | null = null;
 
-  public async post(route: string, data = new FormData(), fromPopState?: boolean) {
+  /**
+   * @public
+   * @param {string} route - Target route for the POST request.
+   * @param {FormData}[data=new FormData()] - Form data to send with the request by default we send FormData with admin dir required by backend.
+   * @param {boolean} [fromPopState] - Indicates if the request originated from a popstate event need by hydration.
+   * @returns {Promise<void>}
+   * @description Sends a POST request to the specified route with optional data and pop state indicator. Cancels any ongoing request before initiating a new one.
+   */
+  public async post(
+    route: string,
+    data: FormData = new FormData(),
+    fromPopState?: boolean
+  ): Promise<void> {
+    // Cancel any previous request if it exists
     if (this.currentRequestAbortController) {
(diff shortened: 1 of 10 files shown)


_dev/src/ts/api/RequestHandler.ts
```diff
@@ -5,30 +5,55 @@ import Hydration from '../utils/Hydration';
...
-        const data = response.data as ApiResponse;
-        this.handleResponse(data, fromPopState);
       });
+
+      const responseData = response.data as ApiResponse;
+      await this.#handleResponse(responseData, fromPopState);
+    } catch (error) {
+      // TODO: catch errors
+      console.error(error);
+    }
```
Quetzacoalt91, line 44: Is this `try` = `catch` really necessary? It silences the error and the callee is now unaware of what's happening.

_dev/src/ts/routing/ScriptHandler.ts
```diff
@@ -28,16 +25,47 @@ export default class ScriptHandler {
...
+    const pageClass = this.#routesMatching[routeName];
+    if (pageClass) {
+      try {
+        this.#currentScript = new pageClass();
+        this.#currentScript.mount();
+      } catch (error) {
+        console.error(`Failed to load script for route ${routeName}:`, error);
+      }
+    } else {
+      console.warn(`No matching page Class found for route: ${routeName}`);
```
Quetzacoalt91: Is this really an issue if there is no matching script? I would be more in favor of a `debug` call.

_dev/src/ts/api/RequestHandler.ts
ga-devfront: I did this so that I could implement the error page in case of 500

_dev/src/ts/routing/ScriptHandler.ts
ga-devfront: if you prefer a console.debug instead of .warn I don't see any problem with that.

Quetzacoalt91 (dismissed):

Quetzacoalt91 (approved):

AureRita (approved):
Hi @ga-devfront 

image


Tested from : 
1.7.8.8 to 8.2 
8.0.4 to 8.2  
8.2 to 9.0.0
8.0.5 to 9.0.0

Thank you

PR #989, opened 2024-11-05
Add requirements listing in new UI

M0rgan01:
| Questions         | Answers
| Description?      | Add requirements listing in new UI
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | -

### DIFF
--- a/classes/Commands/CheckRequirementsCommand.php
+++ b/classes/Commands/CheckRequirementsCommand.php
@@ -28,6 +28,8 @@
 namespace PrestaShop\Module\AutoUpgrade\Commands;
 
 use Exception;
+use PrestaShop\Module\AutoUpgrade\Exceptions\DistributionApiException;
+use PrestaShop\Module\AutoUpgrade\Exceptions\UpgradeException;
 use PrestaShop\Module\AutoUpgrade\Services\DistributionApiService;
 use PrestaShop\Module\AutoUpgrade\Services\PhpVersionResolverService;
 use PrestaShop\Module\AutoUpgrade\Task\ExitCode;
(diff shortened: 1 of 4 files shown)


classes/Commands/CheckRequirementsCommand.php
```diff
@@ -114,47 +116,25 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
...
-                   $this->writeError($this->upgradeSelfCheck->getRequirementWording($error));
-                   foreach ($notWritingDirectories as $notWritingDirectory) {
-                       $this->output->writeln('    ' . $notWritingDirectory);
-                   }
-                   break;
-               default:
-                   $this->writeError($this->upgradeSelfCheck->getRequirementWording($error));
-           }
+            $wording = $this->upgradeSelfCheck->getRequirementWording($error);
+            $this->writeError($wording['value']);
```
ga-devfront: what do you think about `message` instead of `value` ?

classes/UpgradeSelfCheck.php
```diff
@@ -194,101 +195,156 @@ public function getWarnings(): array
...
+                } else {
+                    $params = [
+                        '[1]' => '',
+                        '[/1]' => '',
+                    ];
+                }
+
+                return [
+                    'value' => $this->translator->trans('Maintenance mode needs to be enabled. Enable maintenance mode and add your maintenance IP in [1]Shop parameters > General > Maintenance[/1].', $params),
+                    'list' => [],
```
ga-devfront: why do you add empty list ? I think we can just add list if needed no ?

ga-devfront (dismissed):
I think all since good for me got just few comments

classes/UpgradeSelfCheck.php
M0rgan01: This saves me from having to check if there is a list in the command or in the template.

M0rgan01: Afterwards, it could be cleaner. I'm going to make this optional

ga-devfront (dismissed):

ga-devfront (approved):

ga-devfront (dismissed):

ga-devfront (dismissed):

AureRita (approved):
Hi @M0rgan01 

image


Tested from : 
1.7.8.10 to 8.2 
8.0.5 to 8.2 
8.2 to 9.0.0
8.0.5 to 9.0.0

We also see that we doesn't have scroll bar when we have to much error as you can see : 

image


Because this one'll be corrected in other PR, this PR seems to works as expected, It's QA :heavy_check_mark: 

Thank you

PR #990, opened 2024-11-05
[NEW UI] Optimise version choice loading

ga-devfront:
| Questions         | Answers
| Description?      | the version choice page should load very quickly and the requirements check only runs once on the page. Requirements should only be displayed if there are no errors in our form. This also allows errors to be displayed if present when loading the page (at the end of the API call), which was not the case before.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | Multiple thing to test see bellow.

### Fast Loading of the Version Choice Page:
**Objective:** Ensure the page loads quickly and that the requirements checks do not slow down the initial load.

**Steps:**
- Access the version choice page.
- Measure the loading time until the page is fully displayed.  

**Expected Result:** The page should load without excessive latency, and the requirements checks should not affect loading speed.

### Single Execution of Requirements Check:

**Objective:** Verify that the requirements checks run only once per page load session.

**Steps:**
- Reload the version choice page.
- Confirm that the requirements check only runs once during the initial page load.  

**Expected Result:** The requirements check should execute only once, without repetition.

### Conditional Display of Requirements:

**Objective:** Confirm that requirements are displayed only if there are no errors in the form.

**Steps:**
- Load the version choice page with a form containing no errors.
- Load the version choice page with a form containing errors.

**Expected Result:**
- If the form is valid (no errors): Requirements should be displayed.
- If errors are present in the form: Requirements should not be displayed.

### Error Display After Page Load:

**Objective:** Ensure that errors are displayed if detected at the end of the API call, even after the page has already loaded.

**Steps:**
- Load the page with a form that contains errors detected after the API call completes.

**Expected Result:** Errors should be visible to the user after the page fully loads, following the completion of the API call.

### DIFF
--- a/_dev/src/ts/pages/UpdatePageVersionChoice.ts
+++ b/_dev/src/ts/pages/UpdatePageVersionChoice.ts
@@ -17,6 +17,8 @@ export default class UpdatePageVersionChoice extends UpdatePage {
     this.form.addEventListener('change', this.saveForm.bind(this));
     this.form.addEventListener('submit', this.handleSubmit);
 
+    this.form.dispatchEvent(new Event('change'));
+
     this.onlineCardParent?.addEventListener(Hydration.hydrationEventName, this.handleHydrate);
     this.localCardParent?.addEventListener(Hydration.hydrationEventName, this.handleHydrate);
 
--- a/classes/Parameters/UpgradeFileNames.php
+++ b/classes/Parameters/UpgradeFileNames.php
@@ -33,6 +33,13 @@ namespace PrestaShop\Module\AutoUpgrade\Parameters;
 class UpgradeFileNames
 {
     /**
+     * stateFilename contains all state specific to the autoupgrade module.
+     *
+     * @var string
+     */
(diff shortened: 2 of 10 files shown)


M0rgan01 (dismissed):

M0rgan01 (dismissed):

Quetzacoalt91 (dismissed):

AureRita (approved):
Hi @ga-devfront 

[recording(330).webm](https://github.com/user-attachments/assets/681727e9-c20c-4f12-983e-10a0d2166482)


Tested from : 
1.7.8.8 to 8.2 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

We seen together that when we want to load a local archive from v9 we have an infinite load in check requirement : 

[recording(332).webm](https://github.com/user-attachments/assets/faa26608-c3d8-4446-a8aa-5a660dded6b3)

Because this issue is not related to your PR and this one seems to works as expected, It's QA :heavy_check_mark: 

Thank you

PR #992, opened 2024-11-05
[NEW UI] Add error layouts for 404 and 500 pages

tblivet:
| Questions         | Answers
| Description?      | Add error layouts for 404 and 500 pages in Storybook
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| How to test?      | Check the error layouts in Storybook.


storybook/stories/layouts/Error.stories.js
```diff
@@ -0,0 +1,37 @@
...
+import ErrorLayout from "../../../views/templates/layouts/error.html.twig";
+
+export default {
+  component: ErrorLayout,
+  title: "Layouts/Errors",
+  args: {
+    psBaseUri: "/",
+    error_code: "404",
+  }
+};
```
Quetzacoalt91: Why having this story if the 404 exists as well?

views/templates/errors/404.html.twig
```diff
@@ -0,0 +1,25 @@
...
+  <h2 class="error-page__title">
+    {{ 'Something went wrong...'|trans({}) }}
+  </h2>
+{% endblock %}
+
+{% block description %}
+  <p class="error-page__desc">
+    {{ 'We can’t find your update file in the server.'|trans({}) }}
+    <br/>
+    {{ 'Add a file to restore or update your store.'|trans({}) }}
```
Quetzacoalt91: This is usually caused by something else. I would rather say something like "The address you requested cannot be found... bla bla"

tblivet: I have updated the wording with the text you proposed on Figma, but I think we may need to make some changes later.

views/templates/layouts/error.html.twig
```diff
@@ -0,0 +1,29 @@
...
+      {% endblock %}
+
+      {% block button %}
+        <a class="error-page__button btn btn-primary btn-lg" href="#">
+          {{ 'Go back to Update assistant'|trans({}) }}
+        </a>
+      {% endblock %}
+    </div>
+  </div>
+</div>
```
Quetzacoalt91, line 21: I see this file is like the error 404 template.
Is it necessary to have this duplication? What do you think about having only the skeletons (only the blocks definitions) in this file?

tblivet: I have made some changes to harmonize 'skeleton' across Twig files.

Quetzacoalt91 (approved):

PR #1003, opened 2024-11-08
Improve channel configuration

M0rgan01:
| Questions         | Answers
| Description?      | I moved the default channel configuration to the `UpgradeConfiguration` class, as well as adding some utility methods. Also a bug fix on temporary files
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | -

### DIFF
--- a/classes/Task/Restore/RestoreComplete.php
+++ b/classes/Task/Restore/RestoreComplete.php
@@ -44,6 +44,9 @@ class RestoreComplete extends AbstractTask
     {
         $this->logger->info($this->translator->trans('Restoration process done. Congratulations! You can now reactivate your shop.'));
         $this->next = TaskName::TASK_COMPLETE;
+
+        $this->container->getFileConfigurationStorage()->cleanAllRestoreFiles();
+        $this->container->getFileConfigurationStorage()->cleanAllUpdateFiles();
         $this->container->getAnalytics()->track('Restore Succeeded', Analytics::WITH_RESTORE_PROPERTIES);
 
         $this->container->getState()->setProgressPercentage(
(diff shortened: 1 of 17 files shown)


classes/Task/Restore/RestoreComplete.php
```diff
@@ -44,6 +44,8 @@ public function run(): int
     {
         $this->logger->info($this->translator->trans('Restoration process done. Congratulations! You can now reactivate your shop.'));
         $this->next = TaskName::TASK_COMPLETE;
+
+        $this->container->getFileConfigurationStorage()->cleanAllRestoreFiles();
```
Quetzacoalt91, line 48: I expect a common scenario where we a restore is run after a failed update.
In this case, the update files would not be deleted, and this process would be a fail safe.

```suggestion
        $this->container->getFileConfigurationStorage()->cleanAllRestoreFiles();
        $this->container->getFileConfigurationStorage()->cleanAllUpdateFiles();
```

classes/Task/Update/UpdateComplete.php
```diff
@@ -62,7 +62,7 @@ public function run(): int
 
         $this->next = TaskName::TASK_COMPLETE;
 
-        if ($this->container->getUpgradeConfiguration()->getChannel() != Upgrader::CHANNEL_LOCAL && file_exists($this->container->getFilePath()) && unlink($this->container->getFilePath())) {
+        if ($this->container->getUpgradeConfiguration()->getChannel() != UpgradeConfiguration::CHANNEL_LOCAL && file_exists($this->container->getFilePath()) && unlink($this->container->getFilePath())) {
```
Quetzacoalt91: Worth using your new methods?

```suggestion
        if ($this->container->getUpgradeConfiguration()->isChannelOnline() && file_exists($this->container->getFilePath()) && unlink($this->container->getFilePath())) {
```

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

image
 before 
 
 
image

after

Tested from : 
1.7.8.8 to 8.2 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

Thank you

PR #1019, opened 2024-11-18
[NEW UI] progress tracker script

ga-devfront:
| Questions         | Answers
| Description?      | Start progress tracker component with implement subcomponents
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | N/A
| How to test?      | todo


views/templates/components/progress-bar.html.twig
```diff
@@ -0,0 +1,10 @@
+<div data-component="progress-bar" class="progress log-progress__bar">
+    <div class="progress-bar"
+         role="progressbar"
+         data-title-template="{{ "Progress is {progress_percentage}%"|trans() }}"
```
Quetzacoalt91: I see an empty object sent in all other calls to `trans`, isn't this line triggering any error?

tests/unit/Log/WebLoggerTest.php
```diff
@@ -33,7 +33,7 @@ public function testLastInfoIsRegistered()
         $logger = new WebLogger();
         $logger->log(WebLogger::INFO, 'Hello');
 
-        $this->assertSame('Hello', $logger->getLastInfo());
+        $this->assertSame('INFO - Hello', $logger->getLastInfo());
```
Quetzacoalt91: Following our latest discussion, shall we ignore the severity in this case we cause we know for sure we have an INFO?

_dev/src/ts/components/ComponentAbstract.ts
```diff
@@ -0,0 +1,13 @@
+export default abstract class ComponentAbstract {
+  public element: HTMLElement;
+
+  public constructor(element: HTMLElement) {
+    this.element = element;
+  }
+
+  protected queryElement = <T extends HTMLElement>(selector: string, errorMessage: string): T => {
+    const element = (this.element.querySelector(selector) as T) ?? document.querySelector(selector);
+    if (!element) throw new Error(errorMessage);
```
M0rgan01:
```suggestion
    if (!element) {
       throw new Error(errorMessage);
     }
```

to gain readability and be consistent with the rest of the code

_dev/src/ts/components/ComponentAbstract.ts
```diff
@@ -0,0 +1,13 @@
+export default abstract class ComponentAbstract {
+  public element: HTMLElement;
```
M0rgan01:
```suggestion
  protected element: HTMLElement;
```

_dev/src/ts/components/LogsViewer.ts
```diff
@@ -0,0 +1,232 @@
...
+  /**
+   * @private
+   * @param {MouseEvent} event - The click event object.
+   * @description Handles click events on summary links to scroll to the corresponding log line.
+   * Highlights the target log line briefly for visual focus.
+   */
+  #handleLinkEvent = (event: MouseEvent): void => {
+    const target = event.target as HTMLAnchorElement;
+
+    if (!target || target.tagName !== 'A' || !target.hash) {
```
M0rgan01, line 215: we could extract a constant from 'A', which would also allow us to gain in understanding

_dev/src/ts/components/ComponentAbstract.ts
ga-devfront: If I change it to protected I can't test it and after compilation it changes nothing.

_dev/src/ts/components/LogsViewer.ts
ga-devfront: I don't know if this is the best thing to do and I would like to have your opinion:
on the one hand it makes the code more readable.
however this constant risks being used only here, it adds files and complexity for a tag a which remains simple to understand.

M0rgan01: For a person who comes to the code, and sees this, he does not understand the condition. Adding a constant, a variable or a comment helps with understanding and will allow future contributors to better understand the open-source project.

_dev/src/ts/components/ComponentAbstract.ts
M0rgan01: so let's use a getter

_dev/src/ts/components/LogsViewer.ts
ga-devfront: is it ok for you if I add a comment like `// check that we click on a link that has a link to an anchor` ?

_dev/src/ts/components/ComponentAbstract.ts
ga-devfront: you want to set a private variable and get if from getter right ? That's possibility but this can make the process slightly heavy

M0rgan01: I especially want to avoid having public class variables, it goes against the fundamental principles of encapsulation and modularity in object-oriented programming.

_dev/src/ts/components/LogsViewer.ts
M0rgan01: I especially wanted to highlight what 'A' represents here, a link, representing the name of the HTML tag <a>. a comment like yours or `// Checks if the clicked element is an <a> tag` works very well

_dev/src/ts/utils/logsUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * @public
+ * @param {string} log - The log string to be parsed. Should be in the format: "SEVERITY - Message".
+ * @returns {LogEntry} An object containing the severity class and log message.
+ * @description Parses a log string to extract its severity and corresponding CSS class for styling.
+ */
+export function parseLogWithSeverity(log: string): LogEntry {
+  const logTrimed = log.trim();
+  const severityPattern = severityValues.join('|');
```
Quetzacoalt91: In the future, we may have to review how well the page performs. It won't make a huge difference, but having the performance in mind it would be better to have this value computed one time instead of during each log addition. The result will always be the same.

views/templates/components/progress-bar.html.twig
ga-devfront: I don't want the content of the {} to be replaced by trans but by the JS

Quetzacoalt91: Understood, but my comment was related to a potential mandatory parameter of `trans()`, even when you have nothing to replace.
It this code works, I wonder why we have put `trans({})` everywhere else.

tests/unit/Log/WebLoggerTest.php
ga-devfront: I'm not sure about your return. If I want to have a generic operation, the severity must be added in all cases, right?

Quetzacoalt91: and the severety will always be pushed in the logs. I'm talking here about the specific case of `getLastInfo`, containing the last INFO that has been added (in the logs as well).
Because this value is displayed in the Progress Tracker summary, I think we don't need the severity there, and running the regex will take some unnecessary CPU time.

What do you think?

ga-devfront: Ho yes you're right I going to fix this.

M0rgan01 (approved):

Quetzacoalt91:
This PR can't be tested by the QA team yet because the feature is not completely implemented.

Merging as-is

PR #1021, opened 2024-11-18
[CLI] Create "update:check-new-version" command

M0rgan01:
| Questions         | Answers
| Description?      | Create "update:check-new-version" command
| Type?             |new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | use new `update:check-new-version` command

### DIFF
--- a/classes/Commands/CheckNewVersionCommand.php
+++ b/classes/Commands/CheckNewVersionCommand.php
@@ -1,0 +1,110 @@
+<?php
+
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License 3.0 (AFL-3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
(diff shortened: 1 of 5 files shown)


classes/Commands/CheckNewVersionCommand.php
```diff
@@ -0,0 +1,116 @@
...
+     * @throws ZipActionException
+     */
+    protected function execute(InputInterface $input, OutputInterface $output): ?int
+    {
+        try {
+            $this->setupContainer($input, $output);
+            $this->upgradeContainer->initPrestaShopAutoloader();
+            $this->upgradeContainer->initPrestaShopCore();
+
+            $this->upgradeContainer->getState()->initDefault(
```
ga-devfront: we going to create a state file only to get current bersion but you already pass it to this method init with `$this->upgradeContainer->getProperty(UpgradeContainer::PS_VERSION)` I think we can remove this part and pass directly $this->upgradeContainer->getProperty(UpgradeContainer::PS_VERSION)

ga-devfront (dismissed):
I think it's ok without the little comment I left

classes/Commands/CheckNewVersionCommand.php
M0rgan01: this part was not necessary for the process, I removed it

ga-devfront (dismissed):

Quetzacoalt91:
Just found out an issue:

image

The online channel should be first in the list, whatever its destination version compared to the local updates.

ga-devfront (approved):

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

image

Tested from : 
1.7.8.8 to 8.2 
8.1.4 to 8.2 
8.2 to 9.0.0

Thank you

PR #1032, opened 2024-11-22
[NEW UI] Feat : progress tracker component

ga-devfront:
| Questions         | Answers
| Description?      | Complete implementation of the progress tracker in the update page. The component is functional and better structures the way calls are made for 'actions'
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | Navigate to update page from new UI and see new progress tracker. If you want to see the summary you need to cause a blocking error during the process.

What this PR does not include:
- The new mockup styles
- The correct link to download the logs
- Full functionality of the udpate page in case of error.

### DIFF
--- a/_dev/src/ts/components/LogsViewer.ts
+++ b/_dev/src/ts/components/LogsViewer.ts
@@ -28,6 +28,10 @@ export default class LogsViewer extends ComponentAbstract {
     'Template summary not found'
   );
 
+  public beforeDestroy = () => {
+    this.#logsSummary.removeEventListener('click', this.#handleLinkEvent);
+  };
+
   /**
    * @public
    * @param {string[]} logs - Array of log strings to be added.
--- a/_dev/src/ts/components/LogsViewer.ts
+++ b/_dev/src/ts/components/LogsViewer.ts
@@ -1,8 +1,9 @@
 import ComponentAbstract from './ComponentAbstract';
 import { SeverityClasses, LogEntry } from '../types/logsTypes';
 import { parseLogWithSeverity } from '../utils/logsUtils';
+import { Destroyable } from '../types/DomLifecycle';
 
-export default class LogsViewer extends ComponentAbstract {
+export default class LogsViewer extends ComponentAbstract implements Destroyable {
   #warnings: string[] = [];
   #errors: string[] = [];
   #isSummaryDisplayed: boolean = false;
@@ -26,6 +27,10 @@ export default class LogsViewer extends ComponentAbstract {
   #templateSummary = this.queryElement<HTMLTemplateElement>(
     '#log-summary',
     'Template summary not found'
+  );
+  #templateSummaryButtons = this.queryElement<HTMLTemplateElement>(
+    '#summary-buttons',
+    'Template summary buttons not found'
   );
 
   public beforeDestroy = () => {
@@ -72,12 +77,13 @@ export default class LogsViewer extends ComponentAbstract {
 
   /**
    * @public
+   * @param {string | null} downloadLogsLink - The link to download update logs.
    * @description Displays a summary of logs, grouping warnings and errors.
    * Summaries include links to the corresponding log lines.
    * Adds a click event listener to handle navigation within the summary.
    * Prevents displaying a summary if no logs are present.
    */
-  public displaySummary(): void {
+  public displaySummary(downloadLogsLink?: string): void {
     if (!this.#logsList.hasChildNodes()) {
       console.warn('Cannot display summary because logs are empty');
       return;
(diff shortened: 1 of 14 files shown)


_dev/src/ts/components/LogsViewer.ts
```diff
@@ -68,12 +77,13 @@ export default class LogsViewer extends ComponentAbstract {
...
   /**
    * @public
+   * @param {string | null} downloadLogsLink - The link to download update logs.
    * @description Displays a summary of logs, grouping warnings and errors.
    * Summaries include links to the corresponding log lines.
    * Adds a click event listener to handle navigation within the summary.
    * Prevents displaying a summary if no logs are present.
    */
-  public displaySummary(): void {
+  public displaySummary(downloadLogsLink?: string): void {
```
Quetzacoalt91: Should we really allow a case where `downloadLogsLink` is `null`?

_dev/src/ts/components/LogsViewer.ts
```diff
@@ -211,7 +250,7 @@ export default class LogsViewer extends ComponentAbstract {
   #handleLinkEvent = (event: MouseEvent): void => {
     const target = event.target as HTMLAnchorElement;
 
-    // Checks if the clicked element is an <a> tag
+    // Checks if the clicked element is an <a> tag point to and ID
```
Quetzacoalt91: // Checks if the clicked element is an `<a>` tag pointing towards an ID

_dev/src/ts/components/ProgressTracker.ts
```diff
@@ -0,0 +1,63 @@
...
+    this.#logsViewer.addLogs(data.nextQuickInfo);
+  }
+
+  /**
+   * @public
+   * @returns {void}
+   * @description Ends the progress tracking and displays a summary of error logs in the logs viewer.
+   */
+  public endProgress(): void {
+    // Todo: we need to retriew the download link
```
Quetzacoalt91: Not really important but while we have other things to change

```suggestion
    // Todo: we need to retrieve the download link
```

_dev/src/ts/components/ProgressTracker.ts
```diff
@@ -0,0 +1,63 @@
...
+   * @param {ApiResponseAction} data - API response data containing the progress information,
+   * next description, and log details.
+   * @returns {void}
+   * @description Updates the progress tracker components:
+   * - Updates the log summary with the next description.
+   * - Updates the progress bar with the progress percentage.
+   * - Adds new logs to the logs viewer.
+   */
+  public updateProgress(data: ApiResponseAction): void {
+    this.#logsSummary.setLogsSummaryText(String(data.next_desc));
```
Quetzacoalt91: Why casting to String as you're already have a string when not null?

You probably meant something like `data.next_desc || [default value]` ?

_dev/src/ts/pages/UpdatePageUpdate.ts
```diff
@@ -1,13 +1,61 @@
...
 
   constructor() {
     super();
   }
 
-  public mount() {
+  public mount = async () => {
     this.initStepper();
+
+    const updateAction = 'UpdateInitialization';
```
Quetzacoalt91, line 18: Maybe worth having this information sent by the back end in order to use the constant `TaskName::TASK_UPDATE_INITIALIZATION`?

_dev/src/ts/pages/UpdatePageUpdate.ts
```diff
@@ -1,13 +1,61 @@
...
+
+    const process = new Process({
+      onProcess: this.#onProcess,
```
Quetzacoalt91: Can we find a more specific name for this? Because I did not understand what part of the process it is talking about.

Something like `onProcessUpdate`, `onProcessStep`,  `onResponse`...

_dev/src/ts/pages/UpdatePageUpdate.ts
```diff
@@ -1,13 +1,61 @@
...
+  get #progressTrackerContainer(): HTMLDivElement {
+    const progressTrackerContainer = document.querySelector(
+      '[data-component="progress-tracker"]'
+    ) as HTMLDivElement;
+
+    if (!progressTrackerContainer) {
+      throw new Error('Progress tracker container not found');
+    }
+
+    return progressTrackerContainer;
```
Quetzacoalt91, line 42: Any reason for not using the `queryElement`?

```suggestion
    return this.queryElement<HTMLTemplateElement>(
      '[data-component="progress-tracker"]'
      'Progress tracker container not found'
    );
```

If this is because it does not extends `ComponentAbstract`, it could be nice to make PageAbstract extending it to get this feature.
This would make some changes on all existing pages, this can be done in another PR.

_dev/src/ts/pages/UpdatePageUpdate.ts
ga-devfront: I going to do it in another PR

_dev/src/ts/pages/UpdatePageUpdate.ts
ga-devfront: Yes I going to check this in another PR

_dev/src/ts/components/LogsViewer.ts
```diff
@@ -95,10 +105,39 @@ export default class LogsViewer extends ComponentAbstract {
       this.#logsSummary.addEventListener('click', this.#handleLinkEvent);
     }
 
+    if (downloadLogsLink) {
```
gericfo: @ga-devfront The `if()`  is always true, you can remove it.

Quetzacoalt91 (approved):

_dev/src/ts/types/Process.ts
```diff
@@ -0,0 +1,9 @@
+import { ApiResponseAction } from './apiTypes';
+
+type Callbacks = {
+  onProcessResponse: (response: ApiResponseAction) => void | Promise<void>;
+  onProcessEnd: (response: ApiResponseAction) => void | Promise<void>;
+  onError: (response: ApiResponseAction) => void | Promise<void>;
+};
```
M0rgan01, line 7: Elsewhere we declare our types with the keyword `interface`, is there a reason to use `type` here?

_dev/src/ts/components/ProgressTracker.ts
```diff
@@ -0,0 +1,63 @@
...
+      'Progress bar not found'
+    );
+  }
+
+  get #logsViewerContainer() {
+    return this.queryElement<HTMLDivElement>(
+      '[data-component="logs-viewer"]',
+      'Logs viewer not found'
+    );
+  }
```
M0rgan01, line 36: We could add a return type

ga-devfront: we already got it implicitly if you check `this.queryElement<HTMLDivElement>` it's what we return then the TS know we going to return a `HTMLDivElement`

_dev/src/ts/types/Process.ts
ga-devfront: `interface` is more useful for complex objects or those that need to be extended. In this case there is no risk of extending it, because it is used as a parameter of a function.

M0rgan01 (approved):

PR #1033, opened 2024-11-22
Add check on headers

Quetzacoalt91:
| Questions         | Answers
| Description?      | Check each file being distributed has an up-to-date header about AFL-3.0. This PR adds a first commit showing the check running properly, then another commit will be added before the release of the version 7 to make sure everything is set without bringing too many conflicts. 
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | CI must be triggered and must succeed


.github/workflows/build-release.yml
```diff
@@ -30,6 +30,13 @@ jobs:
           npm install --prefix ./_dev
           npm run vite:build --prefix ./_dev
 
+      - name: Install composer dev dependencies for tests
+        working-directory: tests
+        run: composer install --ansi --prefer-dist --no-interaction --no-progress
+
+      - name: Apply header stamps
+        run: php tests/vendor/bin/header-stamp --license=tests/vendor/prestashop/header-stamp/assets/afl.txt --exclude=vendor,node_modules,storybook/var
```
Quetzacoalt91: What's the point in applying the headers on all the files if we had a check before merging pull-requests?

If you meant to update built files before release, it may be worth a comment.

M0rgan01 (approved):

PR #1036, opened 2024-11-24
Fix: Autoupgrade attempts to update non-installed modules, causing the upgrade process to fail

### LINKED ISSUE #37445: Autoupgrade attempts to update non-installed modules, causing the upgrade process to fail
### Prerequisites

When updating Prestashop using the "autoupgrade" module, the system attempts to update modules located in the "modules" folder even if they are not actually installed.

I encountered this issue while updating a website that had the "ps_mbo" module in the "modules" folder, but the module was not installed.

The upgrade procedure fails to complete and stops at the step:

```
=== Step upgradeModules
Updating module ps_mbo...
Module ps_mbo has been successfully downloaded from Addons.
The content of the archive \prestashop_prova_aggiornamento\**admin_folder**\autoupgrade\tmp\ps_mbo.zip has been extracted 
(1/11) Applying migration file upgrade-2.0.0.php.
```
To complete the upgrade, I had to delete the module's folder, after which the procedure was successfully finalized.

In this case, the issue arises because the module update process cannot be completed as the module is not installed.
I believe the upgrade should only process modules that are actually installed.

The list of modules to be "upgraded" is retrieved here without verifying whether these modules are really installed:

https://github.com/PrestaShop/autoupgrade/blob/4ea36f0b3f668b54dddf775ed58dd54aec2542f9/classes/Task/Upgrade/UpgradeModules.php#L62-L75

Image

I believe that here we should check if the module is actually installed.

https://github.com/PrestaShop/autoupgrade/blob/4ea36f0b3f668b54dddf775ed58dd54aec2542f9/classes/Task/Upgrade/UpgradeModules.php#L96-L103

### Expected behavior

_No response_

### Steps to reproduce

1. Install autoupgrade module
2. Copy the ps_mbo module into the modules folder without installing it.
3. Perform the upgrade to the latest version of Prestashop.

8.x

_No response_

autoupgrade

@codencode

Codencode:
Added a check for the module version, ensuring that only the modules actually installed are updated.

| Questions         | Answers
| Description?      | When updating Prestashop using the "autoupgrade" module, the system attempts to update modules located in the "modules" folder even if they are not actually installed.
| Type?             | bug fix 
| BC breaks?        |  no
| Deprecations?     |  no
| Fixed ticket?     | Fixes [#37445](https://github.com/PrestaShop/PrestaShop/issues/37445)
| Sponsor company   | @Codencode
| How to test?      | 1. Install autoupgrade module - 2. Copy the ps_mbo module into the modules folder without installing it. - 3. Perform the upgrade to the latest version of Prestashop.

### DIFF
--- a/classes/Task/Update/UpdateModules.php
+++ b/classes/Task/Update/UpdateModules.php
@@ -92,11 +92,11 @@ class UpdateModules extends AbstractTask
 
                     if ($dbVersion !== '') {
                         $module = \Module::getInstanceByName($moduleInfos['name']);
-    
+
                         if (!($module instanceof \Module)) {
                             throw (new UpgradeException($this->translator->trans('[WARNING] Error when trying to retrieve module %s instance.', [$moduleInfos['name']])))->setSeverity(UpgradeException::SEVERITY_WARNING);
                         }
-    
+
                         $moduleMigrationContext = new ModuleMigrationContext($module, $dbVersion);
     
                         if (!$moduleMigration->needMigration($moduleMigrationContext)) {
--- a/classes/Task/Update/UpdateModules.php
+++ b/classes/Task/Update/UpdateModules.php
@@ -89,20 +89,23 @@ class UpdateModules extends AbstractTask
                     $moduleUnzipper->unzipModule($moduleUnzipperContext);
 
                     $dbVersion = (new ModuleVersionAdapter())->get($moduleInfos['name']);
-                    $module = \Module::getInstanceByName($moduleInfos['name']);
 
-                    if (!($module instanceof \Module)) {
-                        throw (new UpgradeException($this->translator->trans('[WARNING] Error when trying to retrieve module %s instance.', [$moduleInfos['name']])))->setSeverity(UpgradeException::SEVERITY_WARNING);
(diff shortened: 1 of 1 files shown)


Hlavtox:
Hey @Codencode, I am thinking about this and I am not sure if it's the correct way to handle this.

I think that native modules that are on the disk should be updated, because they could break the module page after the upgrade. Many modules are instantiated on the module page - even if not installed.

Codencode:
> Hey @Codencode, I am thinking about this and I am not sure if it's the correct way to handle this.
> 
> I think that native modules that are on the disk should be updated, because they could break the module page after the upgrade. Many modules are instantiated on the module page - even if not installed.


Hi @Hlavtox,

The check I implemented is meant to prevent the module from being updated at the database level, meaning that the scripts in the module's **upgrade** folder are not executed.

In practice, the module is still downloaded, and the files are updated, but if the module is not actually installed, the database update is not performed.
I believe this behavior is correct: it wouldn’t make sense to execute database operations for a module that is not truly installed.

Codencode:
The issue has been resolved in the "dev" version, so it should no longer exist in the next release.

PR #1046, opened 2024-11-27
Improve state management

M0rgan01:
| Questions         | Answers
| Description?      | Improve state management
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | Previous configuration are no longer taken into account, log files are no longer cut during an update

### DIFF
--- a/classes/Commands/UpdateCommand.php
+++ b/classes/Commands/UpdateCommand.php
@@ -31,6 +31,7 @@ use Exception;
 use InvalidArgumentException;
 use PrestaShop\Module\AutoUpgrade\DeveloperDocumentation;
 use PrestaShop\Module\AutoUpgrade\Parameters\UpgradeConfiguration;
+use PrestaShop\Module\AutoUpgrade\Parameters\UpgradeFileNames;
 use PrestaShop\Module\AutoUpgrade\Task\ExitCode;
 use PrestaShop\Module\AutoUpgrade\Task\Runner\AllUpdateTasks;
 use PrestaShop\Module\AutoUpgrade\Task\TaskName;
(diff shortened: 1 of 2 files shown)


classes/State.php
```diff
@@ -298,36 +303,44 @@ public function getRestoreVersion(): ?string
...
         return $this;
     }
 
     public function setBackupDbFilename(string $backupDbFilename): State
     {
         $this->backupDbFilename = $backupDbFilename;
+        $this->save();
 
         return $this;
     }
```
ga-devfront: you can change this methods to private method and only save on `setBackupName`

ga-devfront (changes requested):

Quetzacoalt91 (dismissed):

ga-devfront (dismissed):

classes/Commands/UpdateCommand.php
```diff
@@ -127,17 +129,17 @@ private function chainCommand(OutputInterface $output): int
             return ExitCode::SUCCESS;
         }
 
-        if (strpos($lastInfo, 'bin/console update:start') !== false) {
+        if (strpos($lastInfo, self::$defaultName) !== false) {
```
M0rgan01, line 134: this posed a problem if options were placed before the command (ex: `bin/console -v update:start`)

classes/Commands/UpdateCommand.php
```diff
@@ -83,6 +84,7 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
             if ($isFirstUpdateProcess) {
                 $this->logger->debug('Cleaning previous state files.');
                 $this->upgradeContainer->getFileConfigurationStorage()->cleanAllUpdateFiles();
+                $this->upgradeContainer->getFileConfigurationStorage()->clean(UpgradeFileNames::CONFIG_FILENAME);
```
M0rgan01, line 92: We could end up with the config of an old update

ga-devfront (approved):

Quetzacoalt91:
Result of the logs folder after an update:

```
root@b653c57bea07:/var/www/html/admin-dev/autoupgrade/logs# ls -lh
total 3.3M
-rw-r--r-- 1 www-data www-data 3.3M Dec  3 15:35 2024-12-03-163132-update.txt
```

And the log file has the whole process. We're good to merge!

PR #1050, opened 2024-11-28
[NEW UI] Check wording parts 1

tblivet:
| Questions         | Answers
| Description?      | Check wording (first session) for already functional parts
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| How to test?      | -

### DIFF
--- a/classes/UpgradeSelfCheck.php
+++ b/classes/UpgradeSelfCheck.php
@@ -327,7 +327,7 @@ class UpgradeSelfCheck
 
             case self::NOT_WRITING_DIRECTORY_LIST_NOT_EMPTY:
                 return [
-                    'message' => $this->translator->trans('It\'s not possible to write in the following folders: Provide write access to the user running PHP with appropriate permission & ownership: '),
+                    'message' => $this->translator->trans('It\'s not possible to write in the following folders. Provide write access to the user running PHP with appropriate permission & ownership: '),
                     'list' => $this->getNotWritingDirectories(),
                 ];
 
--- a/classes/UpgradeSelfCheck.php
+++ b/classes/UpgradeSelfCheck.php
@@ -213,18 +213,25 @@ class UpgradeSelfCheck
             case self::PHP_COMPATIBILITY_INVALID:
                 return [
                     'message' => $this->translator->trans(
-                        'Your current PHP version isn\'t compatible with your PrestaShop version. (Expected: %s - %s | Current: %s)',
-                        [$phpCompatibilityRange['php_min_version'], $phpCompatibilityRange['php_max_version'], $phpCompatibilityRange['php_current_version']]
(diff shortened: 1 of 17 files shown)


classes/UpgradeSelfCheck.php
```diff
@@ -213,18 +213,25 @@ public function getRequirementWording(int $requirement, bool $isWebVersion = fal
...
+                        [$this->prodRootPath]
+                    ),
                 ];
 
             case self::ADMIN_UPGRADE_DIRECTORY_NOT_WRITABLE:
                 return [
-                    'message' => $this->translator->trans('The "/admin/autoupgrade" directory isn\'t writable. Provide write access to the user running PHP with appropriate permission & ownership.'),
+                    'message' => $this->translator->trans(
+                        'The %s directory isn\'t writable. Provide write access to the user running PHP with appropriate permission & ownership.',
+                        [$this->autoUpgradePath]
```
Quetzacoalt91, line 234: Is the path displayed is sanitized? We should make sure the actual admin folder is not displayed to keep the merchants data safe in case he makes a screenshot of this page.

classes/UpgradeSelfCheck.php
```diff
@@ -296,14 +319,15 @@ public function getRequirementWording(int $requirement, bool $isWebVersion = fal
...
+
             case self::KEY_GENERATION_INVALID:
                 return [
                     'message' => $this->translator->trans('Unable to generate private keys using openssl_pkey_new. Check your OpenSSL configuration, especially the path to openssl.cafile.'),
                 ];
 
             case self::NOT_WRITING_DIRECTORY_LIST_NOT_EMPTY:
                 return [
-                    'message' => $this->translator->trans('It\'s not possible to write in the following folders, please provide write access to the user running PHP with appropriate permission & ownership: '),
+                    'message' => $this->translator->trans('It\'s not possible to write in the following folders: Provide write access to the user running PHP with appropriate permission & ownership: '),
```
Quetzacoalt91: What about?

```suggestion
                    'message' => $this->translator->trans('It\'s not possible to write in the following folders. Provide write access to the user running PHP with appropriate permission & ownership: '),
```

classes/UpgradeSelfCheck.php
tblivet: Ok, so we should go with `/your-admin-directory/autoupgrade/` ?

Quetzacoalt91: Yep, one good way would be to use the sanitizer currently present in the Logger class.

Quetzacoalt91: If you can't do this in this PR, please add a `TODO`

tblivet: I've added a TODO, I think we need to wait for a response to the wording comment in Figma before proceeding?

Quetzacoalt91 (approved):

PR #1062, opened 2024-12-03
UI Test - Add v9.1 to the possible directory 

AureRita:
| Questions         | Answers
| Description?      | I add a v9.1 to the different directory and add it to sanity to verify if it works 
| BC breaks?        |  no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | Prestashop Core 
| How to test?      | CI is green 

Blocked by PrestaShop/PrestaShop#37516 (by @boherm)

### DIFF
--- a/.github/workflows/ui-test/sanity.json
+++ b/.github/workflows/ui-test/sanity.json
@@ -107,6 +107,12 @@
       "PS_VERSION_END": "9.0.0",
       "PHP_VERSION": "8.1",
       "UPGRADE_CHANNEL":  "local"
+    },
+    {
+      "PS_VERSION_START": "8.1.7",
+      "PS_VERSION_END": "9.1.0",
+      "PHP_VERSION": "8.1",
+      "UPGRADE_CHANNEL":  "local"
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


Progi1984 (approved):

.github/workflows/ui-test/sanity.json
```diff
@@ -107,6 +107,12 @@
       "PS_VERSION_END": "9.0.0",
       "PHP_VERSION": "8.1",
       "UPGRADE_CHANNEL":  "local"
+    },
+    {
+      "PS_VERSION_START": "8.1.7",
+      "PS_VERSION_END": "9.1.0",
```
Quetzacoalt91, line 113: @AureRita If you plan to keep this test in the future, I would recommend to rename this destination version as `NIGHTLY`, to avoid modifying the value on each release.

Quetzacoalt91:
With all the changes brought in the project in the meantime and the incoming #1421, it is probably better to prepare a new case for nightly builds without the destination version.
I recommend starting over from the latest changes of `dev` once the PR is merged if the test is still relevant.

PR #1075, opened 2024-12-10
Set new option of lang prefix dynamically

Hlavtox:
| Questions         | Answers
| Description?      | Sets the new option of default language URL prefixes dynamically. More info below. Related PR - https://github.com/PrestaShop/PrestaShop/pull/37592.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      | 

### Description
- If someone will be running a shop for a while and he adds another language, if this option is ON (or previous behavior), it will highly damage his SEO, because all URLs will permanently change.
- So, if there are multiple active langs on the shop (prefixes for all langs) > upgrade to YES. No URL changes.
- If there is only single language on the shop (nonprefixed URLs) > upgrade to NO. So when he adds a language, his URLs will remain the same. Unless he decides otherwise.

### DIFF
--- a/upgrade/php/ps_900_set_url_lang_prefix.php
+++ b/upgrade/php/ps_900_set_url_lang_prefix.php
@@ -1,0 +1,45 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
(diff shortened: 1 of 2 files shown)


Quetzacoalt91 (approved):

M0rgan01 (approved):

AureRita (approved):
Hi @Hlavtox 

[recording(25).webm](https://github.com/user-attachments/assets/b8276e97-133a-4491-8cf1-c53412105fdb)


[recording(24).webm](https://github.com/user-attachments/assets/7388161d-f43e-4a0f-8ca2-f674af683471)

Tested from : 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

As you can see, after the Upgrade to v9.0.0 we still have : PS_DEFAULT_LANGUAGE_URL_PREFIX 
but it's because it's not already set to : 1 

Thank you

Hlavtox:
Thank you @AureRita :-) So, works like it should. Thanks!

PR #1087, opened 2024-12-16
Ensure logs use the correct timezone in CLI and Web contexts

M0rgan01:
| Questions         | Answers
| Description?      | see below
| Type?             | feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | See below

This PR updates the logging behavior to ensure timestamps are correctly aligned with the execution context:

- CLI Context: Logs will use the system's timezone.
- Web Context: Logs will use the timezone configured in the PrestaShop settings.

### DIFF
--- a/classes/Commands/AbstractCommand.php
+++ b/classes/Commands/AbstractCommand.php
@@ -81,6 +81,9 @@ abstract class AbstractCommand extends Command
         define('_PS_ADMIN_DIR_', $adminDir);
 
         $this->upgradeContainer = new UpgradeContainer($prodRootDir, $adminDir);
+        // We need to store the timezone this early because it can be altered by the core initialization later.
+        $this->upgradeContainer->getLogsState()->setTimeZone(date_default_timezone_get());
+
         $this->logger->debug('Update container initialized.');
 
         $this->logger->debug('Logger initialized: ' . get_class($this->logger));
--- a/classes/Task/Runner/ChainedTasks.php
+++ b/classes/Task/Runner/ChainedTasks.php
@@ -29,6 +29,8 @@ namespace PrestaShop\Module\AutoUpgrade\Task\Runner;
 
 use Exception;
 use PrestaShop\Module\AutoUpgrade\AjaxResponse;
+use PrestaShop\Module\AutoUpgrade\DbWrapper;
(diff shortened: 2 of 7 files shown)


nicosomb (dismissed):

Quetzacoalt91 (dismissed):

classes/Commands/AbstractCommand.php
```diff
@@ -81,6 +81,8 @@ protected function setupEnvironment(InputInterface $input, OutputInterface $outp
         define('_PS_ADMIN_DIR_', $adminDir);
 
         $this->upgradeContainer = new UpgradeContainer($prodRootDir, $adminDir);
+        $this->upgradeContainer->getLogsState()->setTimeZone(date_default_timezone_get());
```
Quetzacoalt91, line 85:
```suggestion
        // We need to store the timezone this early because it can be altered by the core initialization later.
        $this->upgradeContainer->getLogsState()->setTimeZone(date_default_timezone_get());
```

classes/Task/Runner/ChainedTasks.php
```diff
@@ -112,21 +114,41 @@ protected function checkIfRestartRequested(AjaxResponse $response): bool
...
+    /**
+     * @throws Exception
+     */
     private function setupLogging(): void
     {
+        $logsState = $this->container->getLogsState();
+        $timeZone = $logsState->getTimeZone();
+        if ($timeZone) {
+            date_default_timezone_set($timeZone);
+        }
```
Quetzacoalt91: What about moving this part as a `else` of the test `if (in_array($this->step, $initializationSteps)) {`?

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

[recording(63).webm](https://github.com/user-attachments/assets/76d4b29b-e783-4f1a-92f8-2c4e23a21e56)


Tested from : 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

Currently we see that we have two hour different between the CLI and the web as you said it, because of that, we have not really the last rollback in the top of that as we can see in this sceenshot : 

image

Moreover, on the video we can see that there is some trouble with the logs that I can't reproduce so we'll investigate on it not in this PR 

Thank you

PR #1098, opened 2024-12-20
Improve archive download

M0rgan01:
| Questions         | Answers
| Description?      | Add retry mechanic when archive download. Add empty file detection.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | -

### DIFF
--- a/classes/Services/DownloadService.php
+++ b/classes/Services/DownloadService.php
@@ -1,0 +1,92 @@
+<?php
+
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License 3.0 (AFL-3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ */
+
+namespace PrestaShop\Module\AutoUpgrade\Services;
+
+use PrestaShop\Module\AutoUpgrade\Exceptions\UpgradeException;
+use PrestaShop\Module\AutoUpgrade\Log\Logger;
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator;
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
+
+class DownloadService
+{
+    const MAX_DOWNLOAD_TRY = 3;
+    const WAIT_BETWEEN_RETRY_IN_SECONDS = 2;
+
+    /** @var Logger */
+    private $logger;
+    /** @var Translator */
+    private $translator;
+
+    public function __construct(Translator $translator, Logger $logger)
+    {
+        $this->translator = $translator;
+        $this->logger = $logger;
+    }
+
+    /**
+     * @throws UpgradeException
+     */
+    public function downloadWithRetry(string $downloadUrl, string $destinationPath, int $retryCount = self::MAX_DOWNLOAD_TRY, int $delayInSeconds = self::WAIT_BETWEEN_RETRY_IN_SECONDS): void
+    {
+        $attempt = 0;
+
+        while ($attempt < $retryCount) {
+            ++$attempt;
+
+            try {
+                $this->download($downloadUrl, $destinationPath);
+
+                return;
+            } catch (IOException $exception) {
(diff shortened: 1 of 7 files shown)


classes/Task/Update/Download.php
```diff
@@ -41,6 +43,7 @@
 class Download extends AbstractTask
 {
     const TASK_TYPE = TaskType::TASK_TYPE_UPDATE;
+    const MAX_DOWNLOAD_TRY = 3;
```
ga-devfront: You can remove this variable, you don't use it anymore.

classes/Services/DownloadService.php
```diff
@@ -0,0 +1,64 @@
...
+
+    public function download(string $destinationPath, string $downloadUrl): bool
+    {
+        try {
+            $filesystem = new Filesystem();
+            $filesystem->copy($downloadUrl, $destinationPath);
+        } catch (IOException $e) {
+            // If the Symfony filesystem failed, we can try with
+            // the legacy method which uses curl.
+            Tools14::copy($downloadUrl, $destinationPath);
```
Quetzacoalt91: Now that we have the retry system, it may interesting to remove this fallback and eventually allow the method to throw all the necessary exceptions? (Removing the try / catch block + throwing one new exception if the file is empty?

The method signature would now return `void`,and throw the exception in case of error.

classes/Task/Update/Download.php
```diff
@@ -69,30 +70,39 @@ public function run(): int
...
+    {
+        $destPath = realpath($this->container->getProperty(UpgradeContainer::DOWNLOAD_PATH)) . DIRECTORY_SEPARATOR . $this->container->getProperty(UpgradeContainer::ARCHIVE_FILENAME);
+        $archiveUrl = $this->container->getUpgrader()->getOnlineDestinationRelease()->getZipDownloadUrl();
+        $res = $this->container->getDownloadService()->downloadWithRetry($destPath, $archiveUrl);
+        if ($res) {
+            $md5file = md5_file(realpath($this->container->getProperty(UpgradeContainer::ARCHIVE_FILEPATH)));
+            if ($md5file == $this->container->getUpgrader()->getOnlineDestinationRelease()->getZipMd5()) {
+                $this->next = TaskName::TASK_UNZIP;
+                $this->logger->debug($this->translator->trans('Download complete.'));
+                $this->logger->info($this->translator->trans('Download complete. Now extracting...'));
```
Quetzacoalt91: Shall we take the opportunity to simplify the contents sent to the logger?
I guess it was necessary before because the info was in a separate panel from the rest of the logs, but as it is not the case anymore with the v7, the call to `info()` seems enough.

classes/Task/Update/Download.php
```diff
@@ -69,30 +70,39 @@ public function run(): int
...
+            } else {
+                $this->logger->error($this->translator->trans('Download complete but MD5 sum does not match (%s).', [$md5file]));
+                $this->logger->info($this->translator->trans('Download complete but MD5 sum does not match (%s). Operation aborted.', [$md5file]));
```
Quetzacoalt91: Same comment here about the duplicate contents sent to the logger, the `error()` would be enough.

Quetzacoalt91 (dismissed):

classes/UpgradeTools/Module/ModuleDownloader.php
```diff
@@ -90,6 +95,7 @@ private function attemptDownload(ModuleDownloaderContext $moduleDownloaderContex
 
         if ($moduleSource->isZipped()) {
             $destinationPath .= '/' . $moduleDownloaderContext->getModuleName() . '.zip';
+            $this->downloadService->downloadWithRetry($destinationPath, $moduleSource->getPath());
```
ga-devfront: if we fail we don't need to continue with copy no ? I think a message in loger can be usefull too.

classes/Services/DownloadService.php
```diff
@@ -0,0 +1,90 @@
...
+
+            try {
+                $this->download($destinationPath, $downloadUrl);
+
+                return;
+            } catch (IOException $exception) {
+                $this->logger->debug($this->translator->trans('Download attempt %d/%d failed: %s', [$attempt, $retryCount, $exception->getMessage()]));
+            }
+
+            $this->wait($delayInSeconds);
```
Quetzacoalt91: No need to wait after the last failed download

classes/UpgradeContainer.php
```diff
@@ -891,6 +896,18 @@ public function getPrestashopVersionService(): PrestashopVersionService
...
+    /**
+     * @return DownloadService
+     */
+    public function getDownloadService(): DownloadService
+    {
+        if (null !== $this->downloadService) {
+            return $this->downloadService;
+        }
+
+        return $this->downloadService = new DownloadService($this->getTranslator(), $this->logger);
```
Quetzacoalt91:
```suggestion
        return $this->downloadService = new DownloadService($this->getTranslator(), $this->getLogger());
```

Quetzacoalt91 (dismissed):

Quetzacoalt91 (approved):

Quetzacoalt91:
I've been playing with the wifi by turning it on & off during the update of modules, I can confirm the feature work properly.

```
DEBUG - Step UpdateModules
DEBUG - Checking updates of module statsproduct...
DEBUG - Download attempt 1/3 failed: Failed to copy "https://api.addons.prestashop.com/?id_module=15280&method=module&version=8.2.0" to "/var/www/html/modules/autoupgrade/bin/../../../admin-dev/autoupgrade/tmp/statsproduct.zip" because source file could not be opened for reading.
NOTICE - Module statsproduct update files (2.1.1 => 2.1.3) have been fetched from https://api.addons.prestashop.com/?id_module=15280&method=module&version=8.2.0.
DEBUG - Content of archive /var/www/html/modules/autoupgrade/bin/../../../admin-dev/autoupgrade/tmp/statsproduct.zip is extracted
INFO - Module statsproduct does not need to be migrated. Module is up to date.
INFO - 15 modules left to check.
```

PR #1103, opened 2025-01-03
[LOG] Improve logs wording, add classname in log files

M0rgan01:
| Questions         | Answers
| Description?      | Improve logs wording, add classname in log files
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | -

### DIFF
--- a/classes/Log/Logger.php
+++ b/classes/Log/Logger.php
@@ -175,9 +175,37 @@ abstract class Logger implements LoggerInterface
      */
     public function log($level, string $message, array $context = []): void
     {
+        $className = $this->getCallingClass();
+
         if (is_resource($this->fd)) {
-            fwrite($this->fd, '[' . date('Y-m-d H:i:s') . '] ' . self::$levels[$level] . ' - ' . $message . PHP_EOL);
+            fwrite(
+                $this->fd,
+                '[' . date('Y-m-d H:i:s') . '] ' . self::$levels[$level] . ' - ' .
+                ($className ? $className . ' - ' : '') .
+                $message . PHP_EOL
+            );
         }
+    }
+
+    /**
+     * Get the name of the class that called the logger.
+     */
+    private function getCallingClass(): ?string
+    {
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+
+        foreach ($trace as $frame) {
+            if (
+                isset($frame['class']) &&
+                !is_subclass_of($frame['class'], self::class) &&
+                $frame['class'] !== self::class
+            ) {
+                // Extract the class name without the namespace
+                return substr(strrchr($frame['class'], '\\') ?: $frame['class'], 1);
+            }
+        }
+
+        return null;
     }
 
     /**
--- a/classes/Log/Logger.php
+++ b/classes/Log/Logger.php
@@ -192,13 +192,12 @@ abstract class Logger implements LoggerInterface
      */
     private function getCallingClass(): ?string
     {
-        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 8);
 
         foreach ($trace as $frame) {
             if (
                 isset($frame['class']) &&
-                !is_subclass_of($frame['class'], self::class) &&
-                $frame['class'] !== self::class
+                strpos($frame['class'], __NAMESPACE__) === false
             ) {
                 // Extract the class name without the namespace
                 return substr(strrchr($frame['class'], '\\') ?: $frame['class'], 1);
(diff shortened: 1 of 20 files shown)


classes/Task/Backup/BackupFiles.php
```diff
@@ -51,8 +51,7 @@ public function run(): int
         if (empty($backupFilesFilename)) {
             $this->next = TaskName::TASK_ERROR;
             $this->setErrorFlag();
-            $this->logger->info($this->translator->trans('Error during backupFiles'));
-            $this->logger->error($this->translator->trans('[ERROR] backupFiles filename has not been set'));
+            $this->logger->error($this->translator->trans('Backup files filename has not been set'));
```
ga-devfront:
```suggestion
            $this->logger->error($this->translator->trans('Backup filename has not been set'));
```

ga-devfront (dismissed):
A small typo otherwise everything is good, thank you!

classes/Task/Backup/BackupFiles.php
M0rgan01: In fact the error refers to the name of the zip... We lose the sense of error.

Maybe "The filename for backup files has not been set." ? Or "No filename has been set for the backup files."

ga-devfront (dismissed):

classes/Log/Logger.php
```diff
@@ -175,9 +175,37 @@ public function updateLogsPath(string $logsPath): void
...
+            );
+        }
+    }
+
+    /**
+     * Get the name of the class that called the logger.
+     */
+    private function getCallingClass(): ?string
+    {
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
```
Quetzacoalt91: I loved the idea but I didn't know this method would be used, it is known to take a long time to return the stack. Can you please check the impact on the performance?

Another solution would be to set the class of the running task with a getter / setter.

M0rgan01:
@Quetzacoalt91 


# With classname (with dev branch) in seconde

1. 72.21 
2. 67.57
3. 68.83
4. 69.28
5. 68.72
6. 68.43
7. 69.23
8. 70.80
9. 68.67
10. 73.84

**Average: 69.76**

# With classname in seconde

1. 68.54
2. 71.46
3. 67.55
4. 68.26
5. 70.99
6. 74.65
7. 69.34
8. 71.15
9. 68.11
10. 69.54

**Average: 69.96**

_test conditions_: I use `time` command with the `update:start` command. 1.7.7.0 to 8.2.0 with https://github.com/PrestaShop/SeamlessUpgradeToolbox

Total logs for 1 update: 30 000
PHP version: 7.4

Quetzacoalt91:
I've created a task that only calls the logger in a reasonable way: 

```php
class Test extends AbstractTask
{
    const TASK_TYPE = TaskType::TASK_TYPE_UPDATE;

    /**
     * @throws Exception
     */
    public function run(): int
    {
        for ($i = 0 ; $i < 450 ; ++$i) {
            $this->logger->info($this->translator->trans(
                'Logger called'
            ));
        }

        $this->next = TaskName::TASK_UPDATE_COMPLETE;

        return ExitCode::SUCCESS;
    }
}
```

It takes 12 on the `dev` branch and 17 ms on this PR, which seems still acceptable.

When we profile it, getting the caller is taking ~13% of the script execution time and is flagged by Blackfire as an intensive node. 
https://blackfire.io/profiles/77d55665-82d3-436d-a69e-bfdd7c9a0229/graph
image

And when focusing on this method on particular:
Capture d’écran du 2025-01-07 11-46-47
I'll suggest a small improvement on the subclass check, and this part should be unit tested to be future-proof.

classes/Log/Logger.php
```diff
@@ -175,9 +175,37 @@ public function updateLogsPath(string $logsPath): void
...
+
+        foreach ($trace as $frame) {
+            if (
+                isset($frame['class']) &&
+                !is_subclass_of($frame['class'], self::class) &&
+                $frame['class'] !== self::class
+            ) {
```
Quetzacoalt91:
```suggestion
        // We don't need the whole stack. The 5th level is enough to get
        // the called of the logger but we keep a buffer in case it goes deeper.
        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 8);

        foreach ($trace as $frame) {
            if (
                isset($frame['class']) &&
                strpos($frame['class'], __NAMESPACE__) === false
            ) {
```

Quetzacoalt91 (approved):

ga-devfront (approved):

Quetzacoalt91:
[2025-01-08-111321-backup.txt](https://github.com/user-attachments/files/18345713/2025-01-08-111321-backup.txt)

[2025-01-08-110915-update.txt](https://github.com/user-attachments/files/18345690/2025-01-08-110915-update.txt)

Caller class is properly reported in the log file.

PR #1107, opened 2025-01-08
Introduce Error 404 page

Quetzacoalt91:
| Questions         | Answers
| Description?      | Implement the display of the page "Page Not Found"
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | When you attempt to load a non-existing page, the page 404 is displayed.


_dev/src/ts/api/RequestHandler.ts
```diff
@@ -32,16 +32,25 @@ export class RequestHandler {
...
+      if (error instanceof AxiosError) {
+        if (error.response?.data) {
+          const responseData = error.response.data;
+          responseData.new_route = 'error-page';
+          await this.#handleResponse(responseData, true);
+        }
+      } else {
+        // TODO: catch errors
+        console.error(error);
+      }
```
M0rgan01, line 53: Axios allows the use of interceptor, which allows specific behavior based on an error code. This would allow us to easily control/orient the view used and potentially have logic on a specific error.

Ex:

```
axiosInstance.interceptors.response.use(undefined, (error) => {
  const { response } = error;
  if (!response) {
    history.push(ERROR_UNKNOWN);
  } else {
    const { status } = response;

    switch (status) {
      case 409:
        break;
      case 404:
        history.push(ERROR_NOT_FOUND);
        break;
      case 401:
        history.push(ERROR_UNAUTHORIZED);
        break;
      case 403:
        history.push(ERROR_FORBIDDEN);
        break;
      case 500:
      case 503:
      case 400:
      default:
        history.push(ERROR_UNKNOWN);
    }
  }
  return Promise.reject(error);
});
```

_dev/src/ts/pages/ErrorPage.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+  }
+
+  public mount = () => {
+    this.errorPage?.mount();
+  };
+
+  public beforeDestroy = () => {
+    this.errorPage?.beforeDestroy();
+  };
```
M0rgan01:
```suggestion
  public mount = (): void => {
    this.errorPage?.mount();
  };

  public beforeDestroy = (): void => {
    this.errorPage?.beforeDestroy();
  };
```

_dev/src/ts/pages/error/ErrorPage404.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  public mount = () => {
+    this.#activeActionButton.classList.remove('hidden');
+    this.#form.addEventListener('submit', this.#onSubmit);
+  };
+
+  public beforeDestroy = () => {
+    this.#form.removeEventListener('submit', this.#onSubmit);
+  };
```
M0rgan01:
```suggestion
  public mount = (): void => {
    this.#activeActionButton.classList.remove('hidden');
    this.#form.addEventListener('submit', this.#onSubmit);
  };

  public beforeDestroy = (): void => {
    this.#form.removeEventListener('submit', this.#onSubmit);
  };
```

_dev/src/ts/pages/error/ErrorPage404.ts
```diff
@@ -0,0 +1,53 @@
...
+  get #exitButton(): HTMLAnchorElement {
+    const link = document.getElementById('exit-button');
+
+    if (!link || !(link instanceof HTMLAnchorElement)) {
+      throw new Error('Link is not found or invalid');
+    }
+    return link;
+  }
+
+  readonly #onSubmit = async (event: Event) => {
```
M0rgan01:
```suggestion
  readonly #onSubmit = async (event: Event): Promise<void> => {
```

classes/Router/UrlGenerator.php
```diff
@@ -0,0 +1,74 @@
...
+
+        $path = array_splice($path, 0, -$numberOfSubDirs);
+
+        return implode('/', $path) ?: '/';
+    }
+
+    public function getShopAdminAbsolutePathFromRequest(Request $request): string
+    {
+        return rtrim($this->getShopAbsolutePathFromRequest($request), '/') . '/' . $this->adminFolder;
+    }
```
M0rgan01, line 73: It might be worth it to also have one getViewsAbsolutePathFromRequest ?

```
public function getShopAdminAbsolutePathFromRequest(Request $request): string
    {
        returnrtrim($this->getShopAbsolutePathFromRequest($request), '/') . '/modules/autoupgrade/views';
    }
```

controllers/admin/self-managed/Error404Controller.php
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+namespace PrestaShop\Module\AutoUpgrade\Controller;
+
+use PrestaShop\Module\AutoUpgrade\Router\Routes;
+use Symfony\Component\HttpFoundation\Response;
+
+class Error404Controller extends AbstractPageController
+{
+    const ERROR_CODE = 404;
```
M0rgan01: we could use the constants already available in symfony (`Symfony\Component\HttpFoundation\Response::HTTP_NOT_FOUND`)

classes/Router/UrlGenerator.php
Quetzacoalt91: I left this responsibility to the class AssetsGenerator as it is its only job to concatenate this path to the production or the development base URL/Path.

_dev/src/ts/api/RequestHandler.ts
Quetzacoalt91: I like the idea of using the interceptor, especially to validate the response contents from the back end.

I suggest to handle this in another pull-request though, as we only cover the HTTP 404 here.

M0rgan01 (dismissed):

views/templates/layouts/page.html.twig
```diff
@@ -20,6 +20,6 @@
       </div>
     {% endblock %}
   </div>
-
-  <div id="ua_dialog"></div>
 {% endblock %}
+
+<div id="ua_dialog"></div>
```
tblivet, line 25: I think it would be good to wrap the dialog section in a `{% block %}` to improve flexibility and allow for easier customization.
```suggestion
{% block update_assistant_dialog %}
  <div id="ua_dialog"></div>
{% endblock %}
```

ga-devfront (dismissed):

M0rgan01 (dismissed):

ga-devfront (dismissed):

ga-devfront (dismissed):

Quetzacoalt91:
Had to rebase again

Here we go again

ga-devfront (dismissed):

AureRita (approved):
Hi @Quetzacoalt91 

[recording(45).webm](https://github.com/user-attachments/assets/835f084d-b93c-4b35-be8b-8befb6fd3f20)

and : 
image


Tested on : 
8.0.4 and 8.2 

Thank you

ps-jarvis:
PR merged, well done!

PR #1117, opened 2025-01-14
Update lifecycle of state and configuration files

Quetzacoalt91:
| Questions         | Answers
| Description?      | Make sure the state is only defined while a process is run (update / backup / restore) and rely on the configuration class for the rest of the module use.
| Type?             | Refacto
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | <ul><li>Make a process fail : The next attempt must start from the beginning not where it stopped.</li><li>Starting over the update funel will reset the backup completion, displaying again the step asking for the merchant to backup its store.</li></ul>

### DIFF
--- a/classes/Commands/CheckRequirementsCommand.php
+++ b/classes/Commands/CheckRequirementsCommand.php
@@ -32,7 +32,6 @@ use PrestaShop\Module\AutoUpgrade\Exceptions\DistributionApiException;
 use PrestaShop\Module\AutoUpgrade\Exceptions\UpgradeException;
 use PrestaShop\Module\AutoUpgrade\Parameters\UpgradeConfiguration;
 use PrestaShop\Module\AutoUpgrade\Task\ExitCode;
-use PrestaShop\Module\AutoUpgrade\UpgradeContainer;
 use PrestaShop\Module\AutoUpgrade\UpgradeSelfCheck;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
@@ -88,7 +87,6 @@ class CheckRequirementsCommand extends AbstractCommand
 
             $this->upgradeContainer->initPrestaShopAutoloader();
             $this->upgradeContainer->initPrestaShopCore();
-            $this->upgradeContainer->getUpdateState()->initDefault($this->upgradeContainer->getProperty(UpgradeContainer::PS_VERSION), $this->upgradeContainer->getUpgrader()->getDestinationVersion());
 
             $output->writeln('Result of prerequisite checks:');
             $this->exitCode = ExitCode::SUCCESS;
--- a/classes/UpgradeSelfCheck.php
+++ b/classes/UpgradeSelfCheck.php
@@ -85,6 +85,8 @@ class UpgradeSelfCheck
     private $autoUpgradePath;
     /** @var string */
     private $currentVersion;
+    /** @var string */
+    private $destinationVersion;
     /** @var PrestashopConfiguration */
     private $prestashopConfiguration;
     /** @var UpgradeConfiguration */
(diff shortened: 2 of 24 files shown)


classes/Upgrader.php
```diff
@@ -140,6 +140,9 @@ public function getDestinationVersion(): ?string
     {
         if ($this->updateConfiguration->isChannelLocal()) {
             return $this->updateConfiguration->getLocalChannelVersion();
+        } elseif ($this->updateConfiguration->getOnlineChannelVersion()) {
+            // Version of online channel has been cached? Return it.
+            return $this->updateConfiguration->getOnlineChannelVersion();
```
M0rgan01: I'm not convinced by this change for 2 reasons:

- This getter must be called once per proccess, to then be transmitted to the state if the information must be persisted for the process, so adding a cache is limited since the state already does this work
- If a new version is released, it will appear correctly if I went to the version page chosen a little while ago ? Just clicked on the choice between online and local, without continuing, saves the information

classes/Commands/CheckRequirementsCommand.php
```diff
@@ -88,7 +88,11 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
 
             $this->upgradeContainer->initPrestaShopAutoloader();
             $this->upgradeContainer->initPrestaShopCore();
-            $this->upgradeContainer->getUpdateState()->initDefault($this->upgradeContainer->getProperty(UpgradeContainer::PS_VERSION), $this->upgradeContainer->getUpgrader()->getDestinationVersion());
```
M0rgan01, line 91: We need the state for the check of requirements.

Quetzacoalt91: Can you clarify why ? The objective is to not have the state anymore outside the actual process.

M0rgan01: if we go to the UpgradeSelfCheck class and look for occurrences of the state, we can see that this one is used.

> The objective is to not have the state anymore outside the actual process

Can you define "process"? Isn't the requirements check a process in itself?

Quetzacoalt91: The process running the tasks is a part of the (update) funnel. By having the state outside the process applying changes on the store, we struggle to clean the previous context properly in case of error.

Thanks for the class, I'll have a look at it as well.

controllers/admin/AdminSelfUpgradeController.php
```diff
@@ -300,7 +300,8 @@ public function init()
         if (!$this->upgradeContainer->getUpdateState()->isInitialized()) {
```
M0rgan01, line 305: If we follow the logic introduced in the initialization tasks, this code should be removed

Quetzacoalt91: Yep, I added the comment to remove it as soon as the old UI is dropped.

classes/UpgradeSelfCheck.php
```diff
@@ -209,7 +208,7 @@ public function getWarnings(): array
      */
     public function getRequirementWording(int $requirement, bool $isWebVersion = false): array
     {
-        $version = $this->state->getDestinationVersion();
+        $version = $this->upgrader->getDestinationVersion();
```
M0rgan01: Given that we no longer use the state, and that calling this function of the upgrader is rather expensive, we could define passing it to constructor in the same way as the current version

M0rgan01 (approved):

paulnoelcholot:
Hello @Quetzacoalt91,

I tested your PR and it's good for me ! :tada: 


Thanks!

8.1.7 → 8.2
https://www.loom.com/share/1668d35efc9e4ae2aff15d4f1b00c989?sid=b3f0036a-49c8-47b0-a9fa-c34c1bd2bb62

image

paulnoelcholot (approved):

PR #1136, opened 2025-01-24
[NEW UI] add middleware system and first middleware on update options page

ga-devfront:
| Questions         | Answers
| Description?      | Add middleware system and the first middleware on update options page to test if configuration exist and if it's local version if xml and zip exists.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | see bellow

### Test Steps for Update Options Page Redirection Logic

1. **Navigate to the module homepage**.
2. **Test without a version choice**:  
   - Go to the update version choice page without picking any version.
   - Manually change the `route` parameter in the URL to `update-page-update-options` (replacing `update-page-version-choice`).
   - **Expected result**: You are redirected back to `update-page-version-choice`.

3. **Test with the online channel**:  
   - Select the online channel.
   - Change the `route` parameter in the URL to `update-page-update-options`.
   - **Expected result**: You can access `update-page-update-options`.

4. **Test with the local channel**:  
   - Return to `update-page-version-choice`.
   - Select the local channel.
   - Change the `route` parameter in the URL to `update-page-update-options`.
   - **Expected result**: You can access `update-page-update-options`.

5. **Test invalid file references**:  
   - Return to `update-page-version-choice`.
   - Without making any changes to the channel selection, delete the XML or ZIP file from your server that was previously picked.
   - Change the `route` parameter in the URL to `update-page-update-options`.
   - **Expected result**: You are redirected back to `update-page-version-choice`.

**PS: If you have already a configuration set you can delete it manually from you server by removing `/admin-folder/autoupgrade/update_config.varl`**


classes/Router/RoutesConfig.php
```diff
@@ -0,0 +1,199 @@
...
+                'controller' => UpdatePageVersionChoiceController::class,
+                'method' => 'themeTemperedFilesDialog',
+            ],
+            /* step: update options */
+            Routes::UPDATE_PAGE_UPDATE_OPTIONS => [
+                'controller' => UpdatePageUpdateOptionsController::class,
+                'method' => 'index',
+                'middleware' => [
+                    RoutesMiddleware::UPDATE_IS_CONFIGURED,
+                    RoutesMiddleware::LOCAL_CHANNEL_XML_AND_ZIP_EXIST,
```
Quetzacoalt91: Is having the whole definition of `RoutesMiddleware` necessary? Looks like the whole system would be simpler by having directly the classes here.

```suggestion
                    UpdateIsConfigured::class,
                    LocalChannelXmlAndZipExist::class,
```

classes/Router/RoutesMiddleware.php
```diff
@@ -0,0 +1,12 @@
...
+namespace PrestaShop\Module\AutoUpgrade\Router;
+
+use PrestaShop\Module\AutoUpgrade\Router\Middlewares\LocalChannelXmlAndZipExist;
+use PrestaShop\Module\AutoUpgrade\Router\Middlewares\UpdateIsConfigured;
+
+class RoutesMiddleware
+{
+    const UPDATE_IS_CONFIGURED = UpdateIsConfigured::class;
+    const LOCAL_CHANNEL_XML_AND_ZIP_EXIST = LocalChannelXmlAndZipExist::class;
+}
```
Quetzacoalt91: Following the other comment, I'm not sure about the added-value of this class.

classes/Router/Router.php
```diff
@@ -239,10 +51,38 @@ public function __construct(UpgradeContainer $upgradeContainer)
...
+     */
+    private function changeRouteRequestParam(Request $request, string $route): string
+    {
+        $baseUrl = $request->getSchemeAndHttpHost() . $request->getBaseUrl();
+        $queryParams = $request->query->all();
+
+        $queryParams['route'] = $route;
 
-        $method = $route['method'];
+        $queryString = http_build_query($queryParams);
```
Quetzacoalt91: To fix your issue about the unexpected `controllerUri` that appears in your URL after a redirect, I think we can rely on a part of the method `controllers/admin/self-managed/AbstractGlobalController.php` that could be extracted in a public method in UrlGenerator.php.

```php
        function getUrlToRoute($route: string): string
        {
            $params = [];
            parse_str($this->request->server->get('QUERY_STRING'), $params);
            $nextQueryParams = http_build_query(array_merge($params, ['route' => $destinationRoute]));

            return $this->request->getSchemeAndHttpHost() . $this->request->getBaseUrl() . $this->request->getPathInfo() . '?' . $nextQueryParams;
        }
```

You would get instead in this file:

```php
$newUrl = $this->upgradeContainer->getUrlGenerator()->getUrlToRoute($destination);
```

classes/Router/Router.php
```diff
@@ -239,10 +51,38 @@ public function __construct(UpgradeContainer $upgradeContainer)
...
 
-        return (new $route['controller']($this->upgradeContainer, $request))->$method();
+        return $baseUrl . '?' . $queryString;
```
Quetzacoalt91:
```suggestion
        header('Location: ' . $newUrl, true, 302);
        exit;
```

classes/Router/Router.php
```diff
@@ -239,10 +51,38 @@ public function __construct(UpgradeContainer $upgradeContainer)
...
+        return (new $routeParams['controller']($this->upgradeContainer, $request))->$method();
+    }
+
+    /**
+     * @param Request $request
+     * @param Routes::* $route
+     *
+     * @return string
+     */
+    private function changeRouteRequestParam(Request $request, string $route): string
```
Quetzacoalt91:
```suggestion
     */
    private function dirtyRedirectToRoute(Request $request, string $route): never
```

classes/Router/Router.php
```diff
@@ -239,10 +51,38 @@ public function __construct(UpgradeContainer $upgradeContainer)
...
-        $route = self::ROUTES[$routeName] ?? self::ROUTES[Routes::ERROR_404];
+        $route = array_key_exists($routeName, RoutesConfig::ROUTES) ? $routeName : Routes::ERROR_404;
+
+        $route = (new Middleware($this->upgradeContainer))->process($route);
+
+        if ($routeName !== $route) {
+            $newUrl = $this->changeRouteRequestParam($request, $route);
+
+            header('Location: ' . $newUrl, true, 302);
+            exit;
```
Quetzacoalt91:
```suggestion
            $this->dirtyRedirectToRoute($request, $route);
```

classes/Router/Router.php
```diff
@@ -239,10 +51,38 @@ public function __construct(UpgradeContainer $upgradeContainer)
     public function handle(Request $request)
     {
         $routeName = $request->query->get('route') ?? Routes::HOME_PAGE;
-        $route = self::ROUTES[$routeName] ?? self::ROUTES[Routes::ERROR_404];
+        $route = array_key_exists($routeName, RoutesConfig::ROUTES) ? $routeName : Routes::ERROR_404;
```
Quetzacoalt91: What's the reason behind the use of `array_key_exists`? It should be a bit slower than the previous code.

ga-devfront: because if I do `$route = RoutesConfig::ROUTES[$routeName] ? $routeName : Routes::ERROR_404;` PhPStan says ` Ternary operator condition is always true. `

Quetzacoalt91 (approved):

classes/Router/Middlewares/UpdateIsConfigured.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace PrestaShop\Module\AutoUpgrade\Router\Middlewares;
+
+use PrestaShop\Module\AutoUpgrade\Parameters\UpgradeFileNames;
+use PrestaShop\Module\AutoUpgrade\Router\Routes;
+
+class UpdateIsConfigured extends AbstractMiddleware
+{
+    public function process(): ?string
```
M0rgan01, line 10:
```suggestion
/**
     * @return Routes::*|null
     */
    public function process(): ?string
```

classes/Router/Middlewares/LocalChannelXmlAndZipExist.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace PrestaShop\Module\AutoUpgrade\Router\Middlewares;
+
+use PrestaShop\Module\AutoUpgrade\Router\Routes;
+
+class LocalChannelXmlAndZipExist extends AbstractMiddleware
```
M0rgan01, line 7: The validation function does more than just check the presence of the zip/xml pair, right?

classes/Router/Router.php
```diff
@@ -239,10 +51,35 @@ public function __construct(UpgradeContainer $upgradeContainer)
...
+
+    /**
+     * @param Request $request
+     * @param Routes::* $route
+     *
+     * @return never
+     */
+    private function dirtyRedirectToRoute(Request $request, string $route): void
+    {
+        $newUrl = $this->upgradeContainer->getUrlGenerator()->getUrlToRoute($request, $route, ['_redirected' => '1']);
```
M0rgan01, line 80: what a shame that we are not using symfony routes... we would have used the "redirectToRoute" functions directly, without having "dirty" code like that

classes/Router/UrlGenerator.php
```diff
@@ -71,4 +71,20 @@ public function getShopAdminAbsolutePathFromRequest(Request $request): string
...
+    /**
+     * @param Request $request
+     * @param string $destinationRoute
+     * @param array<string, mixed> $params
+     *
+     * @return string
+     */
+    public function getUrlToRoute(Request $request, string $destinationRoute, array $params = []): string
+    {
+        $queryStringParams = [];
```
M0rgan01, line 84: we can remove the declaration

classes/Router/UrlGenerator.php
```diff
@@ -71,4 +71,20 @@ public function getShopAdminAbsolutePathFromRequest(Request $request): string
...
+        parse_str($request->server->get('QUERY_STRING'), $queryStringParams);
+        $nextQueryParams = http_build_query(array_merge($queryStringParams, $params, ['route' => $destinationRoute]));
+
+        return $request->getSchemeAndHttpHost() . $request->getBaseUrl() . $request->getPathInfo() . '?' . $nextQueryParams;
```
M0rgan01, line 88: what do you think ?

```suggestion
    $currentQueryParams = $request->query->all();
    $nextQueryParams = array_merge($currentQueryParams, $params, ['route' => $destinationRoute]);
    return $request->getSchemeAndHttpHost()
        . $request->getBaseUrl()
        . $request->getPathInfo()
        . '?' . http_build_query($nextQueryParams);
```

AureRita (approved):
Hi @ga-devfront 

[recording(52).webm](https://github.com/user-attachments/assets/f8cf4296-0d78-4542-9b66-9aab23370c6f)

I also tested when you delete the file "update_config.var", and it works as expected ( You have to select again your version destination )

Tested from : 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

Thank you

classes/Router/Middlewares/LocalChannelXmlAndZipExist.php
ga-devfront: Yes and no, it check validty only if we are on channel local.

classes/Router/UrlGenerator.php
ga-devfront: no if we add it directly inside $params, old $params are crashed.

classes/Router/UrlGenerator.php
ga-devfront: we do `get('QUERY_STRING')` because we got some hidden query we don't want.
For nextParams we need to merge passed `$params`, the retriewed with `get('QUERY_STRING')`, and the new route !

classes/Router/Middlewares/LocalChannelXmlAndZipExist.php
M0rgan01: for me in addition to validating the presence of files, it also validates the presence of versions and the match of this one. So the middleware doesn't really reflect what is being done

classes/Router/UrlGenerator.php
M0rgan01: `parse_str($request->server->get('QUERY_STRING'), $queryStringParams);` and `$request->query->all()` do the same job here... in addition to gaining readability

classes/Router/Middlewares/LocalChannelXmlAndZipExist.php
ga-devfront: You think `LocalChannelXmlAndZipAreValid` is better ?

classes/Router/UrlGenerator.php
Quetzacoalt91: Their behavior is slightly different. $this->query can be modified at runtime while QUERY_STRING is a read only variable containing the original parameters sent to the current HTTP request.

PR #1147, opened 2025-01-30
UI Test - Remove the part previous part to go to the new UI

AureRita:
| Questions         | Answers
| Description?      | I remove the part where we specified that we go on new UI because currently it's not here anymore
| Type?             | bug fix
| BC breaks?        |  no
| Deprecations?     | yes 
| Fixed ticket?     | 
| Sponsor company   | PS
| How to test?      | CI is Green


tests/UI/campaigns/functional/01_upgrade.spec.ts
```diff
@@ -186,7 +186,7 @@ test.describe('Verify the New UI', () => {
       expect(isModuleVisible).toEqual(true);
     });
 
-    // Go to the old UI
+    // Go to the New UI
```
Quetzacoalt91:
```suggestion
```

No need to specify anymore if we are using the old or the new UI, the old one has been removed

Progi1984 (approved):

PR #1156, opened 2025-02-04
Refactoring of the admin/autoupgrade directory structure

M0rgan01:
| Questions         | Answers
| Description?      | see below
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | -

This PR reorganizes the admin/autoupgrade directory structure to improve temporary file management and prevent accidental overwriting of local releases.

### Changes introduced:

- The `download` folder no longer stores the downloaded release in online mode and is no longer erased in certain scenarios, preventing the loss of locally stored releases.
- The `latest` folder, which contained the extracted destination release, has been renamed to `files` and moved to `tmp/`.
- A new `tmp/releases` folder has been created to store the downloaded release in online mode.
- A new `tmp/modules` folder has been created to store module ZIP files and migration scripts (previously located at the root of `tmp/`).
- Module migration scripts are now deleted at the end of the update process.

These changes enhance the reliability of the update process and optimize the organization of temporary files.
#### Before:
```
admin/autoupgrade/
 ├── download/              (contained the downloaded release and local archives)
 ├── latest/                (contained extracted destination release)
 ├── tmp/                   (contained module zip files and migration scripts)
 ```
 
 
 #### After:
```
 admin/autoupgrade/
 ├── download/              (no longer stores downloaded releases, is now only present for local archives)
 ├── tmp/
 │   ├── files/             (previously "latest", deleted at the end of the update process)
 │   ├── releases/          (stores downloaded releases in online mode, emptied at the end of the update process)
 │   ├── modules/           (stores module zip files and migration scripts, emptied at the end of the update process)
 ```

### DIFF
--- a/classes/Task/Update/Unzip.php
+++ b/classes/Task/Update/Unzip.php
@@ -57,7 +57,7 @@ class Unzip extends AbstractTask
         );
 
         if ($this->container->getFilesystemAdapter()->clearDirectory($destExtract)) {
-            $this->logger->debug($this->translator->trans('files directory has been emptied'));
+            $this->logger->debug($this->translator->trans('Temporary files directory has been emptied'));
         }
         $relative_extract_path = str_replace(_PS_ROOT_DIR_, '', $destExtract);
         $report = '';
--- a/classes/Task/Update/Unzip.php
+++ b/classes/Task/Update/Unzip.php
@@ -49,22 +49,15 @@ class Unzip extends AbstractTask
      */
     public function run(): int
     {
-        $filepath = $this->container->getFilePath();
-        $destExtract = $this->container->getProperty(UpgradeContainer::LATEST_PATH);
+        $filepath = $this->container->getArchiveFilePath();
+        $destExtract = $this->container->getProperty(UpgradeContainer::TMP_FILES_PATH);
 
         $this->container->getUpdateState()->setProgressPercentage(
             $this->container->getCompletionCalculator()->getBasePercentageOfTask(self::class)
         );
 
-        if ($this->container->getFileSystem()->exists($destExtract)) {
-            foreach (scandir($destExtract) as $item) {
-                if ($item !== '.' && $item !== '..') {
-                    $path = $destExtract . DIRECTORY_SEPARATOR . $item;
-                    $this->container->getFileSystem()->remove($path);
-                }
-            }
-
-            $this->logger->debug($this->translator->trans('"/latest" directory has been emptied'));
+        if ($this->container->getFilesystemAdapter()->clearDirectory($destExtract)) {
+            $this->logger->debug($this->translator->trans('files directory has been emptied'));
         }
         $relative_extract_path = str_replace(_PS_ROOT_DIR_, '', $destExtract);
         $report = '';
--- a/classes/Task/Update/UpdateComplete.php
+++ b/classes/Task/Update/UpdateComplete.php
@@ -58,8 +58,8 @@ class UpdateComplete extends AbstractTask
         $this->next = TaskName::TASK_COMPLETE;
 
         $filesystem = $this->container->getFileSystem();
-        $filePath = $this->container->getFilePath();
-        $latestPath = $this->container->getProperty(UpgradeContainer::LATEST_PATH);
+        $filePath = $this->container->getArchiveFilePath();
+        $latestPath = $this->container->getProperty(UpgradeContainer::TMP_FILES_PATH);
 
         if ($filesystem->exists($filePath)) {
             if ($this->container->getUpdateConfiguration()->isChannelOnline()) {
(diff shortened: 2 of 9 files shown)


Hlavtox:
@M0rgan01 Would it be possible to move these files into the folder of the module, or a var directory?

Why I ask - we run integrity checks on our shops every night and we have to whitelist these files as they could change depending on if autoupgrade is installed, I think that the admin directory shouldn't really change.

M0rgan01:
@Hlavtox We actually asked ourselves this question. We use this folder because we are forced to do so for several technical reasons, but also to remain compatible with old versions (not to force users to move their backup / local archives) but also to be able to consult the logs by a URL hidden behind the admin folder. 

For v7 it will stay like this but it is not excluded that it will change in the future

classes/UpgradeContainer.php
```diff
@@ -84,16 +84,22 @@
...
     const BACKUP_PATH = 'backup';
     const DOWNLOAD_PATH = 'download';
-    const LATEST_PATH = 'latest'; // AdminSelfUpgrade::$latestRootDir
-    const LATEST_DIR = 'latest/';
     const LOGS_PATH = 'logs';
+
+    // temporary folders
     const TMP_PATH = 'tmp';
+    const FILES_PATH = 'files';
+    const FILES_DIR = 'files/';
```
Quetzacoalt91: Shouldn't these constants be called 

```suggestion
    const TMP_FILES_PATH = 'files';
    const TMP_FILES_DIR = 'files/';
```

like the 3 next ones? As they are all subdirectories of the `tmp/` folder.

classes/UpgradeContainer.php
```diff
@@ -275,25 +281,45 @@ public function getProperty(string $property): ?string
...
-            case self::LATEST_DIR:
-                return $this->autoupgradeWorkDir . DIRECTORY_SEPARATOR . 'latest' . DIRECTORY_SEPARATOR;
             case self::TMP_PATH:
                 return $this->autoupgradeWorkDir . DIRECTORY_SEPARATOR . 'tmp';
+            case self::TMP_MODULES_PATH:
+                return $this->getProperty(self::TMP_PATH) . DIRECTORY_SEPARATOR . 'modules';
+            case self::TMP_MODULES_DIR:
+                return $this->getProperty(self::TMP_MODULES_PATH) . DIRECTORY_SEPARATOR;
+            case self::TMP_RELEASES_DIR:
+                return $this->getProperty(self::TMP_PATH) . DIRECTORY_SEPARATOR . 'releases' . DIRECTORY_SEPARATOR;
```
Quetzacoalt91: Shall we keep the consistency of having two properties for each folder?

classes/Task/Update/UpdateComplete.php
```diff
@@ -79,6 +79,17 @@ public function run(): int
         }
 
         // removing temporary files
+        $tmpModulePath = $this->container->getProperty(UpgradeContainer::TMP_MODULES_PATH);
+        if ($this->container->getFileSystem()->exists($tmpModulePath)) {
+            foreach (scandir($tmpModulePath) as $item) {
+                if ($item !== '.' && $item !== '..' && $item !== 'index.php') {
```
Quetzacoalt91: I've seen these checks on another place, I wonder is a helper method for this is a good addition to the project.

ga-devfront (commented):
if the download folder is only used for local archives why not rename it local archives?

M0rgan01:
> if the download folder is only used for local archives why not rename it local archives? 

For a matter of practicality and history. This folder is already present when unzipping a store, even if the module is not installed. Moreover, this prevents users from having to move or lose their archives.

classes/UpgradeContainer.php
M0rgan01: For me, it's a matter of need; it's not necessary for every folder to always have a constant path and another directory.

Quetzacoalt91 (dismissed):

Quetzacoalt91 (dismissed):

Quetzacoalt91 (dismissed):

classes/Task/Update/Unzip.php
```diff
@@ -49,22 +49,15 @@ public function init(): void
...
-            foreach (scandir($destExtract) as $item) {
-                if ($item !== '.' && $item !== '..') {
-                    $path = $destExtract . DIRECTORY_SEPARATOR . $item;
-                    $this->container->getFileSystem()->remove($path);
-                }
-            }
-
-            $this->logger->debug($this->translator->trans('"/latest" directory has been emptied'));
+        if ($this->container->getFilesystemAdapter()->clearDirectory($destExtract)) {
+            $this->logger->debug($this->translator->trans('files directory has been emptied'));
```
Quetzacoalt91:
```suggestion
            $this->logger->debug($this->translator->trans('Temporary files directory has been emptied'));
```

Quetzacoalt91 (dismissed):

classes/UpgradeTools/CacheCleaner.php
```diff
@@ -75,7 +75,7 @@ public function cleanFolders(): void
                 if ($file[0] === '.' || $file === 'index.php') {
                     continue;
                 }
-                $this->container->getFilesystemAdapter()->clearDirectory($dir . $file, true);
+                $this->container->getFilesystemAdapter()->clearDirectory($dir . $file);
                 $this->logger->debug($this->container->getTranslator()->trans('File %s removed', [$file]));
```
Quetzacoalt91: If you wanted to backport the changes from https://github.com/PrestaShop/autoupgrade/pull/1180, does it lack a part?

Quetzacoalt91: This change probably conflicts with the recently merged #1180

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

[recording(80).webm](https://github.com/user-attachments/assets/bf6d1303-fdbd-4789-afb1-a71c10d63c71)

Tested from : 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

It seems to fix also https://github.com/PrestaShop/PrestaShop/issues/37921

Thank you

PR #1159, opened 2025-02-05
Use clearDirectory for cache cleaner

M0rgan01:
| Questions         | Answers
| Description?      | Use clearDirectory for cache cleaner
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | https://github.com/PrestaShop/PrestaShop/issues/37979
| Sponsor company   | -
| How to test?      | https://github.com/PrestaShop/PrestaShop/issues/37979

### DIFF
--- a/classes/UpgradeTools/FilesystemAdapter.php
+++ b/classes/UpgradeTools/FilesystemAdapter.php
@@ -25,27 +25,24 @@ use FilesystemIterator;
 use RecursiveCallbackFilterIterator;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
 
 class FilesystemAdapter
 {
-    /**
-     * @var FileFilter
-     */
+    /** @var Filesystem */
+    private $filesystem;
+
+    /** @var FileFilter */
     private $fileFilter;
 
-    /**
-     * @var string
-     */
+    /** @var string */
     private $autoupgradeDir;
 
-    /**
-     * @var string
-     */
+    /** @var string */
     private $adminSubDir;
 
-    /**
-     * @var string
-     */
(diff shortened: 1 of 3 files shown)


Quetzacoalt91:
Shall we wait for https://github.com/PrestaShop/autoupgrade/pull/1159 to be merged first?

ga-devfront (dismissed):

classes/UpgradeTools/FilesystemAdapter.php
```diff
@@ -230,4 +228,27 @@ public function isReleaseValid(string $path): bool
...
+                }
+            }
+
+            if ($deleteFolder) {
+                $this->filesystem->remove($folderToClear);
+                $hasDeletedItems = true;
+            }
+        }
+
+        return $hasDeletedItems;
```
Hlavtox, line 263: @M0rgan01 If there is nothing to delete, shouldnt we also return true? Maybe sometimes you will call it by accident on an empty folder and wonder why it returns false when the folder is empty.

M0rgan01: @Hlavtox 

Thank you for your feedback. 

This boolean is present to indicate if elements have been deleted, for the purpose of conditioning potential logs, we could totally do without it.

To know if the process is going well, you should use a try catch.

I will add a more detailed doc.

Hlavtox: @M0rgan01 No worries, just a thought. :-) I understand.

nicosomb (approved):

AureRita (approved):
Hi @M0rgan01 

[recording(69).webm](https://github.com/user-attachments/assets/3b2f7c05-b436-4a1f-b08e-0de1c6cc5e6b)


Tested from : 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

Thank you

PR #1160, opened 2025-02-05
Do not initialize in tasks the Core by default, only when requested

Quetzacoalt91:
| Questions         | Answers
| Description?      | We had a weird implementation of a default initialization of the core, that would suppressed by each task if necessary. We should actually rely on the core ONLY when necessary, so only the task needing it to work will call the method `initPrestaShopCore()`. This PR also moves a XML initialization to a specific place instead of at the beginning of each update task.
| Type?             | bug fix + improvement
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| Sponsor company   | /
| How to test?      | Backup, Updates and Restore are still possible, especially when using the local archive.

### DIFF
--- a/classes/Task/AbstractTask.php
+++ b/classes/Task/AbstractTask.php
@@ -180,26 +180,8 @@ abstract class AbstractTask
         }
     }
 
-    /**
-     * @throws Exception
-     */
     public function init(): void
     {
-        $this->container->initPrestaShopCore();
-        $this->setupEnvironment();
-    }
-
-    /**
-     * @throws Exception
-     */
-    protected function setupEnvironment(): void
-    {
-        $updateConfiguration = $this->container->getUpdateConfiguration();
-
-        if ($this::TASK_TYPE === TaskType::TASK_TYPE_UPDATE && $updateConfiguration->isChannelLocal()) {
-            $archiveXml = $updateConfiguration->getLocalChannelXml();
-            $this->container->getFileLoader()->addXmlMd5File($this->container->getUpgrader()->getDestinationVersion(), $this->container->getProperty(UpgradeContainer::DOWNLOAD_PATH) . DIRECTORY_SEPARATOR . $archiveXml);
-        }
     }
 
     abstract public function run(): int;
--- a/classes/Task/Update/UpdateFiles.php
+++ b/classes/Task/Update/UpdateFiles.php
@@ -196,6 +196,12 @@ class UpdateFiles extends AbstractTask
             $this->container->getCompletionCalculator()->getBasePercentageOfTask(self::class)
         );
 
+        $updateConfiguration = $this->container->getUpdateConfiguration();
+        if ($updateConfiguration->isChannelLocal()) {
+            $archiveXml = $updateConfiguration->getLocalChannelXml();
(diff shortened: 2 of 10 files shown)


ga-devfront (dismissed):

classes/Task/AbstractTask.php
```diff
@@ -180,27 +180,7 @@ public function setErrorFlag(): void
...
-    protected function setupEnvironment(): void
-    {
-        $updateConfiguration = $this->container->getUpdateConfiguration();
-
-        if ($this::TASK_TYPE === TaskType::TASK_TYPE_UPDATE && $updateConfiguration->isChannelLocal()) {
-            $archiveXml = $updateConfiguration->getLocalChannelXml();
-            $this->container->getFileLoader()->addXmlMd5File($this->container->getUpgrader()->getDestinationVersion(), $this->container->getProperty(UpgradeContainer::DOWNLOAD_PATH) . DIRECTORY_SEPARATOR . $archiveXml);
-        }
-    }
+    public function init(): void {}
```
ga-devfront: I think you can change this to 
```suggestion
   abstract public function init(): void;
```

Quetzacoalt91: By declaring the method as abstract, I will need to define it on each task even if I have nothing to call. This is not the behavior I wanted.

classes/Task/Update/UpdateFiles.php
```diff
@@ -196,6 +196,12 @@ protected function warmUp(): int
             $this->container->getCompletionCalculator()->getBasePercentageOfTask(self::class)
         );
 
+        $archiveXml = $this->container->getUpdateConfiguration()->getLocalChannelXml();
+        $this->container->getFileLoader()->addXmlMd5File(
+            $this->container->getUpgrader()->getDestinationVersion(),
+            $this->container->getProperty(UpgradeContainer::DOWNLOAD_PATH) . DIRECTORY_SEPARATOR . $archiveXml
+        );
+
```
M0rgan01: this is not only necessary for the local channel

```suggestion


         if ($this->container->getUpdateConfiguration()->isChannelLocal()) {
            $archiveXml = $this->container->getUpdateConfiguration()->getLocalChannelXml();
            $this->container->getFileLoader()->addXmlMd5File($this->container->getUpgrader()->getDestinationVersion(), $this->container->getProperty(UpgradeContainer::DOWNLOAD_PATH) . DIRECTORY_SEPARATOR . $archiveXml);
        }

```

Quetzacoalt91: Forgot to commit it, indeed

M0rgan01 (approved):

paulnoelcholot:
Hello @Quetzacoalt91,

I tested your PR and it's good for me!

Thanks!

804 tp 900 : 
https://github.com/user-attachments/assets/32967701-714b-4e59-9484-48e1471f9d73

1788 to 804 : 
https://github.com/user-attachments/assets/6b117c45-e21d-4c57-8058-800a7c0c713c

paulnoelcholot (approved):

ps-jarvis:
PR merged, well done!

PR #1162, opened 2025-02-06
[FIX] stepper to check init stepper again if it change.

ga-devfront:
| Questions         | Answers
| Description?      | If you come from update page to restore page with hydration, the stepper doesn't work anymore.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | You have to make and update with fail, click on restore after fail. If you the step on stepper change on restore the fix worked.

### DIFF
--- a/_dev/src/ts/utils/Stepper.ts
+++ b/_dev/src/ts/utils/Stepper.ts
@@ -19,36 +19,38 @@
 import type { Step } from '../types/Stepper';
 
 export default class Stepper {
-  private stepper: HTMLDivElement;
-  private steps: Step[];
+  #stepper: HTMLDivElement | null = null;
+  #steps: Step[] = [];
 
-  private baseClass = 'stepper__step';
-  private currentClass = `${this.baseClass}--current`;
-  private doneClass = `${this.baseClass}--done`;
-  private normalClass = `${this.baseClass}--normal`;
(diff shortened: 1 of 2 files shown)


_dev/src/ts/utils/Stepper.ts
```diff
@@ -60,43 +62,52 @@ export default class Stepper {
...
+    const stepIndex = this.#getStepIndex(currentStep);
 
     if (stepIndex === -1) {
       console.debug(`Step ${currentStep} not found in list.`);
       return;
     }
-    this.stepper.classList.add('stepper--hydration');
 
-    this.steps.forEach((step, i) => {
+    this.#stepper?.classList.add('stepper--hydration');
```
tblivet: Maybe we can define `stepper--hydration` has `#stepperHydrationClass = 'stepper--hydration';` ?

tblivet (approved):

AureRita (approved):
Hi @ga-devfront 

[recording(66).webm](https://github.com/user-attachments/assets/87608712-9d44-4add-ae0e-0ee6b107229e)


Tested from : 
8.0.4 to 8.2 
8.2 to 9.0.0
8.0.4 to 9.0.0

Thank you

PR #1163, opened 2025-02-06
[FIX] use state version if initialized instead using core files

ga-devfront:
| Questions         | Answers
| Description?      | Use state version if initialized instead using core files
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | You need to try and update from 1.7.8.4 to 9.0.0. You need to crash the update file in middle of process to reproduce ce bug. With the fix you won't have any more bug.


Quetzacoalt91 (approved):

Quetzacoalt91:
@paulnoelcholot I had the opportunity to test this fix as I had the issue this morning. All went well on my side.

ps-jarvis:
PR merged, well done!

PR #1168, opened 2025-02-11
Add new hooks in sql upgrade statement

tleon:
| Questions         | Answers
| Description?      | Added the new hooks from 9.0.x to the sql upgrade statement
| BC breaks?        |  no
| Deprecations?     |  no
| Fixed ticket?     | Fixes #37163
| How to test?      | CI green


kpodemski (commented):
@tleon I think you should update the part that @jolelievre added in his previous PR, because, for example, the part deleting the hooks is identical

upgrade/sql/9.0.0.sql
```diff
@@ -323,3 +323,106 @@ ALTER TABLE `PREFIX_attachment_lang` MODIFY COLUMN `name` varchar(255) DEFAULT N
...
+  (NULL, 'actionOnImageCutAfter', '', '', '1'),
+  (NULL, 'actionShopDataDuplication', '', '', '1'),
+  (NULL, 'termsAndConditions', '', '', '1'),
+  (NULL, 'actionValidateStepComplete', '', '', '1'),
+  (NULL, 'dashboardZoneOne', '', '', '1'),
+  (NULL, 'dashboardZoneTwo', '', '', '1'),
+  (NULL, 'dashboardData', '', '', '1'),
+  (NULL, 'actionPasswordRenew', '', '', '1'),
+  (NULL, 'actionDownloadAttachment', '', '', '1'),
+  (NULL, 'fooHook', '', '', '1'),
```
M0rgan01: Is there really a hook with this name in production?

M0rgan01:
How will users be able to use hooks without them having a description ?

upgrade/sql/9.0.0.sql
M0rgan01: https://github.com/PrestaShop/PrestaShop/blob/9.0.x/tests/Unit/Core/Hook/HookDispatcherTest.php#L80

It's a test hook

kpodemski: yes, this is the same problem as with the generated docs, some of the hooks should be excluded, I added more comments [on this PR ](https://github.com/PrestaShop/docs/pull/1902)

tleon: In the end i don't know anymore if this command is a good idea i'm a but tired of all of this :/

M0rgan01: @tleon I understand you

kpodemski: @tleon 

it is a good idea and it works great, the most important remaining thing is to have block-list of hooks available, so you don't generate them, hooks from `tests` directory, and hooks that are deprecated :)

jolelievre: Instead of a s list of blocked hooks we could have instead a list of folders that the scanner shouldn't scan

upgrade/sql/9.0.0.sql
```diff
@@ -323,3 +323,106 @@ ALTER TABLE `PREFIX_attachment_lang` MODIFY COLUMN `name` varchar(255) DEFAULT N
 /* Fix category thumbnail images */
 /* https://github.com/PrestaShop/PrestaShop/pull/36877 */
 /* PHP:ps_900_migrate_category_images(); */;
+
+/* Auto generated hooks added for version 9.0.0 */
+INSERT INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `position`) VALUES
```
jolelievre: Be careful the list of inserted hooks is already upper in this file The command is able to generate the SQL query but it cannot modify this file "in place", it only appends the content at the end So we need to manually keep the file consistent (either move this added code upper in the file or remove the upper part)

jolelievre (changes requested):

jolelievre (approved):

M0rgan01:
image

New hooks are correctly insert.

PR #1175, opened 2025-02-19
Introduce Twig function asset()

Quetzacoalt91:
| Questions         | Answers
| Description?      | Align the project with a function `asset()` available on Twig when used in a Symfony environment. This replaces the current implementation where we always have to think about sending the base path from all controller. 
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Images are still displayed

### DIFF
--- a/classes/UpgradeContainer.php
+++ b/classes/UpgradeContainer.php
@@ -48,6 +48,8 @@ use PrestaShop\Module\AutoUpgrade\State\LogsState;
 use PrestaShop\Module\AutoUpgrade\State\RestoreState;
 use PrestaShop\Module\AutoUpgrade\State\UpdateState;
 use PrestaShop\Module\AutoUpgrade\Task\TaskType;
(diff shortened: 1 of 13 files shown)


classes/UpgradeContainer.php
```diff
@@ -664,13 +670,15 @@ public function getTwig()
                 $loader->addPath(realpath(__DIR__ . '/..') . '/views/templates', 'ModuleAutoUpgrade');
                 $twig = new Environment($loader);
                 $twig->addExtension(new TransFilterExtension3($this->getTranslator()));
+                $twig->addExtension(new AssetFunctionExtension3($this));
```
Quetzacoalt91, line 673: Having the container as a parameter triggers the autowiring on the storybook side. Because of its `string` parameters, it cannot complete and makes the startup fail.

PR #1187, opened 2025-02-21
Handle symbolic links when updating the files

jolelievre:
| Questions         | Answers
| Description?      | Handle symbolic links when updating the files
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | ~
| Sponsor company   | ~
| How to test?      | This new feature is only testable and useful starting from prestashop 9.0.0 because it's the only one so far that contains a symbolic link You need to test the upgrade based on the 9.0.0 classic beta version https://github.com/PrestaShopCorp/prestashop-classic/releases/tag/9.0.0-beta.1-classic When you upgrade towards this version the file in `admin-dev/bundles/apiplatform` should be a symbolic link targetting `../../vendor/api-platform/core/src/Symfony/Bundle/Resources/public/`

### DIFF
--- a/classes/Progress/Backlog.php
+++ b/classes/Progress/Backlog.php
@@ -38,39 +38,30 @@ class Backlog
     private $backlog;
 
     /**
-     * @var array<string, string> Indexed by relative path from root folder, value is the symbolic link target
-     */
-    private $symbolicLinks;
-
-    /**
      * @param mixed[] $backlog
-     * @param int $initialTotal
-     * @param array<string, string> $symbolicLinks
      */
-    public function __construct(array $backlog, int $initialTotal, array $symbolicLinks = [])
+    public function __construct(array $backlog, int $initialTotal)
     {
         $this->backlog = $backlog;
         $this->initialTotal = $initialTotal;
-        $this->symbolicLinks = $symbolicLinks;
     }
 
     /**
-     * @param array{'backlog':mixed[],'initialTotal':int, 'symbolicLinks':array<string, string>} $contents
+     * @param array{'backlog':mixed[],'initialTotal':int} $contents
      */
     public static function fromContents($contents): self
     {
-        return new self($contents['backlog'], $contents['initialTotal'], $contents['symbolicLinks']);
+        return new self($contents['backlog'], $contents['initialTotal']);
     }
 
     /**
-     * @return array{'backlog':mixed[],'initialTotal':int, 'symbolicLinks':array<string, string>}
+     * @return array{'backlog':mixed[],'initialTotal':int}
      */
     public function dump(): array
     {
         return [
             'backlog' => $this->backlog,
             'initialTotal' => $this->initialTotal,
-            'symbolicLinks' => $this->symbolicLinks,
         ];
     }
 
@@ -90,13 +81,5 @@ class Backlog
     public function getInitialTotal(): int
     {
         return $this->initialTotal;
-    }
-
-    /**
-     * @return array<string, string>
-     */
-    public function getSymbolicLinks(): array
-    {
-        return $this->symbolicLinks;
     }
 }
--- a/classes/Progress/Backlog.php
+++ b/classes/Progress/Backlog.php
@@ -38,30 +38,39 @@ class Backlog
     private $backlog;
 
     /**
-     * @param mixed[] $backlog
+     * @var array<string, string> Indexed by relative path from root folder, value is the symbolic link target
      */
-    public function __construct(array $backlog, int $initialTotal)
+    private $symbolicLinks;
+
+    /**
+     * @param mixed[] $backlog
+     * @param int $initialTotal
+     * @param array<string, string> $symbolicLinks
+     */
+    public function __construct(array $backlog, int $initialTotal, array $symbolicLinks = [])
     {
         $this->backlog = $backlog;
         $this->initialTotal = $initialTotal;
+        $this->symbolicLinks = $symbolicLinks;
     }
 
     /**
-     * @param array{'backlog':mixed[],'initialTotal':int} $contents
+     * @param array{'backlog':mixed[],'initialTotal':int, 'symbolicLinks':array<string, string>} $contents
      */
     public static function fromContents($contents): self
     {
-        return new self($contents['backlog'], $contents['initialTotal']);
+        return new self($contents['backlog'], $contents['initialTotal'], $contents['symbolicLinks']);
     }
 
     /**
-     * @return array{'backlog':mixed[],'initialTotal':int}
+     * @return array{'backlog':mixed[],'initialTotal':int, 'symbolicLinks':array<string, string>}
      */
     public function dump(): array
     {
         return [
             'backlog' => $this->backlog,
             'initialTotal' => $this->initialTotal,
+            'symbolicLinks' => $this->symbolicLinks,
         ];
     }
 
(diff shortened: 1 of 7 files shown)


classes/Progress/Backlog.php
```diff
@@ -37,31 +37,40 @@ class Backlog
...
+     */
+    private $symbolicLinks;
+
     /**
      * @param mixed[] $backlog
+     * @param int $initialTotal
+     * @param array<string, string> $symbolicLinks
      */
-    public function __construct(array $backlog, int $initialTotal)
+    public function __construct(array $backlog, int $initialTotal, array $symbolicLinks = [])
```
Quetzacoalt91: `Backlog` class is something used in many tasks, this parameters seems very specific to one of them and would look for another solution.

A better solution can be storing the details of symbolic links inside the `$backlog` variable, an additional file as suggested by @M0rgan01, or another Backlog dedicated in enabling symbolic links.

classes/Task/Update/UpdateFiles.php
```diff
@@ -118,7 +115,19 @@ public function upgradeThisFile($orig): bool
...
-        if (is_dir($orig)) {
+
+        // Special case for symbolic links
+        if (!empty($symbolicLinkTarget)) {
+            // Remove the file/folder in case it already exists
+            if ($this->container->getFileSystem()->exists($dest)) {
+                $this->container->getFileSystem()->remove($dest);
+            }
+
+            symlink($symbolicLinkTarget, $dest);
```
M0rgan01: Symfony filesystem has a symlink method, i think we can use it for better exception handling

classes/Progress/Backlog.php
```diff
@@ -37,31 +37,40 @@ class Backlog
      */
     private $backlog;
 
+    /**
+     * @var array<string, string> Indexed by relative path from root folder, value is the symbolic link target
+     */
+    private $symbolicLinks;
+
```
M0rgan01: I'll let @Quetzacoalt91 tell me his opinion as well, but the Backlog class is supposed to be generic and usable for all kinds of scenarios. 

Adding a `symbolicLinks` variable directs us towards more precise use of processes.

 Maybe this could be integrated into the backlog which can contain a multitude of elements, and add a file type to differentiate symbolic links from others ?

classes/Task/Update/UpdateFiles.php
```diff
@@ -97,17 +99,12 @@ public function run(): int
      * upgradeThisFile.
      *
      * @param mixed $orig The absolute path to the file from the upgrade archive
+     * @param string $file Relative file path from root
      *
      * @throws Exception
      */
-    public function upgradeThisFile($orig): bool
+    protected function upgradeThisFile($orig, string $file, ?string $symbolicLinkTarget): bool
```
Quetzacoalt91: I'm wondering if we should have a dedicated method for symbolic links, because I see them being handled as special cases. This one would remain for the current management of basic files and folders.

classes/Progress/Backlog.php
Quetzacoalt91: I would suggest another Backlog for the symbolic links, the task taking care of them after having handled all the basic files.

classes/Progress/Backlog.php
jolelievre: So if I follow you two instead of having additional data in this `Backlog` class I should use another instance of `Backlog` that contains the list of symbolic links? I see the only idea, the only problem is that this class is thought like a pile that returns the next item like an iterator

But in the step the files Backlog is the one that really drives the loop, and in the middle of it the symbolic links Backlog would be used only from time to time. So to handle this I guess the `Backlog` class would need to have two more methods:
- `getAll` to get the list of symbolic links and check if the current file is one of them
- `remove` to remove the symbolic link when it's been handled

What do you think about this @M0rgan01 @Quetzacoalt91 ?

Or maybe since the usage is completely different from the usual use of `Backlog` this class is not adapted in this case and I should use another dumpable DTO?

classes/Xml/ChecksumCompare.php
```diff
@@ -144,6 +144,18 @@ public function getTamperedFilesOnShop(string $version)
...
+     * @param string $version
+     *
+     * @return string[] Indexed by relative path from root folder, value is the symbolic link target
+     */
+    public function getSymbolicLinks(string $version): array
+    {
+        $checksum = $this->fileLoader->getXmlMd5File($version);
+
+        return $this->symbolicLinksAsArray($checksum->ps_root_dir[0]);
+    }
```
Quetzacoalt91, line 157: Before merge, it would be nice to have tests on this method.

classes/Progress/Backlog.php
Quetzacoalt91: I think we could have two backlogs, because the actions we run for each of them are different.

I don't think these methods are necessary. Maybe I miss something but from my understanding, the current Backlog won't get the symbolic links because these files do not exist in the PrestaShop zip file, used to create the list of changed files.
If they don't exist by default, then we can have a whole new variable containing the symbolic links that won't conflict with the existing backlog?

jolelievre (commented):
I can't approve my own PR but it seems ok to me

classes/Task/Update/UpdateFiles.php
```diff
@@ -268,13 +322,13 @@ protected function warmUp(): int
             UpgradeFileNames::FILES_TO_UPGRADE_LIST
         );
 
-        if ($total_files_to_upgrade === 0) {
+        if ($total_files_to_upgrade === 0 && $totalSymbolicLinks === 0) {
```
M0rgan01: The two variables have the same purpose but are formatted differently. One in snakeCase and one in camelCase. Perhaps we could unify this?

classes/Xml/ChecksumCompare.php
```diff
@@ -259,6 +271,27 @@ protected function md5FileAsArray(SimpleXMLElement $node, string $dir = '/')
...
+        $links = [];
+        foreach ($node as $child) {
+            if (is_object($child) && $child->getName() == 'dir') {
+                $subDir = $dir ? $dir . DIRECTORY_SEPARATOR : '';
+                $subDir .= (string) $child['name'];
+                $subDirLinks = $this->symbolicLinksAsArray($child, $subDir);
+                $links = $links + $subDirLinks;
+            } elseif (is_object($child) && $child->getName() == 'link') {
+                $linkName = (string) $child['name'];
+                $links[] = ['name' => $dir . DIRECTORY_SEPARATOR . $linkName, 'target' => (string) $child];
```
M0rgan01, line 288: 'target' => (string) $child

child is an object here ?

Quetzacoalt91: This is indeed an instance of `SimpleXMLElement`, because a XML node can have children and all the tree is shown as `SimpleXMLElement`s

classes/Task/Update/UpdateFiles.php
Quetzacoalt91: This helps the review to avoid mixing improvements and refacto and other stuff in the same pull-request, but I don't see any issue to do it now in a separate commit.

M0rgan01 (approved):

ga-devfront (approved):
image
Approved by Dave !

Quetzacoalt91:
Thank you @jolelievre

PR #1188, opened 2025-02-21
[FIX] save iso for translations

### LINKED ISSUE #38110: autoupgrade translations not working

I have no idea why, but the translations seem to be broken in some parts of the module.

**They work on the initial page**

Image

**But not on the second one**

Image

Ping @Quetzacoalt91 @M0rgan01 

9.0.0

8.1

autoupgrade v7.0.0-dev.4

ga-devfront:
| Questions         | Answers
| Description?      | Translations fix https://github.com/PrestaShop/PrestaShop/issues/38110
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fix https://github.com/PrestaShop/PrestaShop/issues/38110
| How to test?      | Try reproduce issue in ticket. Make sure all translations are available on translation fail from your language before reporting another issue.

### DIFF
--- a/classes/Parameters/UpgradeFileNames.php
+++ b/classes/Parameters/UpgradeFileNames.php
@@ -46,6 +46,8 @@ class UpgradeFileNames
 
     const RESTORE_CONFIG_FILENAME = 'restore_config.var';
 
+    const LANGUAGE_CONFIG_FILENAME = 'language_config.var';
+
     /**
      * during upgradeFiles process,
      * this files contains the list of files left to upgrade in a serialized array.
--- a/classes/UpgradeContainer.php
+++ b/classes/UpgradeContainer.php
@@ -30,6 +30,7 @@ use PrestaShop\Module\AutoUpgrade\Log\WebLogger;
 use PrestaShop\Module\AutoUpgrade\Parameters\ConfigurationStorage;
 use PrestaShop\Module\AutoUpgrade\Parameters\ConfigurationValidator;
 use PrestaShop\Module\AutoUpgrade\Parameters\FileStorage;
+use PrestaShop\Module\AutoUpgrade\Parameters\LanguageConfiguration;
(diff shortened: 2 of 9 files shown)


Quetzacoalt91 (approved):

classes/UpgradeContainer.php
```diff
@@ -651,11 +655,19 @@ public function getTranslator(): Translator
...
             // @phpstan-ignore booleanAnd.rightAlwaysTrue (If PrestaShop core is not instantiated properly, do not try to translate)
             if (method_exists('\Context', 'getContext') && \Context::getContext()->language) {
-                $locale = \Context::getContext()->language->iso_code;
+                $newConfiguration = [
+                    LanguageConfiguration::ISO_LANGUAGES => \Context::getContext()->language->iso_code,
+                ];
+
+                $languageConfiguration->merge($newConfiguration);
+                $this->configurationStorage->save($languageConfiguration);
             }
```
M0rgan01, line 667: maybe we want to add a condition to know if the configuration already exists? to avoid a multitude of configuration updates during the process

classes/Parameters/UpgradeFileNames.php
```diff
@@ -46,6 +46,8 @@ class UpgradeFileNames
 
     const RESTORE_CONFIG_FILENAME = 'restore_config.var';
 
+    const LANGUAGE_CONFIG_FILENAME = 'language_config.var';
```
M0rgan01, line 49: do we want to clean this file after a process? like we do with others

paulnoelcholot:
Hello @ga-devfront,

I tested your PR and it's good for me! :tada: 

Thanks! 

image

paulnoelcholot (approved):

ga-devfront:
After reflection I think that the solution is not totally viable because we are basing ourselves on a single language for all users but potentially 2 users with different languages ​​can access the module pages at the same time which can cause a translation problem.
I think I implemented the saving of an associative array: user_id => iso

PR #1228, opened 2025-03-07
Fix display of error 404 page

Quetzacoalt91:
| Questions         | Answers
| Description?      | Any redirection
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | When the next route is supposed to lead to the page 404, you get the 404 error page, not the bad response error.

### DIFF
--- a/classes/Router/Router.php
+++ b/classes/Router/Router.php
@@ -47,7 +47,7 @@ class Router
         $routeName = $request->query->get('route') ?? Routes::HOME_PAGE;
         $redirected = $request->query->get('_redirected') === '1';
 
-        $route = isset(RoutesConfig::ROUTES[$routeName]) ? $routeName : Routes::ERROR_404;
+        $route = $routeName = isset(RoutesConfig::ROUTES[$routeName]) ? $routeName : Routes::ERROR_404;
 
         if (!$redirected) {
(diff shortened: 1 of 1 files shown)


classes/Router/Router.php
```diff
@@ -47,7 +47,7 @@ public function handle(Request $request)
         $routeName = $request->query->get('route') ?? Routes::HOME_PAGE;
         $redirected = $request->query->get('_redirected') === '1';
 
-        $route = isset(RoutesConfig::ROUTES[$routeName]) ? $routeName : Routes::ERROR_404;
+        $route = $routeName = isset(RoutesConfig::ROUTES[$routeName]) ? $routeName : Routes::ERROR_404;
```
ga-devfront, line 50: If we doing this the 
```
if ($routeName !== $route) {
                $this->dirtyRedirectToRoute($request, $route);
            }
```
is always false then we never doing directyRedirectToRoute

Quetzacoalt91: Are you sure? `$route` is still modified a few lines later if the middleware decides there is a redirection to do.

ga-devfront (approved):

PR #1155, opened 2025-02-04
Do not overwrite configuration

Hlavtox:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Adds a bit more resiliency to the upgrade process. If a shop has backported some logic from newer versions, this will prevent the attempts of overwriting the configuration keys, if they are already in the database. `I did not check, but it would either fail or insert a duplicate entry.`
| Type?             | improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     |
| Sponsor company   | 
| How to test?      |

PR #1234, opened 2025-03-11
Add missing scripts update from 1.7.0 to 8.2.0

M0rgan01:
| Questions         | Answers
| Description?      | Add missing scripts update from 1.7.0 to 8.2.0
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | check that the SQL instructions are executed correctly. You can use https://github.com/PrestaShop/SeamlessUpgradeToolbox for that


ga-devfront:
image
Something is wrong with the CI

upgrade/php/ps_1770_add_states.php
```diff
@@ -0,0 +1,78 @@
...
+        'Tamil Nadu' => 'TN',
+        'Telangana' => 'TG',
+        'Tripura' => 'TR',
+        'Uttarakhand' => 'UT',
+        'Uttar Pradesh' => 'UP',
+        'West Bengal' => 'WB',
+    ];
+
+    foreach ($states as $key => $value) {
+        $query = 'INSERT INTO `' . _DB_PREFIX_ . 'state` (`id_country`, `id_zone`, `name`, `iso_code`, `tax_behavior`, `active`) VALUES (' . $row['id_country'] . ', ' . $row['id_zone'] . ', \'' . $key . '\', \'' . $value . '\', 0, 1);';
```
Quetzacoalt91: Shall we use some `pSQL` with the data retrieved from the other table?

upgrade/sql/1.7.2.0.sql
```diff
@@ -3,6 +3,8 @@ SET NAMES 'utf8';
 
 ALTER TABLE `PREFIX_store` MODIFY `hours` text;
 
+INSERT INTO `PREFIX_configuration` (`name`, `value`, `date_add`, `date_upd`) VALUES ('PS_PRODUCT_SHORT_DESC_LIMIT', '800', NOW(), NOW());
```
Quetzacoalt91: You'll need to rebase in order to retrieve #1155, then you could write

```suggestion
/* PHP:add_configuration_if_not_exists('PS_PRODUCT_SHORT_DESC_LIMIT', '800'); */;
```

upgrade/php/ps_1770_add_states.php
Quetzacoalt91: As we deal with integers

```suggestion
        $query = 'INSERT INTO `' . _DB_PREFIX_ . 'state` (`id_country`, `id_zone`, `name`, `iso_code`, `tax_behavior`, `active`) VALUES (' . (int) $row['id_country'] . ', ' . (int) $row['id_zone'] . ', \'' . $key . '\', \'' . $value . '\', 0, 1);';
```

Quetzacoalt91 (dismissed):

AureRita (commented):
Hi @M0rgan01 

Thank you for your PR, I tested it, All seems to works as expected with all the different version except the case in 1.7.8.0, we have a dupplicate of a configuration as you can see : 

image
 
 I reproduced it with the SeamlessUpgradeToolbox, 
 from version : 1.7.6.0 
 to version :  8.2.1 
 
 With PHP 7.2 
 
 Waiting for feedback

AureRita (approved):
Hi @M0rgan01 

image


Tested from : 
1.7.1.0 to 1.7.2.0
1.7.1.0 to 1.7.3.0
1.7.1.0 to 1.7.4.0
1.7.1.0 to 1.7.6.0
1.7.6.0 to 1.7.7.0
1.7.6.0 to 8.2.1

Nice Catch ! 

Thank you

PR #1241, opened 2025-03-14
[SQL scripts] Add verification for states and delete update query configuration

M0rgan01:
| Questions         | Answers
| Description?      | Add verification for states and delete update query configuration
| BC breaks?        | no
| Deprecations?     |no
| Fixed ticket?     | -
| Sponsor company   |-
| How to test?      | Update to 1.7.7.0 and 8.0.0 should work, Update from 1.7.6.0 it should not have a warning during Update database part

### DIFF
--- a/upgrade/php/ps_1770_add_states.php
+++ b/upgrade/php/ps_1770_add_states.php
@@ -66,11 +66,11 @@ function ps_1770_add_states()
     ];
 
     foreach ($states as $key => $value) {
-        $row = DbWrapper::getRow(
+        $duplicateRow = DbWrapper::getRow(
             'SELECT * FROM `' . _DB_PREFIX_ . 'state` WHERE name =  \'' . $key . '\' AND id_country = ' . (int) $row['id_country'] . ''
         );
 
-        if (!empty($row)) {
+        if (!empty($duplicateRow)) {
             continue;
         }
 
--- a/upgrade/php/ps_1770_add_states.php
+++ b/upgrade/php/ps_1770_add_states.php
@@ -66,6 +66,14 @@ function ps_1770_add_states()
     ];
 
     foreach ($states as $key => $value) {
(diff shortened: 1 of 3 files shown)


upgrade/php/ps_1770_add_states.php
```diff
@@ -66,6 +66,14 @@ function ps_1770_add_states()
...
 
     foreach ($states as $key => $value) {
+        $row = DbWrapper::getRow(
+            'SELECT * FROM `' . _DB_PREFIX_ . 'state` WHERE name =  \'' . $key . '\' AND id_country = ' . (int) $row['id_country'] . ''
+        );
+
+        if (!empty($row)) {
+            continue;
+        }
+
```
Quetzacoalt91, line 76: I would rather recommend instead the use of `ON DUPLICATE KEY UPDATE`, because the current suggestion overwrites the previous variable `$row` only to retrieve the same data.

Plus, it is expected to assert there is no existing line, and if `$row` is **actually empty**... it makes `$row['id_zone']` and `$row['id_country']` undefined? How the next request is supposed to work?

M0rgan01: Thanks for the review ;) ON DUPLICATE KEY UPDATE would have been a good option if we had uniqueness constraints other than the primary key, which isn't the case here.

As for the rest, it's simply the variable in the loop that accidentally has the same name as above; this will be changed :)

Quetzacoalt91 (approved):

Quetzacoalt91 (approved):

AureRita (approved):
Hi @M0rgan01 

image

image



Tested from : 
1.7.7.0 to 8.0.0
1.7.6.0 to 8.0.0 
8.0.0 to 8.2.1 
1.7.0.6 to 1.7.1.1

Thank you

PR #1290, opened 2025-04-04
Fix dependancies collisions on v9.0.0

M0rgan01:
| Questions         | Answers
| Description?      | Fix dependancies collisions on v9.0.0
| Type?             | bug fix
| BC breaks?        |  no
| Deprecations?     | no
| Fixed ticket?     | https://github.com/PrestaShop/PrestaShop/issues/38315
| Sponsor company   | -
| How to test?      | See https://github.com/PrestaShop/PrestaShop/issues/38315. Also use `update:check-new-version` command

### DIFF
--- a/classes/UpgradeContainer.php
+++ b/classes/UpgradeContainer.php
@@ -67,8 +67,12 @@ use PrestaShop\Module\AutoUpgrade\UpgradeTools\Translation;
 use PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator;
 use PrestaShop\Module\AutoUpgrade\Xml\ChecksumCompare;
 use PrestaShop\Module\AutoUpgrade\Xml\FileLoader;
+use Symfony\Component\Console\Helper\Helper;
+use Symfony\Component\Console\Helper\Table;
+use Symfony\Component\Console\Helper\TableStyle;
 use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\Filesystem\Filesystem;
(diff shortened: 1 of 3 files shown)


classes/UpgradeContainer.php
```diff
@@ -1029,6 +1033,17 @@ public function getUrlGenerator(): UrlGenerator
         return $this->urlGenerator;
     }
 
+    /**
+     * Loads classes that should normally have collisions with Prestashop dependencies.
+     */
+    public function loadNecessaryClasses(): void
+    {
+        class_exists(Table::class);
```
Quetzacoalt91, line 1042: Only to provide some context when we come back later on the code

```suggestion
        // Avoid collision while running on PrestaShop v9
        class_exists(Table::class);
```

M0rgan01: I can specify the version in the PHP doc instead, since there is already a comment

Quetzacoalt91: It's only in case in the future we get more PrestaShop version with conflicts. i.e 9.1 brings more class to pre-load, then 9.2 etc...

M0rgan01: I hope we will have another solution in the future :)

Quetzacoalt91 (approved):

Quetzacoalt91:
I needed to run the module on PrestaShop 9.0 today, and may confirm this PR fixes an issue preventing me to reach the requirements results.
Capture d’écran du 2025-04-15 11-45-55

PR #1298, opened 2025-04-08
[FIX] issue then multiple employee snoozed modal

ga-devfront:
| Questions         | Answers
| Description?      | The modal is displayed continuously if we are not the first user. After investigation this is due to the array_filter which will keep the original key of the user which will therefore not be 0 afterwards.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | N/A
| How to test?      | See internal ticket


tblivet (approved):

AureRita (approved):
Hi @ga-devfront 

https://github.com/user-attachments/assets/d6e63d22-a8fd-4a52-8b9b-c7d12cf359f4



Tested from : 
8.0.3

Thank you

PR #1304, opened 2025-04-09
Improve web archive retrieval

M0rgan01:
| Questions         | Answers
| Description?      | We no longer display versions lower than the current one in the archives, we no longer display versions that do not have the corresponding xml or zip in the archive
| BC breaks?        | no
| Deprecations?     |no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | use the `update:check-new-version` CLI command and the local channel on the web ui.


classes/Services/LocalVersionFilesService.php
```diff
@@ -117,4 +117,31 @@ private function groupFilesByVersion(array $files, string $type, array &$grouped
...
+            if (!empty($files[self::TYPE_XML])) {
+                $xmlList = array_merge($xmlList, $files[self::TYPE_XML]);
+            }
+        }
+
+        return [
+            self::TYPE_ZIP => $zipList,
+            self::TYPE_XML => $xmlList,
+        ];
+    }
```
Quetzacoalt91, line 146: Can you please add a unit test for this new function?

M0rgan01: image

M0rgan01: I will do it, of course :)

Quetzacoalt91 (approved):

ga-devfront (approved):

AureRita (approved):
Hi @M0rgan01 

image

https://github.com/user-attachments/assets/da444d01-a2e3-430f-aed1-85e4f81a64ce


Tested from : 
8.0.5 to 8.1.2 
8.1.2 to 8.2.1
8.0.5 to 8.2.1

Thank you

PR #1307, opened 2025-04-11
[FIX] segment update dialog async call

ga-devfront:
| Questions         | Answers
| Description?      | Actually then you click on "udate" from update dialog if your shop redirect you too fast you can't send the event from segment. This PR fix this with async/await.
| Type?             | bug fix
| BC breaks?        |  no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | try "update" button on modal and check if something is send to segment/mixpanel


tblivet (approved):

Quetzacoalt91 (approved):

AureRita (approved):
Hi @ga-devfront 

image


Tested from : 
8.0.4 to 9.0.0

Thank you

PR #1317, opened 2025-04-15
Handle manually the creation of symlinks while creating a ZIP archive

Quetzacoalt91:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Now that we have handled symlinks while upgrading PrestaShop, it appears we now have to handle them on backup and restore.
| Type?             | improvement
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/38493
| Sponsor company   | @PrestaShopCorp
| How to test?      | Making a backup and restoring a store with symlinks does not fail, and the extraction of a new release of PrestaShop with symlinks (i.e 9.0.0) still works.<ul><li>Update to PrestaShop v9 (from any version of PrestaShop). The update must succeed and a symbolic link `admin-dev/bundles/apiplatform` must exist.</li><li>Make a backup of a store running v9, try updating to a fake PrestaShop v.9.1. The process must succeed.</li><li>Restore to PrestaShop v9, The process must succeed.</li><li>Delete the symbolic link created by PrestaShop (`admin-dev/bundles/apiplatform`), then restore again to PrestaShop v9. It must be restored by the process.</li></ul>

## Roadmap:

- [x] Create a test suite to completely test the management of symlinks in ZIP archives
- [x] Handle symlinks on backup
- [x] Handle symlinks on restore
- [x] Improve previously merged https://github.com/PrestaShop/autoupgrade/pull/1187 with management of symlinks on update.

PR #1333, opened 2025-04-28
Add content verification during ZIP extraction and file copy operations by comparing file MD5 checksums.

M0rgan01:
| Questions         | Answers
| Description?      | Add content verification during ZIP extraction and file copy operations by comparing file MD5 checksums. If the MD5 checksums match, the process continues with the next file without interruption. If a mismatch is detected, the process stops immediately as previously implemented.
| Type?             |improvement
| BC breaks?        |  no
| Deprecations?     |  no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | do an update and a restore with permission problems on certain files which have not changed, do the same thing with files that have evolved by checking that there is indeed a blocking error as before


classes/Task/Update/UpdateFiles.php
```diff
@@ -184,6 +184,17 @@ protected function upgradeThisFile($orig): bool
 
                 return true;
             } catch (IOException $e) {
+                $checksumOrig = md5_file($orig);
+                $checksumDest = md5_file($dest);
+
+                $isFileUnchanged = $checksumOrig === $checksumDest;
+
+                if ($isFileUnchanged) {
+                    $this->logger->warning($this->translator->trans('Unable to copy %1$s – source and destination appear identical, skipping.', [$file]));
```
Quetzacoalt91: This wording makes me think we won't even try to copy if the file is the same. Here is another version:

```suggestion
                    $this->logger->warning($this->translator->trans('Unable to copy %s, but ignoring as source and destination appear identical.', [$file]));
```

classes/ZipAction.php
```diff
@@ -165,21 +165,37 @@ public function extract(string $from_file, string $to_dir): bool
...
-                $error = $zip->getStatusString();
+            $fileName = $zip->getNameIndex($i);
+            if (!$zip->extractTo($to_dir, [$fileName])) {
+                $stat = $zip->statIndex($i);
+                $filePath = $stat['name'];
+
+                $fullSystemPath = rtrim($to_dir, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $filePath;
+                $md5FromZip = md5($zip->getFromIndex($i));
+                $md5FromSystem = is_file($fullSystemPath) ? md5_file($fullSystemPath) : null;
+                $isFileUnchanged = $md5FromZip === $md5FromSystem;
```
Quetzacoalt91: Can we export this to a dedicated method? This would allow some tests to be added as well. (I see for instance an interesting condition about a potential folder / file)

classes/ZipAction.php
```diff
@@ -165,21 +165,37 @@ public function extract(string $from_file, string $to_dir): bool
...
-            if (!$zip->extractTo($to_dir, [$zip->getNameIndex($i)])) {
-                $error = $zip->getStatusString();
+            $fileName = $zip->getNameIndex($i);
+            if (!$zip->extractTo($to_dir, [$fileName])) {
+                $stat = $zip->statIndex($i);
+                $filePath = $stat['name'];
+
+                $fullSystemPath = rtrim($to_dir, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $filePath;
+                $md5FromZip = md5($zip->getFromIndex($i));
+                $md5FromSystem = is_file($fullSystemPath) ? md5_file($fullSystemPath) : null;
```
Quetzacoalt91:
```suggestion
                $md5FromDisk = is_file($fullSystemPath) ? md5_file($fullSystemPath) : null;
```

`System` makes me feed we are talking about the System32 folder of Windows, or another folder of an operating software.

classes/ZipAction.php
```diff
@@ -165,21 +165,37 @@ public function extract(string $from_file, string $to_dir): bool
...
+                $isFileUnchanged = $md5FromZip === $md5FromSystem;
+
+                $error = $zip->getStatusString() ?: 'Unknown extraction error (possibly permission denied or filesystem error)';
```
Quetzacoalt91:
```suggestion
                $error = $zip->getStatusString() ?: $this->translator->trans('Unknown extraction error (possible permission issue or filesystem error)');
```

classes/ZipAction.php
```diff
@@ -165,21 +165,37 @@ public function extract(string $from_file, string $to_dir): bool
...
+
+                if ($isFileUnchanged) {
+                    $error .= ' – file unchanged, skipped';
+                }
```
Quetzacoalt91:
```suggestion
                $nonCriticalErrorNotice = $isFileUnchanged ? $this->translator->trans(', but ignoring as source and destination appear identical') : '';
```

classes/ZipAction.php
```diff
@@ -165,21 +165,37 @@ public function extract(string $from_file, string $to_dir): bool
...
 
-                $this->logger->error(
+                $this->logger->warning(
                     $this->translator->trans(
                         'Could not extract %file% from archive: %error%',
                         [
-                            '%file%' => $zip->statIndex($i)['name'],
+                            '%file%' => $filePath,
```
Quetzacoalt91: On https://github.com/PrestaShop/autoupgrade/pull/1317/files#diff-8e29e32fd69cad53f2beb198e81f9338413e6f7db022df279eaa5b2cba5b987dR218, I've set a new variable `$nameIndex` based on the value of `$zip->getNameIndex($i);`, if you want to avoid an incoming conflict.

classes/ZipAction.php
```diff
@@ -165,21 +165,37 @@ public function extract(string $from_file, string $to_dir): bool
...
+                $error = $zip->getStatusString() ?: 'Unknown extraction error (possibly permission denied or filesystem error)';
+
+                if ($isFileUnchanged) {
+                    $error .= ' – file unchanged, skipped';
+                }
 
-                $this->logger->error(
+                $this->logger->warning(
                     $this->translator->trans(
                         'Could not extract %file% from archive: %error%',
```
Quetzacoalt91: A bit more complex to read, but what do you think about:

```php                        
                $this->logger->warning(
                    $this->translator->trans(
                        'Could not extract %file% from archive%nonCriticalErrorNotice%: %error%.',
                        [
                            '%file%' => $filePath,
                            '%nonCriticalErrorNotice%' => $nonCriticalErrorNotice,
                            '%error%' => $error,
                        ]
                    )
                );
```

I've moved the addition of the notice in the middle of the existing text, for the main reason the `%error%` contents may be in another language, we would make the whole sentence weird if we switch languages 2 times.

An example of result would be

> Could not extract classes/index.php from archive, but ignoring as source and destination appear identical: Unknown extraction error (possibly permission denied or filesystem error).

classes/ZipAction.php
M0rgan01: After thinking about it, I would use "issue" rather than "error" since we are dealing with a warning log. Same for `$error`

classes/ZipAction.php
M0rgan01: `$nonCriticalErrorNotice ` -> `$nonCriticalIssueNotice`

classes/ZipAction.php
Quetzacoalt91: yep, that's even better

Quetzacoalt91 (approved):

ingridusta (approved):
Hi @M0rgan01 & @Quetzacoalt91, 

Tested in 8.1.7. 
Upgrades and restores are no more blocked with permissions management. 

https://github.com/user-attachments/assets/8fc6775a-c0fb-4309-80d6-f5a5407a3199

QA approved ✅

PR #1337, opened 2025-05-02
[FIX] Duplicate segment call

ga-devfront:
| Questions         | Answers
| Description?      | We currently have an issue with auto-tracking that triggers the event multiple times. This PR aims to fix this issue.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | See internal related ticket

### DIFF
--- a/_dev/src/ts/api/SegmentApi.ts
+++ b/_dev/src/ts/api/SegmentApi.ts
@@ -21,6 +21,7 @@ import { maskSensitiveInfoInUrl } from '../appUI/utils/urlUtils';
 
 class Analytics {
   analytics: AnalyticsBrowser;
+  #isAutoTrackSet = false;
 
   /**
    * @description Initializes the Analytics class and loads the Segment analytics with the provided write key.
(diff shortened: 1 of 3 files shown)


_dev/src/ts/api/SegmentApi.ts
```diff
@@ -94,4 +102,6 @@ class Analytics {
   };
 }
 
-export default Analytics;
+const analyticsInstance = new Analytics();
+
+export default analyticsInstance;
```
Quetzacoalt91, line 107: Is this change enough to prevent the duplicated events? No need of the new class variable `#isAutoTrackSet` ?

ga-devfront: Yes, #isAutoTrackSet is only used if the same instance tries to re-set autotrack (a security feature that I added), however the problem comes from duplication, which comes from the fact that there were 2 instances. By creating the instance directly and sharing it, the problem disappears.

tblivet (approved):

AureRita (approved):
Hi @ga-devfront 

Capture d’écran du 2025-05-05 11-50-42

https://github.com/user-attachments/assets/d4a5b12f-21ef-4df7-b001-83af9d85fa81

Tested from : 
1.7.8.9 to 8.2.1

Thank you

ps-jarvis:
PR merged, well done!

PR #1344, opened 2025-05-06
[FIX] handle link to admin maintenance on PS 1.7.4

ga-devfront:
| Questions         | Answers
| Description?      | We are using a hardcoded route for version 1.7.4 because handling `getAdminLink` for the `AdminMaintenance` controller results in an error.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | N/A
| How to test?      | Check related internal ticket


Quetzacoalt91 (approved):

AureRita (approved):
Hi @ga-devfront 

image


Tested from : 
1.7.4.2 to 8.2.1
1.7.4.4 to 8.2.1

Thank you

PR #1374, opened 2025-05-19
Add missing hooks for PS v8.2.2 and v9.0.0

Quetzacoalt91:
| Questions         | Answers
| Description?      | Changes have been brought to the Core repository but are missing on this one. This PR sync the list, otherwise we won't be able to release Update Assistant 7.1 once PS v9 is ready.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Related to https://github.com/PrestaShop/PrestaShop/pull/38232, https://github.com/PrestaShop/PrestaShop/pull/38408, https://github.com/PrestaShop/PrestaShop/pull/38488 and https://github.com/PrestaShop/PrestaShop/pull/38567
| How to test?      | Clean installation of PS v9 and updates to this version bring the same changes.

### DIFF
--- a/upgrade/sql/9.0.0-catchup.sql
+++ b/upgrade/sql/9.0.0-catchup.sql
@@ -60,7 +60,6 @@ UPDATE `PREFIX_hook_module` AS hm
 INNER JOIN `PREFIX_hook` AS hfrom ON hm.id_hook = hfrom.id_hook AND hfrom.name = 'displayAdminOrderSideBottom'
 INNER JOIN `PREFIX_hook` AS hto ON hto.name = 'displayAdminOrderBottom'
 SET hm.id_hook = hto.id_hook;
-DELETE FROM `PREFIX_hook` WHERE name = 'displayAdminOrderSideBottom';
 
 UPDATE `PREFIX_country` SET `zip_code_format` = 'NNNNN' WHERE `iso_code` = "KR";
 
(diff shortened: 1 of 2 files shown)


upgrade/sql/9.0.0.sql
```diff
@@ -400,7 +410,8 @@ DELETE FROM `PREFIX_hook` WHERE `name` IN (
   'actionAdminLoginControllerResetAfter',
   'actionBeforeEnableMobileModule',
   'actionBeforeDisableMobileModule',
-  'actionAjaxDieBefore'
+  'actionAjaxDieBefore',
+  'createAccount'
```
M0rgan01: strange, I don't see it in the hooks of v8.2.1, but it is present in the hook_alias table

https://github.com/PrestaShop/PrestaShop/blob/8.2.x/install-dev/data/xml/hook_alias.xml#L100C34-L100C36

upgrade/sql/9.0.0-catchup.sql
```diff
@@ -60,7 +60,6 @@ UPDATE `PREFIX_hook_module` AS hm
 INNER JOIN `PREFIX_hook` AS hfrom ON hm.id_hook = hfrom.id_hook AND hfrom.name = 'displayAdminOrderSideBottom'
 INNER JOIN `PREFIX_hook` AS hto ON hto.name = 'displayAdminOrderBottom'
 SET hm.id_hook = hto.id_hook;
-DELETE FROM `PREFIX_hook` WHERE name = 'displayAdminOrderSideBottom';
```
M0rgan01, line 63: This was removed in 1.7 in favor of "displayAdminOrderBottom", but it was actually brought back in v9... strange

M0rgan01: https://github.com/PrestaShop/PrestaShop/blob/9.0.x/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/view.html.twig#L59-L65 There is only one use left, that's why the generation "put it back"

upgrade/sql/9.0.0.sql
Quetzacoalt91: and this alias is not new ! https://github.com/PrestaShop/PrestaShop/commit/d1a4399630b3b27346348977c9579f51c508c03f#diff-d1c7bb639cb0862436f0b318985438bc060b20be10a0ae50d4ebc246b95fe77aR105

ga-devfront (dismissed):

ga-devfront (approved):

M0rgan01 (approved):

AureRita (approved):
Hi @Quetzacoalt91 

image


Tested from : 
8.2 to 9.0.0
8.0.3 to 9.0.0

FYI : there is no missing hook, but there are two hooks that are not in the basic v9 with an upgrade from 8.0.3 and only one with an upgrade from 8.2.0 
(`createAccount` and `displayProductButtons` )

Thank you

PR #1396, opened 2025-06-11
Update composer.json

tm8544:
Prestahop 9.0 has support for PHP 8.1 - 8.4,
but Autoupgrade does not have even PHP 8.1 support due to version 1.8.0 of segmentio/analytics-php.

Bump segmentio/analytics-php version to gain compatibility with supported versions of PHP.
3.6.0 brings PHP8.1 * 8.2 support.
3.8.0 brings PHP.8.3 support.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "ext-zip": "*",
         "symfony/filesystem": "^3.0",
         "doctrine/collections": "^1.8",
-        "segmentio/analytics-php": "^1.8",
+        "segmentio/analytics-php": "^3.6",
         "symfony/dotenv": "^4.4",
         "symfony/http-foundation": "^4.4",
         "symfony/console": "^3.4",


Quetzacoalt91:
I'm afraid this can't be added because the version 2+ of `segmentio/analytics-php` requires PHP 7.4 while the module is compliant with PHP 7.1+.

What issue did you encounter while being on PS 9? Did it happen during a `composer require`?

tm8544:
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in \kauppa\modules\autoupgrade\vendor\segmentio\analytics-php\lib\Segment.php on line 112
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in \kauppa\modules\autoupgrade\vendor\segmentio\analytics-php\lib\Segment\Consumer\ForkCurl.php on line 47
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in \kauppa\modules\autoupgrade\vendor\segmentio\analytics-php\lib\Segment\Consumer\ForkCurl.php on line 82
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in \kauppa\modules\autoupgrade\vendor\segmentio\analytics-php\lib\Segment\Consumer\ForkCurl.php on line 82
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in \kauppa\modules\autoupgrade\vendor\segmentio\analytics-php\lib\Segment\Consumer\LibCurl.php on line 72
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in \kauppa\modules\autoupgrade\vendor\segmentio\analytics-php\lib\Segment\Consumer\LibCurl.php on line 72
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in \kauppa\modules\autoupgrade\vendor\segmentio\analytics-php\lib\Segment\Consumer\Socket.php on line 182
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in \kauppa\modules\autoupgrade\vendor\segmentio\analytics-php\lib\Segment\Consumer\Socket.php on line 182

Quetzacoalt91:
If these errors only happen during a process (update / restore / backup) and do not prevent them to run completely, we can stay like this until we need to split the supported versions of PrestaShop in two separate versions of Update Assistant.

tm8544:
Ok.

PR #1408, opened 2025-06-16
[FIX] module migration files

ga-devfront:
| Questions         | Answers
| Description?      | Currently we got and issue then trying to update a module that try to access internal files (bad practice). This PR will fix it.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | Try upgrade from PrestaShop `1.7.8.3` to `8.2.1` with `ps_accounts` version `5.0.3`. If everything work the fix is done !


Quetzacoalt91 (approved):

AureRita (approved):
Hi @ga-devfront 

https://github.com/user-attachments/assets/4c46770e-58d0-48d0-b660-820882e9aafa



Tested from : 
1.7.8.3 to 8.2.1 with ps_account 7.0.3

Thank you

PR #1411, opened 2025-06-23
Defer the creation of links to increase the chances the destination exists

Quetzacoalt91:
| Questions         | Answers
| Description?      | Following the PHP documentation about symlink() whch states on some OSes the destination must exist in order the link creation to work, we defer the creation of links after the whle zip extraction.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | https://github.com/PrestaShop/PrestaShop/issues/38892
| How to test?      | On Unix the result should be the same, on Windows the extraction of PS v9 should (now?) work

### DIFF
--- a/classes/ZipAction.php
+++ b/classes/ZipAction.php
@@ -182,27 +182,15 @@ class ZipAction
             return false;
         }
 
+        /** @var int[] */
+        $deferredLinks = [];
+
         for ($i = 0; $i < $zip->numFiles; ++$i) {
             $nameIndex = $zip->getNameIndex($i);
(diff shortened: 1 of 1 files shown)


M0rgan01 (approved):

AureRita (approved):
Hi @Quetzacoalt91 

https://github.com/user-attachments/assets/4807d7f8-6a08-4fc4-9306-fe6fba576de4


Tested from : 

8.1.7 to 9.0.0
8.0.3 to 9.0.0

Thank you

PR #1435, opened 2025-07-29
Remove Tools14

Hlavtox:
| Questions         | Answers
| Description?      | Methods in Tools14 are the same as the ones in 1.7.0.0, so we don't need them anymore.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | Automatic test upgrading 1.7.0.0 to latest versions should reveal any issue.

cc @Quetzacoalt91


classes/Task/Backup/BackupDatabase.php
```diff
@@ -31,7 +31,6 @@
 use PrestaShop\Module\AutoUpgrade\Task\ExitCode;
 use PrestaShop\Module\AutoUpgrade\Task\TaskName;
 use PrestaShop\Module\AutoUpgrade\Task\TaskType;
-use PrestaShop\Module\AutoUpgrade\Tools14;
```
M0rgan01, line 34: use Tools;

classes/UpgradeTools/Translation.php
```diff
@@ -22,7 +22,6 @@
 namespace PrestaShop\Module\AutoUpgrade\UpgradeTools;
 
 use PrestaShop\Module\AutoUpgrade\Log\LoggerInterface;
-use PrestaShop\Module\AutoUpgrade\Tools14;
```
M0rgan01, line 25: use Tools;

M0rgan01 (dismissed):

kpodemski:
Having an extra layer above such methods, in my opinion, is quite beneficial. This way, you can be sure about their behavior and don't worry about future changes in the core. The more independent from the core the module is, the better, again, in my opinion. The decision is @Quetzacoalt91 

Although `nl2br` or `getValue` are complete nonsense there 😅

Edit: I see that there are no other uses of this class apart from these few occurrences, so it should be good.

Quetzacoalt91:
Thanks @Hlavtox, removing Tools14 is a good idea.
There is one thing about this to keep in mind. Tools (from the core) will be loaded for sure while the controllers are called, but during a update / backup / restore process, the core (and its `Tools` class) may be unavailable. The change to apply could be more complex than only replacing `Tools14` with `Tools`.

M0rgan01:
@kpodemski I understand this point of view; but in this case it is not only about the Tools class, but about all the uses of the Core that we have on the module. I also think that in the case where the tools class would become incompatible, the min/max range would come to be updated, like the other modules

M0rgan01:
@Quetzacoalt91 It is likely that the symfony we are embedding has the same functions as the tools class, for the most part.

Hlavtox:
Well, I think a best way would be to have a completely independent codebase to process the upgrade. May sound crazy, but it would be the most reliable way.

It's a miracle this whole upgrade process works when it just replaces the code under it's hands. :-)

M0rgan01:
@Hlavtox Yes, the idea for it to be something reliable would actually be an external app, not a module that has the slightest dependency on the core

AureRita:
Hi @Hlavtox 

Thank you for your PR, first of all, can you resolve the conflict with the files "controllers/admin/AdminSelfUpgradeController.php" 

Moreover, I try to modify each of my files to have the same things than your PR, When I want to do an upgrade on UI, I have an exeption 500 as you can see : 


https://github.com/user-attachments/assets/bf25a6d0-ba9b-4b60-b74c-b2453e439643

We also see this result on the different CLI that test the UI 

Waiting for feedbacks

Quetzacoalt91:
Replaced by #1459

PR #1438, opened 2025-07-31
Disable upgrade notification modal

Hlavtox:
| Questions         | Answers
| Description?      | I need to bring this up to attention. Just this week we had like 5 customers on Openservis hosting who upgraded and broke their shop because of upgrade to 9.0.0, that doesn't work. When we ask `"Why did you upgrade"`, the customer tells us `"Because it shown a popup, so I clicked it"`. I know there is a environment variable that disables this, but we can't put it to env of every customer, this must be a default behavior. Upgrading a shop is not something that a regular user should do in production. It fails many times if the shop is not prepared. People don't know their modules are not compatible, they are not prepared for it, they forget they have custom modifications. What also doesn't help is that the only version it offers to upgrade is 9.0.0, **we can't push people to a version that is not ready**.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | [OpenServis](https://www.openservis.cz/)
| How to test?      | 

cc @ShaiMagal @Quetzacoalt91 @M0rgan01

### DIFF
--- a/classes/Hooks/DisplayBackOfficeHeader.php
+++ b/classes/Hooks/DisplayBackOfficeHeader.php
@@ -115,6 +115,9 @@ class DisplayBackOfficeHeader
      */
     public function renderUpdateNotification(): string
     {
+        // Temporarily disable this modal before better solution is found
+        return $this->content;
+
         $disableModal = getenv('AUTOUPGRADE_DISABLE_AVAILABLE_UPDATE_MODAL');
         if (filter_var($disableModal, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE)) {
             return $this->content;


M0rgan01:
cc @gericfo

gericfo:
Thanks @Hlavtox  for your feedback and your contribution!

We’re closely monitoring all feedback regarding updates and their performance (especially success rates). That’s precisely why we introduced an environment variable allowing the update notification modal to be disabled.




### New Opportunity – Reintroducing Minor and Patch Updates

We’ve acknowledged the issues experienced with updates to version 9. This has led us to open an opportunity to reintroduce Minor and Patch updates in the module (both Web UI and CLI), instead of only suggesting the latest Major version.




### Global Situation Analysis

Before making a drastic decision like entirely disabling the modal (as you suggested), we took a step back and analyzed the data from the past 30 days:

- **7400 merchants** saw the update notification modal:
  - **5000** (~70%) chose to postpone the update.
  - **2400** (~30%) opened the Update Assistant module.
- Among those **2400 merchants:**
  - **500** (~20%) actually triggered an update.
    - **300** were successful (~60%).
    - **200** failed (~40%).

These numbers show that most users do not proceed with the update after seeing the modal — it’s far from a systematic behavior. Users are making conscious decisions to delay.
They also show that failures are not the majority. In fact, successful updates are more common than failed ones.




### Specific Case: Updates to Version 9

However, we do see a clear difference based on the target version:

- Success rate for updates to **V8: ~80%**
- Success rate for updates to **V9: ~40%**

Improving the success rate for V9 is one of our top priorities — through module compatibility, ecosystem adaptation, and optimizations to the Update Assistant itself. Reintroducing Minor and Patch updates will also help smooth the transition.

Regarding the 5 failed upgrades you mentioned:

- Could you open issues for these failures? This kind of insight is extremely helpful to identify recurring issues and work on concrete improvements.
- Do you also have visibility on how many merchants had a successful update during the same period? It’s important to look at both sides for a complete picture.




### What can we do about the update notification modal?

Now that we have a more comprehensive view of the situation, here are the options we're currently considering:

- **A/ Add an environment variable to disable the modal** → Already available since Update Assistant v7.3.0
- **B/ Completely cut the modal (your suggestion)** → This would be unfortunate, as it would prevent the majority of users who succeed in updating from accessing a helpful feature.
- **C/ Adapt the modal behavior to focus on V8 updates only:**
  - Only show the modal when an update to **V8.x** is available (excluding V9).
  - Re-enable suggestions for **V9** when its success rate improves.

We believe option **C** offers the most balanced approach: it protects users while still promoting updates where the upgrade path is more reliable.

Hlavtox:
Hi @gericfo, thank you for your message! :-) I will try to reply step-by-step.

> We’ve acknowledged the issues experienced with updates to version 9. This has led us to open an opportunity to reintroduce Minor and Patch updates in the module (both Web UI and CLI), instead of only suggesting the latest Major version.

That is very good, appreciate this step. 👍 There is just too much things broken now - images, module upgrades, attribute colors, 

> 7400 merchants saw the update notification modal:
> 5000 (~70%) chose to postpone the update.
> 2400 (~30%) opened the Update Assistant module.

Yes, we can clearly see that absolute majority of merchants don't want to upgrade and don't want to be disturbed by it. And they probably know how dangerous this is. But unfortunately, not all people are like this. They see the modal and just click it. **Upgrade is something to do by the administrator when the time comes, after absolutely perfect preparation.**

I understand you want to push people to newest version possible, trust me, we also do.

(Can't express how much would @ShaiMagal immediately wipe all 1.6 support from our modules. 😄) But we must do it in a careful way.

> Improving the success rate for V9 is one of our top priorities — through module compatibility, ecosystem adaptation, and optimizations to the Update Assistant itself. Reintroducing Minor and Patch updates will also help smooth the transition.

Upgrading PrestaShop is and always will be a dangerous task, but for 9.0.0, yes, it's even more dangerous, looking at the stats. Most of the topics have been opened already. Specifically:
- symlink requirement that makes it crash on majority of public hostings - https://github.com/PrestaShop/PrestaShop/issues/38960
- broken module upgrade mechanism - https://github.com/PrestaShop/PrestaShop/issues/39151
- problems with company modules - https://github.com/PrestaShop/PrestaShop/issues/38983
- clearing cache is broken - https://github.com/PrestaShop/PrestaShop/issues/38596 & https://github.com/PrestaShop/PrestaShop/issues/38139

and that's just things that can break upgrade.

> What can we do about the update notification modal?
> Now that we have a more comprehensive view of the situation, here are the options we're currently considering:
> 
> A/ Add an environment variable to disable the modal → Already available since Update Assistant v7.3.0
> B/ Completely cut the modal (your suggestion) → This would be unfortunate, as it would prevent the majority of users who succeed in updating from accessing a helpful feature.
> C/ Adapt the modal behavior to focus on V8 updates only:
> Only show the modal when an update to V8.x is available (excluding V9).
> Re-enable suggestions for V9 when its success rate improves.
> We believe option C offers the most balanced approach: it protects users while still promoting updates where the upgrade path is more reliable.

You are thinking in a good direction, I agree. My suggestion:
- Add "Never notify again" in the modal.
- Promote only patch updates for now, or maybe forever, until we find a better solution.
- Add a big red alert on the upgrade page that will say something like:
  - That they should never update a production shop.
  - Do it after a careful preparation.
  - Make sure to have also manual backup.
  - Aknowledge that things can break during or after the upgrade.
  - etc.

gericfo:
Hi @Hlavtox! Thanks again for your feedback! Below are my answer to your suggestions:

**- Add "Never notify again" in the modal**
At this stage, we don’t plan to include a "Never notify again" option in the modal. That would essentially mean hiding the modal completely — which is already possible using the environment variable we’ve introduced for that purpose.
As you mentioned _"can't express how much would [**@ShaiMagal**](https://github.com/ShaiMagal) immediately wipe all 1.6 support from our modules"_, we do want to encourage users to update and keep reminding them of this possibility. Otherwise, many users risk staying on outdated versions without even realizing that upgrades are available or needed.


**- Promote only patch updates for now (or maybe forever)**
We plan to remove all V9 suggestions from the update notification modal, and keep only updates to V8 (Major, Minor, and Patch versions). This will prevent the modal from encouraging users to upgrade to V9 until we see significantly improved success rates.


**- Add a big red alert on the upgrade page, warning users not to update in production, etc.**
We’ve chosen not to reintroduce red warning banners in the UI. In fact, we've redesigned the UX of the Update Assistant module specifically to avoid the "Christmas tree effect" of too many warning banners from older versions, and to reassure users during the upgrade process (without creating fear or confusion).

-> However, we will improve the wording and guidance in the UI to reflect your suggestions more clearly:
- Remind users not to upgrade directly on production environments, but to prepare and test on a staging environment instead.
- Recommend a manual backup, in addition to the automatic one proposed by the module.
- Acknowledge that updates may break shops if not properly prepared.

Let me know if this sounds good to you! Thanks :)

Quetzacoalt91:
Following what @gericfo said

> **- Promote only patch updates for now (or maybe forever)**
We plan to remove all V9 suggestions from the update notification modal, and keep only updates to V8 (Major, Minor, and Patch versions). This will prevent the modal from encouraging users to upgrade to V9 until we see significantly improved success rates.

I'm closing this PR in favor of #1449, which is a first step toward a better management of updates with the future introduction of versions recommended.
The result match you short term expectation: Removing the v9 from the notification modal.

Thanks

PR #1472, opened 2025-09-23
Fix: Issue upgrading PrestaShop 9.0.0 on Windows

### LINKED ISSUE #39607: Issue upgrading PrestaShop 9.0.0 on Windows
### Prerequisites

I am experiencing an issue when trying to upgrade PrestaShop to version 9.0.0 on Windows.

When I try the upgrade using the autoupgrade module, I get the following error:

Image

However, if I run the upgrade using Local archive, with the ZIP package generated by the [tools/build/CreateRelease.php](https://github.com/PrestaShop/PrestaShop/blob/9.0.x/tools/build/CreateRelease.php)
 script, the upgrade works fine without any problem.

### Expected behavior

From what I can see, the error happens here:

https://github.com/PrestaShop/autoupgrade/blob/d6c3ea6cd073287e23dcd339b48e06f5e71a3f57/classes/ZipAction.php#L224-L226

At that point, the system tries to create a symlink, which fails on Windows.

The issue seems related to the official package downloaded from the API:
https://api.prestashop-project.org/assets/prestashop-classic/9.0.0-1.0/prestashop.zip

In this package, inside the admin folder, there is a `bundles` subfolder containing the `apiplatform` file.
In the locally generated packages (using CreateRelease.php), this bundles folder is not present.

### Question

Should the bundles folder inside admin actually be there?

If yes, does this mean that upgrading through the autoupgrade module will never work on Windows (due to symlink creation)?

### Steps to reproduce

1. install autoupgrade module
2. upgrade to 9.0.0

8.x

_No response_

_No response_

autoupgrade

@codencode

Codencode:
Fix symlink handling: skip symbolic links on Windows to prevent errors

| Questions         | Answers
| Description?      | On Windows, symbolic links are not handled correctly. To avoid update failures, instead of trying to create symlinks with `Filesystem->symlink`, we simply extract the file.This is not the most elegant solution, but it prevents the upgrade from failing on Windows environments.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/39607
| Sponsor company   | @Codencode
| How to test?      | Update PrestaShop to version 9.0.0 on Windows operating system

### DIFF
--- a/classes/ZipAction.php
+++ b/classes/ZipAction.php
@@ -61,6 +61,11 @@ class ZipAction
      */
     private $prodRootDir;
 
+    /**
+     * @var bool
+     */
+    private $isWinOs;
+
     public function __construct(Filesystem $filesystem, Translator $translator, LoggerInterface $logger, UpgradeConfiguration $updateConfiguration, string $prodRootDir)
     {
         $this->filesystem = $filesystem;
@@ -69,6 +74,7 @@ class ZipAction
         $this->prodRootDir = $prodRootDir;
         $this->configMaxNbFilesCompressedInARow = $updateConfiguration->getNumberOfFilesPerCall();
         $this->configMaxFileSizeAllowed = $updateConfiguration->getMaxFileToBackup();
+        $this->isWinOs = stripos(PHP_OS, 'WIN') === 0;
     }
 
     /**
@@ -222,7 +228,16 @@ class ZipAction
             $nameIndex = $zip->getNameIndex($i);
 
             try {
-                $this->filesystem->symlink($zip->getFromIndex($i), $to_dir . DIRECTORY_SEPARATOR . $nameIndex);
+                if ($this->isWinOs) {
+                    $path = $zip->getFromIndex($i);
+                    if (!file_exists(realpath($path))) {
+                        // The path does not exist, so we are using the path that was extracted from the ZIP archive.
+                        $path = $to_dir . DIRECTORY_SEPARATOR . $this->normalizeWinPath($path);
+                    }
+                    $this->filesystem->symlink($path, $to_dir . DIRECTORY_SEPARATOR . $nameIndex, true);
+                } else {
+                    $this->filesystem->symlink($zip->getFromIndex($i), $to_dir . DIRECTORY_SEPARATOR . $nameIndex);
+                }
             } catch (IOException $e) {
                 $this->logger->warning(
                     $this->translator->trans(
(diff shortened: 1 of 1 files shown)


Codencode:
@PrestaShop/committers 
I understand that it won't be easy to verify the issue and the PR, but on Windows, when updating to version 9.0.0, the process gets stuck at this point:
https://github.com/PrestaShop/autoupgrade/blob/d6c3ea6cd073287e23dcd339b48e06f5e71a3f57/classes/ZipAction.php#L224-L226

The change only impacts Windows, where it currently doesn't work in this situation anyway. With this modification, at least the update is performed.

nicosomb (changes requested):
I understand your issue, but as Windows is not actively supported (see https://devdocs.prestashop-project.org/9/basics/installation/system-requirements/), I don't know if we should have a fix for this OS.

Codencode:
> I understand your issue, but as Windows is not actively supported (see https://devdocs.prestashop-project.org/9/basics/installation/system-requirements/), I don't know if we should have a fix for this OS.

I understand. However, it's about how to install it on Windows.
https://devdocs.prestashop-project.org/9/basics/installation/environments/windows-specific/

Note that the issue does not occur with a ZIP package I created myself using Docker, because the symlink is not created.
Can you tell me how I could look deeper into this?
Because I believe that at least with Docker (on Windows) it should work.

Thanks.

Quetzacoalt91 (commented):
Even if Windows is not officially supported, I took some time to try and confirm the creation of symlinks was possible on this operating system.

Let's have a deeper look to see if there is an actual bug to fix, but I'd like to avoid disabling the feature just because we are on Windows.

Quetzacoalt91:
The PR indeed copies the files when the environment does not allow the symlink. But even when Windows is able to symlink, this PR changes the behavior for a copy of the destination contents.
Because we don't get the same result between installing PrestaShop v9.0.0 and updating to the same version, I don't think we should proceed with this change.

The whole discussion about symlinks with Windows can be found on the related issue https://github.com/PrestaShop/PrestaShop/issues/39607

Even if I don't accept this PR, be aware the incoming 9.0.1 will contain another fix about the symbolic links you encountered. Please wait for its release that should happen in the next couple of weeks.

Thanks!

jolelievre:
Hmm by the way @Quetzacoalt91 since we changed the way the bundles assets are installed I believe we don't have symlinks anymore in the zip. This symlink feature was only added in the autoupgrade module to handle thse assets as fair as I remember.

So maybe we could actually drop it from the module since it's useless, if just removing this allows upgrade on windows it might be a valuable quick win no?

Quetzacoalt91:
We don't have symlink in PrestaShop itself but the feature is kept in the module for the backups and restores (i.e symlinks created by the merchants or hosting providers).

PR #1478, opened 2025-09-26
Update call to bin/console by relying on PHP binary first

Quetzacoalt91:
| Questions         | Answers
| Description?      | When installing the assets for PrestaShop 9.0.1, some environment cannot call `bin/console` without being prefixed by `php`, because of the following error: `/usr/bin/env: 'php': No such file or directory`. This PR tries to rely first on the PHP binary, and directly on the filein case of error. It is even able to fallback on the second call instead of choosing only one to call.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Update to PrestaShop 9.0.1. The Docker image `prestashop/prestashop-flashlight:8.2.3-8.1-fpm-bookworm-nginx` used to fail, and should not with this PR. Other tests should be done to find an environment we've missed even with this fix.

### DIFF
--- a/classes/UpgradeTools/CoreConsoleExecutable.php
+++ b/classes/UpgradeTools/CoreConsoleExecutable.php
@@ -1,0 +1,81 @@
+<?php
+
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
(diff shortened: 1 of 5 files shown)


ga-devfront (dismissed):

ga-devfront:
Please add a milestone @Quetzacoalt91

Quetzacoalt91:
> Please add a milestone @Quetzacoalt91

Usually done right before merging

classes/UpgradeTools/CoreConsoleExecutable.php
ga-devfront, line 1: Did you think it's possible to add some unite tests for this file ?

ga-devfront (dismissed):

classes/UpgradeTools/CoreConsoleExecutable.php
Quetzacoalt91: image

ga-devfront (approved):

ingridusta (approved):
Hello Thomas, 

I reproduced the bug with flashlight images (to 9.0.1 local) from dev env : 
8.2.3-8.1 ❌ 
8.2.2-8.1 ❌ 
8.2.0-8.1 ❌ 
8.1.6-8.1 ✅  (upgrade to 9.0.1 local archives : success)
8.0.5-8.1 ❌ 

With your pr I retested all these versions, and were all successful : 
8.2.3-8.1 ✅ 
8.2.2-8.1 ✅
8.2.0-8.1 ✅
8.1.6-8.1 ✅
8.0.5-8.1 ✅

It's QA ✅

PR #1485, opened 2025-10-07
Introduce a safe mode of PrestaShop to avoid loading modules

Quetzacoalt91:
| Questions         | Answers
| Description?      | When updating shops from classic v8 to classic v9, we encounter many errors regarding the modules.<br><br>When we update from PrestaShop v8 to v9, the gap is quite wide regarding the improvement we're brought on PrestaShop. We made a big step forward on the Symfony version, for instance we have PHP code with strict typing... and this leads to major changes in the modules as well.<br>With this version 9, we encounter a lot of Fatal Error from the modules that don't match the updated core functions signature. And the more modules you have on the shop, higher is the risk of encountering it. In the Update process, it seems we now require a moment where we avoid loading modules until we try to update them. This covers from the **end of files update**, **database**, to the **warm-up of modules update**.<br><br>I a perfect world, I should be able to ask the Core to not load modules. However until the feature in available, we have to find another even though it looks experimental.
| Type?             | New feature
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | <ul><li>https://github.com/PrestaShop/autoupgrade/issues/1560</li></ul>
| How to test?      | Stability is improved during updates between 2 classic versions.

### DIFF
--- a/classes/Task/Update/UpdateDatabase.php
+++ b/classes/Task/Update/UpdateDatabase.php
@@ -70,6 +70,7 @@ class UpdateDatabase extends AbstractTask
                 return ExitCode::SUCCESS;
             }
             $this->container->getFileStorage()->clean(UpgradeFileNames::SQL_TO_EXECUTE_LIST);
+            $this->container->getSafeMode()->disable();
             $this->getCoreUpgrader()->finalizeCoreUpdate();
         } catch (UpgradeException $e) {
             $this->next = TaskName::TASK_ERROR;
@@ -113,7 +114,6 @@ class UpdateDatabase extends AbstractTask
             $this->container->getCacheCleaner()->cleanFolders();
         }
 
-        // Migrating settings file
         $this->container->initPrestaShopAutoloader();
         $this->container->initPrestaShopCore();
     }
@@ -129,6 +129,8 @@ class UpdateDatabase extends AbstractTask
         $this->container->getUpdateState()->setProgressPercentage(
             $this->container->getCompletionCalculator()->getBasePercentageOfTask(self::class)
         );
+
+        $this->container->getSafeMode()->enable();
 
         $this->getCoreUpgrader()->writeNewSettings();
 
--- a/classes/Task/Update/UpdateDatabase.php
+++ b/classes/Task/Update/UpdateDatabase.php
@@ -136,7 +136,7 @@ class UpdateDatabase extends AbstractTask
             $this->logger->info($this->container->getTranslator()->trans('Keeping non native modules enabled'));
         }
 
-        $this->container->getSafeMode()->enable();
+        $this->container->getQuarantineZone()->addAll();
 
         $this->getCoreUpgrader()->writeNewSettings();
 
--- a/classes/Task/Update/UpdateDatabase.php
+++ b/classes/Task/Update/UpdateDatabase.php
@@ -70,7 +70,6 @@ class UpdateDatabase extends AbstractTask
                 return ExitCode::SUCCESS;
             }
             $this->container->getFileStorage()->clean(UpgradeFileNames::SQL_TO_EXECUTE_LIST);
-            $this->container->getSafeMode()->disable();
             $this->getCoreUpgrader()->finalizeCoreUpdate();
         } catch (UpgradeException $e) {
             $this->next = TaskName::TASK_ERROR;
@@ -130,6 +129,13 @@ class UpdateDatabase extends AbstractTask
             $this->container->getCompletionCalculator()->getBasePercentageOfTask(self::class)
         );
 
+        if ($this->container->getUpdateConfiguration()->shouldDeactivateCustomModules()) {
+            $this->logger->info($this->container->getTranslator()->trans('Disabling all non native modules'));
+            $this->getCoreUpgrader()->disableCustomModules();
+        } else {
+            $this->logger->info($this->container->getTranslator()->trans('Keeping non native modules enabled'));
+        }
+
         $this->container->getSafeMode()->enable();
 
         $this->getCoreUpgrader()->writeNewSettings();
@@ -142,13 +148,6 @@ class UpdateDatabase extends AbstractTask
         }
 
         $this->getCoreUpgrader()->setupUpdateEnvironment();
-
-        if ($this->container->getUpdateConfiguration()->shouldDeactivateCustomModules()) {
-            $this->logger->info($this->container->getTranslator()->trans('Disabling all non native modules'));
-            $this->getCoreUpgrader()->disableCustomModules();
-        } else {
-            $this->logger->info($this->container->getTranslator()->trans('Keeping non native modules enabled'));
-        }
 
         return ExitCode::SUCCESS;
     }
--- a/classes/Task/Update/UpdateModules.php
+++ b/classes/Task/Update/UpdateModules.php
@@ -154,6 +154,8 @@ class UpdateModules extends AbstractTask
             return ExitCode::FAIL;
         }
 
+        $this->container->getSafeMode()->disable();
+
         if ($total_modules_to_upgrade) {
             $this->logger->info($this->translator->trans('%s modules will be updated.', [$total_modules_to_upgrade]));
         }
--- a/classes/Task/Update/UpdateModules.php
+++ b/classes/Task/Update/UpdateModules.php
@@ -69,6 +69,7 @@ class UpdateModules extends AbstractTask
 
             try {
                 $this->logger->debug($this->translator->trans('Checking updates of module %module%...', ['%module%' => $moduleInfos['name']]));
+                $this->container->getQuarantineZone()->removeOne($moduleInfos['name']);
 
                 $moduleDownloaderContext = new ModuleDownloaderContext($moduleInfos);
                 $moduleSourceList->setSourcesIn($moduleDownloaderContext);
@@ -124,6 +125,9 @@ class UpdateModules extends AbstractTask
             $this->next = TaskName::TASK_UPDATE_MODULES;
             $this->logger->info($this->translator->trans('%s modules left to check.', [$modules_left]));
         } else {
+            // Remove all remaining modules from the quarantine
+            $this->container->getQuarantineZone()->removeAll();
+
             $this->stepDone = true;
             $this->status = 'ok';
             $this->next = TaskName::TASK_CLEAN_DATABASE;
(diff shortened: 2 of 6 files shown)


Hlavtox:
I like the idea where you are going, but I think that the things that make the store break are mostly not dependent on module being active or not. :-( It's usually PHP stuff - class definitions, services, tabs and other things like this, that just break if the module is installed.

kpodemski:
@Hlavtox this change can lead to the module files not being loaded which should do the job, what I do is that sometimes I'm changing the directory name from `somemodule` to `somemodule-disabled`, thing is that the moment you bring it back, you will have an error anyway

so it's more like doing an update, and fixing later, I kinda like this idea, but it should probably be for advanced users only

> When updating shops from classic v9 to classic v9, we encounter many errors regarding the modules.

You meant classic v8 to classic v9, right?

Quetzacoalt91:
> You meant classic v8 to classic v9, right?

Exactly Krystian, I've changed the description accordingly. Thanks.

------
I've not provided much details about this PR, sorry about that. I've opened it for anyone curious who would like to have a look on the issue we're now facing and trying to solve.

When we update from PrestaShop v8 to v9, the gap is quite wide regarding the improvement we're brought on PrestaShop. We made a big step forward on the Symfony version, for instance we have PHP code with strict typing... and this leads to major changes in the modules as well.
With this version 9, we encounter a lot of Fatal Error from the modules that don't match the updated core functions signature. And the more modules you have on the shop, higher is the risk of encountering it. In the Update process, it seems we now require a moment where we avoid loading modules until we try to update them. This covers from the **end of files update**, **database**, to the **warm-up of modules update**.

I a perfect world, I should be able to ask the Core to not load modules. However until the feature in available, we have to find another even though it looks experimental.

kpodemski:
@Quetzacoalt91 

This has been discussed in other places of the project, I'm pretty sure of that, so maybe I will search for some old discussions where we shared, @Hlavtox also, numerous ways to handle it

I remember we discussed forcing developers to provide a maximum version of the PrestaShop compatible, so instead of:

`max_version => PS_VERSION`, they'd do something relevant

I also recall some ideas about sandboxing updates, where the updater attempts to load modules. If any fail, they are moved to a temporary directory.

Anyway, as mentioned above, what I do is that I change the module's name :) This does the trick, as long as there are no hardcoded lines like:

`require_once _PS_MODULE_DIR_ . 'mymodule/classes/SomeClass.php`

But I'm not sure if it's even possible to handle all the scenarios of badly written code I can imagine.

I'm happy that you guys took on this challenge to handle incompatible modules better 👏🏻

Quetzacoalt91:
> Anyway, as mentioned above, what I do is that I change the module's name :) This does the trick, as long as there are no hardcoded lines like

That is what we're doing with this PR, but on the database side. :)


You're suggesting interesting ideas as well. Regarding the `ps_version_compliancy` property, which is only checked at the module installation, we could check if the module has module updates. If nothing's available, we could disable it.
But this is another topic that would be covered with a whole new step pre-update, where we ask the merchant what to do for each module.

classes/UpgradeTools/Module/QuarantineZone.php
```diff
@@ -0,0 +1,65 @@
...
+    const DISABLED_BY_SAFE_MODE = 'UA|';
+    const ENABLED = 1;
+
+    /** @var Db */
+    private $db;
+
+    public function __construct(Db $db)
+    {
+        $this->db = $db;
+    }
```
M0rgan01: I think we can use DbWrapper class here. For better error handling, with exceptions

classes/Task/Update/UpdateModules.php
```diff
@@ -124,6 +125,9 @@ public function run(): int
             $this->next = TaskName::TASK_UPDATE_MODULES;
             $this->logger->info($this->translator->trans('%s modules left to check.', [$modules_left]));
         } else {
+            // Remove all remaining modules from the quarantine
+            $this->container->getQuarantineZone()->removeAll();
```
ga-devfront, line 129: Given that we remove them one by one, is it necessary to repeat this step at the end?

ga-devfront (approved):

classes/Task/Update/UpdateModules.php
Quetzacoalt91: Yes

Earlier in the process, we modify the whole `modules` table to rename all the modules, then we loop on the installed modules (In the database AND on the filesystem) by removing them from the safe zone and updating them.
If for some reason some some modules were in the database but not on the filesystem, they are still affected by this safe mode, hence the call to the `removeAll` to remove the alteration we applied on the rest of the data.

Quetzacoalt91:
I'm removing this pull-request from the expected release of Update Assistant 7.5.

After a few tests, it appears this pull-request improves the stability of updates, BUT...
It basically moves the errors from the incompatible modules AFTER the update. It brings blank pages on FO and BO and may require the debug mode to be enabled.
Then, to fix the issue with incompatible modules, the only solution you have is to manually delete the module from the `modules/` folder, preventing you to run the `uninstall()` method and cleaning everything from the module properly.

This PR must go with something that detects and warns upfront what module are incompatible and must be uninstalled.
https://github.com/PrestaShop/autoupgrade/pull/1528 is a first step.

AureRita (approved):
Hi @Quetzacoalt91 

https://github.com/user-attachments/assets/48468bfd-f15f-41e0-a292-645763c41132


Currently, the stability is correctly improved, we only have to uninstall ps_edition_basic on v1.0.21 before updating and we have to uninstal module that show error with debug mode 

Tested from 8.2.1 classic to 9.0.1

Thank you

classes/UpgradeTools/Module/QuarantineZone.php
```diff
@@ -0,0 +1,65 @@
...
+{
+    const DISABLED_BY_SAFE_MODE = 'UA_';
+
+    /**
+     * Enable the safe mode of PrestaShop by preventing the execution of modules
+     */
+    public function addAll(): void
+    {
+        DbWrapper::execute('UPDATE `' . _DB_PREFIX_ . 'module` m
+            SET `name` = CONCAT("' . self::DISABLED_BY_SAFE_MODE . '", `name`)'
```
jolelievre, line 44: Just a question, should this addAll only perform on module names that don't integrate the prefix yet?
I'm just wondering, in case addAll is called multiple times (form any reason) then the prefix will keep piling up and we could have `UA_UA_module_name`

Maybe even this is not a problem, it depends on the `removeAll` implementation, if the `REPLACE` mysqual method is able to replace all the matching strings then it xould still work I guess

So it's just to check if you anticipated this potential issue?

classes/Task/Update/UpdateDatabase.php
```diff
@@ -130,6 +129,15 @@ protected function warmUp(): int
...
         );
 
+        if ($this->container->getUpdateConfiguration()->shouldDeactivateCustomModules()) {
+            $this->logger->info($this->container->getTranslator()->trans('Disabling all non native modules'));
+            $this->getCoreUpgrader()->disableCustomModules();
+        } else {
+            $this->logger->info($this->container->getTranslator()->trans('Keeping non native modules enabled'));
+        }
+
+        $this->container->getQuarantineZone()->addAll();
```
jolelievre, line 139: So it means all the modules are quarantined, even the native one right?
Shouldn't there be an exception for the autoupgrade module?

jolelievre (commented):
Sorry very late review 😅
Of course not blocking anything, I just had a few questions

classes/UpgradeTools/Module/QuarantineZone.php
Quetzacoalt91: Good question, I wasn't sure about the duplicate from an error on our side, so, I made a quick check:

```
mysql> select * from ps_module where name like "%blockw%";
+-----------+---------------+--------+---------+
| id_module | name          | active | version |
+-----------+---------------+--------+---------+
|         3 | blockwishlist |      1 | 3.0.2   |
+-----------+---------------+--------+---------+
1 row in set (0.00 sec)

mysql> UPDATE `ps_module` m  SET `name` = CONCAT("UA_", `name`) WHERE id_module = 3;
Query OK, 1 row affected (0.03 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> select * from ps_module where name like "%blockw%";
+-----------+------------------+--------+---------+
| id_module | name             | active | version |
+-----------+------------------+--------+---------+
|         3 | UA_blockwishlist |      1 | 3.0.2   |
+-----------+------------------+--------+---------+
1 row in set (0.00 sec)

mysql> UPDATE `ps_module` m  SET `name` = CONCAT("UA_", `name`) WHERE id_module = 3;
Query OK, 1 row affected (0.01 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> select * from ps_module where name like "%blockw%";
+-----------+---------------------+--------+---------+
| id_module | name                | active | version |
+-----------+---------------------+--------+---------+
|         3 | UA_UA_blockwishlist |      1 | 3.0.2   |
+-----------+---------------------+--------+---------+
1 row in set (0.00 sec)

mysql> UPDATE `ps_module` m SET `name` = REPLACE(`name`, "UA_", "");
Query OK, 1 row affected (0.02 sec)
Rows matched: 67  Changed: 1  Warnings: 0

mysql> select * from ps_module where name like "%blockw%";
+-----------+---------------+--------+---------+
| id_module | name          | active | version |
+-----------+---------------+--------+---------+
|         3 | blockwishlist |      1 | 3.0.2   |
+-----------+---------------+--------+---------+
1 row in set (0.00 sec)
```

Looks like we're safe.

Note: I expect NO module to be already prefixed by `UA_`, because technical names of modules are always in lowercase.

classes/Task/Update/UpdateDatabase.php
Quetzacoalt91: Yeah no exception looks necessary here. If you're worried about the update being broken in the middle of the process because we quarantined it, there's actually no issue to fear because we rely on our own entrypoint.

classes/UpgradeTools/Module/QuarantineZone.php
jolelievre: Yep that was the only remaining worry I had, maybe you can use a weirder prefix like `PS_AUTOUPGRAGE_QUARANTINE_PREFIX_THAT_NOBODY_SHOULD_ALREADY_USE_DOGE_`

classes/Task/Update/UpdateDatabase.php
jolelievre: Ok and your own container as well? No possibilities that the services end up being unaccessible in the middle of the process?

Quetzacoalt91: Yes, all the components are redone inside the module, we do not use the service container of Symfony provided by the core because we are modifying it at the same time.

jolelievre (approved):

PR #1489, opened 2025-10-08
Display last output when missing the base command

Quetzacoalt91:
| Questions         | Answers
| Description?      | When encountering an issue with the installation of assets, I though this was caused by a `bin/console` that can't be called, but this actually caused by a module that was breaking the core.
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Delete bin/console during the update, we should expect the installation of assets to fail with some details.

Capture d’écran du 2025-10-08 17-09-48

### DIFF
--- a/classes/UpgradeTools/CoreConsoleExecutable.php
+++ b/classes/UpgradeTools/CoreConsoleExecutable.php
@@ -68,15 +68,16 @@ class CoreConsoleExecutable
         ];
 
         $returnCode = 0;
+        $output = [];
 
         foreach ($candidates as $candidate) {
-            exec($candidate, $null, $returnCode);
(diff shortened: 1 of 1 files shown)


kpodemski:
And it showed some of the errors right away in the CI :)

Quetzacoalt91:
> And it showed some of the errors right away in the CI :)

Actually this error in the CI happens at another moment and is caused by PS Checkout being incompatible with PHP 7.1. It has been reported in https://github.com/PrestaShopCorp/ps_checkout/discussions/1373

Quetzacoalt91:
Equivalent to https://github.com/PrestaShop/autoupgrade/pull/1490, merging.

PR #1504, opened 2025-10-20
[POC] Merge of PRs to test CI (Safe mode + kernel enabled for modules + Additional vendor for PHP 8.1)

Quetzacoalt91:
PR that merges the following ones to try out the CI if everything is merged altogether

List of PRs tested:
* https://github.com/PrestaShop/autoupgrade/pull/1502
* https://github.com/PrestaShop/autoupgrade/pull/1507
* https://github.com/PrestaShop/autoupgrade/pull/1512
* https://github.com/PrestaShop/autoupgrade/pull/1485
* https://github.com/PrestaShop/autoupgrade/pull/1516

PR #1502, opened 2025-10-16
Enable Symfony Kernel for module migrations

Quetzacoalt91:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Some modules need the Container in their migration scripts. This PR inits the Kernel before running migration to make sure modules have the necessary services when needed.
| Type?             | improvement
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| Sponsor company   | @PrestaShopCorp
| How to test?      | During the update of PrestaShop, update a module as well that rely on the Symfony service container. It should fail on the `dev` branch while this PR fixes it.

### With the pull-request:

```
[2025-10-16 19:08:32] NOTICE - ModuleDownloader - Module psshipping update files (1.1.4 => 2.0.5) have been fetched from /var/www/html/**admin_folder**/autoupgrade/modules/psshipping.zip.
[2025-10-16 19:08:32] DEBUG - ZipAction - Content of archive /var/www/html/**admin_folder**/autoupgrade/tmp/modules//psshipping.zip is extracted
[2025-10-16 19:08:32] NOTICE - ModuleMigration - (1/1) Applying migration file upgrade-2.0.0.php.
[2025-10-16 19:08:33] INFO - UpdateModules - All modules have been updated.
```

### `dev` branch:

```[2025-10-14 12:04:40] NOTICE - ModuleDownloader - Module psshipping update files (1.1.4 => 2.0.5) have been fetched from /var/www/html/**admin_folder**/autoupgrade/modules/psshipping.zip.
[2025-10-14 12:04:40] DEBUG - ZipAction - Content of archive /var/www/html/**admin_folder**/autoupgrade/tmp/modules//psshipping.zip is extracted
[2025-10-14 12:04:40] NOTICE - ModuleMigration - (1/1) Applying migration file upgrade-2.0.0.php.
[2025-10-14 12:04:41] CRITICAL - ErrorHandler - /var/www/html/modules/psshipping/src/Domain/Carriers/CarrierRepository.php line 78 - Error: Call to a member function createQueryBuilder() on null
#0 /var/www/html/modules/psshipping/src/Domain/Carriers/CarrierService.php(122): PrestaShop\Module\Psshipping\Domain\Carriers\CarrierRepository->getCarriers()
#1 /var/www/html/modules/psshipping/src/Domain/Carriers/CarrierService.php(89): PrestaShop\Module\Psshipping\Domain\Carriers\CarrierService->get()
[...]

PR #1512, opened 2025-10-23
Create additional autoloader for PHP 8.1+

Quetzacoalt91:
| Questions         | Answers
| Description?      | While attempting fixing others issues we have on the project, regarding modules updates or assets installation, we encounter conflicts on `symfony/console` when we load Symfony while updating in CLI. This is caused by the dependencies we have on our main `composer.json` that lack the strict typing we are supposed to have on PrestaShop v9.<br>To improve the situation, we create a variant of our dependencies that will be loaded first in CLI while running PHP 8.1+. This is hacky but this allows us to keep one module to maintain for PrestaShop 1.7 ~ 9.
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | CI is still green. Furthermore, this change should make the update processes to work again. Some the PRs currently under review fail, and must succeed with this one as we should find out in the test: https://github.com/PrestaShop/autoupgrade/pull/1504

### DIFF
--- a/sub-vendors/php8/composer.json
+++ b/sub-vendors/php8/composer.json
@@ -1,0 +1,26 @@
+{
+    "name": "prestashop/autoupgrade",
+    "description": "PrestaShop module autoupgrade (Variant of libraries for PHP 8)",
+    "homepage": "https://github.com/PrestaShop/autoupgrade",
+    "license": "AFL-3.0",
+    "authors": [
+        {
+            "name": "PrestaShop SA",
+            "email": "contact@prestashop.com"
+        }
+    ],
+    "require": {
+        "php": ">=8.1",
+        "symfony/console": "^6.4"
+    },
+    "config": {
+        "preferred-install": "dist",
+        "platform-check": false,
+        "classmap-authoritative": true,
+        "platform": {
+            "php": "8.1.0"
+        },
+        "prepend-autoloader": false,
+        "optimize-autoloader": true
+    }
+}
--- a/bin/console
+++ b/bin/console
@@ -41,6 +41,10 @@ if (!is_dir(dirname(__DIR__).'/vendor')) {
     throw new LogicException('Dependencies are missing. Try running "composer install".');
 }
 
+if (PHP_VERSION_ID >= 80100) {
+    // Autoloaders are appended. This one must be loaded first.
+    require_once dirname(__DIR__).'/sub-vendors/php8/vendor/autoload.php';
+}
 require_once dirname(__DIR__).'/vendor/autoload.php';
 
 if (!defined('_PS_ROOT_DIR_')) {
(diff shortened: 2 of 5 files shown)


ga-devfront (approved):

ga-devfront (approved):

ga-devfront (dismissed):

sub-vendors/php8/composer.json
```diff
@@ -0,0 +1,26 @@
...
+    "license": "AFL-3.0",
+    "authors": [
+        {
+            "name": "PrestaShop SA",
+            "email": "contact@prestashop.com"
+        }
+    ],
+    "require": {
+        "php": ">=8.1",
+        "symfony/console": "^6.4"
```
M0rgan01, line 14: We'll probably have some surprises with this if we add or edit CLI features. Some classes have likely changed between v3 and v6. I'm surprised it works

Quetzacoalt91: One of the two versions are actually called depending on the running version of PHP. A lot of code has changed, i.e to introduce the strict typing but the methods are actually the same.

About this issue, it is caused by two different versions of Symfony that leads to the error you've seen here: https://github.com/PrestaShop/autoupgrade/pull/1502#issuecomment-3486159353
There is no other valid solution to have a running CLI on the module. They either require PHP 7.4, or require `symfony/console` which leads to same issue.

So we have the choice between re-implementing the CLI by ourselves, using php-scoper or this workaround. I've made my choice.

M0rgan01: I understand that only one version is being called. For me, this solution is confusing; in addition to having two different Composer installations, it should be clearly stated that the v3 version of the dependency should be used, not the v6 version. 

I checked, and there are indeed some changes.

- https://github.com/symfony/symfony/blob/4.0/UPGRADE-4.0.md#console
- https://github.com/symfony/symfony/blob/5.0/UPGRADE-5.0.md#console
- https://github.com/symfony/symfony/blob/6.0/UPGRADE-6.0.md#console

A new version of the Update assistant module, compatible only with v8 -> v9, would probably have solved the problem.

Quetzacoalt91: Among these changelogs, what are the changes that could be impacting the module? 

Yes, a new version would have solved the issue as well but drastically increase the amount of work in the middle of active changes.

M0rgan01: For now, none; otherwise, there would already have been problems. 

As I said, the CLI development needs to focus on v3, not v6, because there will be potential problems related to these changes, in addition to the new features not present in v3.

A paragraph in the README explaining this sub-Composer  would help with understanding it.

M0rgan01 (approved):

AureRita (approved):
Hi @Quetzacoalt91 

Thank you for your PR, I tested it and it seems to works .

Tested from : 
1.7.8.9 to 8.2.3 to 9.0.1
8.2.2 classic to 8.2.3
8.2.2 classic to 9.0.1 with php 8.1
8.2.2 classic to 9.0.1 with php 8.3
8.0.3 to 8.2.3 
8.0.3 to 9.0.1 

Thank you

PR #1513, opened 2025-10-23
Fix update command - get channel version

M0rgan01:
| Questions         | Answers
| Description?      | we need to check that there is a realese before retrieving the version 
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | use update command without channel specified on v9.0.0

### DIFF
--- a/classes/Commands/UpdateCommand.php
+++ b/classes/Commands/UpdateCommand.php
@@ -101,8 +101,6 @@ class UpdateCommand extends AbstractCommand
                 if (!$updateState->isInitialized()) {
                     $updateState->initDefault($this->upgradeContainer->getProperty(UpgradeContainer::PS_VERSION), $this->upgradeContainer->getUpgrader(), $this->upgradeContainer->getUpdateConfiguration());
                 }
-
-                $updateConfiguration = $this->upgradeContainer->getUpdateConfiguration();
             }
 
             $this->calculateUpdateTypeAfterConfigLoad();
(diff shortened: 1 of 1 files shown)


classes/Commands/UpdateCommand.php
```diff
@@ -101,8 +101,6 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
                 if (!$updateState->isInitialized()) {
                     $updateState->initDefault($this->upgradeContainer->getProperty(UpgradeContainer::PS_VERSION), $this->upgradeContainer->getUpgrader(), $this->upgradeContainer->getUpdateConfiguration());
                 }
-
-                $updateConfiguration = $this->upgradeContainer->getUpdateConfiguration();
```
ga-devfront, line 105: I think we need to load this cause of the next command => 
```
$this->calculateUpdateTypeAfterConfigLoad();
```
if we don't do it I think we have an error

M0rgan01: Strange, this is the first thing we do in the method calculateUpdateTypeAfterConfigLoad https://github.com/PrestaShop/autoupgrade/blob/dev/classes/Commands/UpdateCommand.php#L192

M0rgan01: This method need a rename

ga-devfront (approved):

AureRita (approved):
Hi @M0rgan01 

image


Tested from : 
9.0.0 to 9.0.1

Currently the issue find on the video is not due to this PR. 

Thank you

PR #1518, opened 2025-11-03
Change available_date column from date to datetime

tleon:
…port datetime values in API

| Questions         | Answers
| Description?      | Change available_date column from date to datetime to support datetime values in API
| BC breaks?        | yes 
| Deprecations?     |  no
| Fixed ticket?     | Fixes #38787
| How to test?      | see https://github.com/PrestaShop/PrestaShop/pull/39904


Quetzacoalt91 (approved):

Hlavtox (changes requested):
I am not sure about this. The field is a date, not date and time. It was a date since forever.

If it's broken in the API, it should be fixed there no?

Quetzacoalt91:
I understand your concern @Hlavtox, but this should be discussed on the original PR as the one on Update Assistant only reflects the changes merged on PrestaShop.

PR #1535, opened 2025-11-21
Persist current product route when upgrading to 9.0.0

### LINKED ISSUE #40063: PrestaShop 9: Legacy Category/Product URLs Return 404 With New Default Routing (No Redirects or Fallback)
### Prerequisites

**Intro**

While testing an upgrade from a legacy PrestaShop installation to PrestaShop 9, we encountered a situation we wanted to bring to the PrestaShop developers’ attention.

Note: In our upgrade test, we installed PrestaShop 9.0.1 and migrated legacy shop content without making any changes to SEO & URLs settings. The behavior described below occurs with the default PS9 configuration.

This concerns how PS9 handles product URLs when upgrading from stores that previously used category-based product link structures.

**Summary**

In PrestaShop 9, the default product route removes the {category} token:

/{id}-{rewrite}.html

This new “clean” structure works fine for fresh installs.
However, when upgrading a legacy shop, PS9 does not provide redirects, rewrites, or fallback resolution for older product URLs that include the category segment:

/my-category/{id}-{rewrite}.html

Therefore, any external or indexed legacy URL containing a category will return a 404, unless the merchant manually restores a category-aware route or implements redirects.

### Expected behavior

During upgrade acceptance testing we noticed that /category/product links were 404'ing and after some research found the ps9.0.1 default product route removed category.  We readded the category to product route and external links functioned.  
When I heard of the new feature I assumed/expected that links with category would resolve.  

Thanks in advance for ps team review and hope this contributes in some small way. 

### Steps to reproduce

Steps to Reproduce

1. Install PrestaShop 9.0.1 (fresh installation or upgrade).
2. Leave SEO & URLs at the default PS9 configuration (no category)
3. Create a product or use any demo product included with a fresh install.
4. Assign the product to a category (any category, including the default “Home” or a newly created category).
5. Confirm the new short URL works:

/{id}-{rewrite}.html

6. Now manually construct and request the category-based legacy pattern (even on a fresh PS9 install), for example:

/my-category/{id}-{rewrite}.html

or using the real category slug:

/accessories/12-mountain-bike-helmet.html

7. Observe that the page returns 404 Not Found.
8. No redirect or rewrite is generated by default.
9. Breadcrumb still displays the category hierarchy, even though the URL itself does not include the category segment.

9.0.1 from 1.6.1.4

download 9.0.1 and installed on staging and used migration pro to import shop datas

8.2.1

_No response_

PrestaHeroes.com

Hlavtox:
| Questions         | Answers
| Description?      | If routes have not been modified, they are not in ps_configration. If you upgrade to 9.0.0, your product URLs change and you completely RUIN YOUR SEO. Caused by https://github.com/PrestaShop/PrestaShop/pull/37467.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/40063#issuecomment-3564109485
| How to test?      | Upgrade from 8.2.3 to 9.0.0 and check that your product rewrite rule in SEO settings is the same.

### DIFF
--- a/upgrade/php/add_product_routes.php
+++ b/upgrade/php/add_product_routes.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+/**
+ * In Prestashop 9.0.0 the default product routes have been modified by removing
+ * category and EAN from the URL.
+ *
+ * If the merchant kept the original routes the former urls won't be reachable any
+ * more and SEO will be lost. So we force a custom rule matching the former format.
+ *
+ * If the route was customized, no need to do anything. We don't change anything for
+ * multi shop either since it will be used it the merchant has already changed them.
+ */
+function add_product_routes()
+{
+    Configuration::loadConfiguration();
+
+    if (!Configuration::get('PS_ROUTE_product_rule', null, 0, 0)) {
+        Configuration::updateGlobalValue('PS_ROUTE_product_rule', '{category:/}{id}{-:id_product_attribute}-{rewrite}{-:ean13}.html');
+    }
+}
--- a/upgrade/php/ps_900_set_previous_product_route_as_custom.php
+++ b/upgrade/php/ps_900_set_previous_product_route_as_custom.php
@@ -28,10 +28,8 @@
  * If the route was customized, no need to do anything. We don't change anything for
  * multi shop either since it will be used it the merchant has already changed them.
  */
-function add_product_routes()
+function ps_900_set_previous_product_route_as_custom()
 {
-    Configuration::loadConfiguration();
-
     if (!Configuration::get('PS_ROUTE_product_rule', null, 0, 0)) {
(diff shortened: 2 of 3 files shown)


M0rgan01 (dismissed):

upgrade/php/add_product_routes.php
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * If the merchant kept the original routes the former urls won't be reachable any
+ * more and SEO will be lost. So we force a custom rule matching the former format.
+ *
+ * If the route was customized, no need to do anything. We don't change anything for
+ * multi shop either since it will be used it the merchant has already changed them.
+ */
+function add_product_routes()
+{
+    Configuration::loadConfiguration();
```
Quetzacoalt91: Why did you need this line?

Hlavtox: I got inspired here - https://github.com/PrestaShop/autoupgrade/blob/dev/upgrade/php/add_supplier_manufacturer_routes.php

Honestly, I did not think about it too much, maybe it's not needed indeed

Quetzacoalt91: I'll make some tests then. We need to create a catch-up file as well for merchants who already updated to 9.0.0

Hlavtox: @Quetzacoalt91 I would not make a catchup. If we already broke their SEO rules when upgrading to 9.0.0, they probably fixed it, or Google began to reindex it. If we change it again now, it would be even worse.

Quetzacoalt91: Fair enough, I see another issue with a catch-up file, where someone who installed PS 9.0.0 would potentially see its URLs changing while updating to 9.0.2

upgrade/php/add_product_routes.php
```diff
@@ -0,0 +1,38 @@
...
+ * In Prestashop 9.0.0 the default product routes have been modified by removing
+ * category and EAN from the URL.
+ *
+ * If the merchant kept the original routes the former urls won't be reachable any
+ * more and SEO will be lost. So we force a custom rule matching the former format.
+ *
+ * If the route was customized, no need to do anything. We don't change anything for
+ * multi shop either since it will be used it the merchant has already changed them.
+ */
+function add_product_routes()
```
Quetzacoalt91: Can we rename the method and the file name ?

```suggestion
function ps_900_set_previous_product_route_as_custom()
```

Hlavtox:
@Quetzacoalt91 Should be ready. :-)

Quetzacoalt91 (approved):

AureRita (approved):
Hi @Hlavtox 

https://github.com/user-attachments/assets/b612bd83-82ee-4b2d-8a44-af9a72555f06


Tested from : 
8.2.3 to 9.0.1
8.0.4 to 9.0.1

Thank you

PR #1471, opened 2025-09-22
Fix upgrade to 9.0.2: increase customer_message.user_agent field length from 128 to 255

yebor974:
Questions | Answers
-- | --
Description? | This PR updates the auto-upgrade SQL script for 9.0.1 to increase the customer_message.user_agentfield size from 128 to 255 characters. This prevents customer messages from being silently rejected when the user agent string (especially on some mobile devices such as iPhone Safari) exceeds 128 chars.
Type? | bug fix
BC breaks? | no
Deprecations? | no
Fixed ticket? | Related to [PrestaShop/PrestaShop#39606](https://github.com/PrestaShop/PrestaShop/pull/39606)
Sponsor company | N/A
How to test? | 1. Upgrade a shop to 9.0.1 using this module. 2. Check the ps_customer_message table: the column user_agent should now have a length of 255. 3. Try posting a customer message with a long mobile user agent (e.g. Safari iPhone UA ~135 chars). 4. The message should now be saved correctly.

PR #1540, opened 2025-11-27
Fix column wrongly declared as NOT NULL instead of DEFAULT NULL

Quetzacoalt91:
| Questions         | Answers
| Description?      | Fix an error in the column type from https://github.com/PrestaShop/autoupgrade/pull/1471
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | When describing the table `customer_message` after an update to PS 9.0.2, the type of the column must be `user_agent varchar(255) DEFAULT NULL`

### DIFF
--- a/upgrade/sql/9.0.2.sql
+++ b/upgrade/sql/9.0.2.sql
@@ -3,7 +3,7 @@ SET NAMES 'utf8mb4';
 
 -- https://github.com/PrestaShop/PrestaShop/pull/39606
 ALTER TABLE `PREFIX_customer_message`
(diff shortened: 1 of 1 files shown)


ga-devfront (approved):

kseghair (approved):

AureRita (approved):
Hi @Quetzacoalt91 

image


Tested from : 
9.0.1 to 9.0.2
8.2 to 9.0.2
8.0.4 to 9.0.2

Thank you

PR #1528, opened 2025-11-14
Create "update:check-modules" command

M0rgan01:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Added a command to check which modules will be updated, which become incompatible, or which do not change version during the update process.<br/><br/>The channel can be specified (ONLINE, ONLINE_RECOMMENDED, LOCAL). An example of the output is available below. **This data uses only marketplace data**.<br/><br/>This order does not modify anything on the store; it is present for informational purposes only.
| Type?             |improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | use new `update:check-modules` command

image

PR #1547, opened 2025-12-12
Allow a module to be updated even if it does not have a config.xml file

Quetzacoalt91:
| Questions         | Answers
| Description?      | When running https://github.com/PrestaShop/autoupgrade/pull/1528 for tests, I discovered that some modules were not visible even though they are installed on the shop. This happens when the cache file `config.xml` does not exist. Because we don't rely anymore on the XML content the related check can be removed, allowing more modules to be updated.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Delete the config.xml file of specific modules. They must be updated even if the file is missing for their repository.


Hlavtox (dismissed):
This is good. The config.xml file is only required for addons. Shops will create config_isocodeoflanguage.xml automatically. :-)

Nakahiru (dismissed):

ga-devfront (dismissed):

ga-devfront (approved):

AureRita (approved):
Hi @Quetzacoalt91 

Capture d’écran du 2026-01-12 15-38-11

https://github.com/user-attachments/assets/e0f9fe8b-8550-4f1c-bb7a-fd09316d1838



Tested from : 
8.1.4 to 8.2.3 (each version on Open source and with module ps_mbo) 
9.0.0 to 9.0.2 

Thank you

PR #1580, opened 2025-12-18
Initiate new discount types table in 9.1.0

boherm:
| Questions         | Answers
| Description?      | Initiate new discount types table in 9.1.0
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | ~
| How to test?      | Upgrade into 9.1.0 version

### DIFF
--- a/upgrade/php/ps_910_init_cart_rule_type_lang_translations.php
+++ b/upgrade/php/ps_910_init_cart_rule_type_lang_translations.php
@@ -1,0 +1,85 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+use PrestaShop\Module\AutoUpgrade\Database\DbWrapper;
+
+/**
+ * Updates ps_cart_rule_type_lang table to add translations
+ *
+ * This method will update ps_cart_rule_type_lang with translated wordings for all available languages
+ *
+ * @throws \PrestaShop\Module\AutoUpgrade\Exceptions\UpdateDatabaseException
+ */
+function ps_910_init_cart_rule_type_lang_translations()
+{
+    $translator = Context::getContext()->getTranslator();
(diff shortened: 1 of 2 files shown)


jolelievre (dismissed):
Thanks @boherm

jolelievre (dismissed):

Hlavtox (dismissed):
Two things:
1) These translations won't be discoverable, are you sure they will be scannned?

2) I see cart rule compatibility that only resolves compatibility types, this is not enough and it was promised that this will be resolved no?

tleon (dismissed):

boherm:
Hi @Hlavtox,

1. We have create a `CartRuleTypeLang.php` in `classes/lang/KeysReference` (like we do for others things) to discover some translations. (as you see [here](https://github.com/PrestaShop/PrestaShop/pull/40354/changes#diff-c46e5b8f8222f2c720d1dffdc4210fa19813ea79b05b341c7c9b55bac0f0013fR909), the translation tool can discover this new strings)

2. I can't answer to this, but maybe @ibahloul-ps or @jolelievre could 😜

jolelievre (dismissed):

Hlavtox:
@jolelievre Yeah, but by pushing this to 9.1 you create a BC break contract that you cannot take back.

> Maybe it's not enough, but that's what the experimental phase is made for and that's why we need the comunity to test this new discount feature

So you are implementing something, completely breaking up the discount system and you don't know if it's gonna work? Is this for real?

Quetzacoalt91:
@boherm what's the related PR on PrestaShop/PrestaShop repository?

jolelievre:
@Hlavtox stop the drama! This is fully retro compatible with the current system
How is it a breaking change?

Hlavtox:
@jolelievre It's not, in the current system, you define cart rule compatibility with specific cart rules.

In your new "improved" system (a.k.a we removed some functionality), it's only defined by cart rule types.

If you push it now, you cannot take it back, if you don't want to piss developers who already started preparing solutions and modules for it.

And then you openly admit: `Maybe it’s not enough, but that’s what the experimental phase is for and why we need the community to test this new discount feature`. I thought there was a huge amount of exploration behind this with highly esteemed French merchants?

What kind of design is this, where discount compatibility is defined only by the rule type? So a discount is either incompatible with everything, or compatible with everything? What kind of nonsense is that?

jolelievre:
The new system is behind a feature flag, it doesn't prevent the current system from working as long as you don't enable the new feature The purpose is indeed to test things and see if it meets the needs, if it doesn't the feature can still evolve And if we realize the cart rule types are a bad iea we can still remove them from the DB, since they are related to an experimental feature it's still possible without a BC

jolelievre (dismissed):

mattgoud (dismissed):

mattgoud (approved):

jolelievre (approved):

AureRita (approved):
Hi @boherm 

https://github.com/user-attachments/assets/46f13c8f-dabc-489f-8ebb-a48532fd3124



Tested from : 
8.1.4 to 9.1.0 
9.0.2 to 9.1.0 

Waiting for CLI green to remove the waiting QA 

Thank you

boherm:
Hi @Quetzacoalt91,
As asked, this PR is made for this three PRs about new discount system:
- https://github.com/PrestaShop/PrestaShop/pull/39662
- https://github.com/PrestaShop/PrestaShop/pull/40293
- https://github.com/PrestaShop/PrestaShop/pull/38842

PR #1590, opened 2026-01-07
Change how we get the Request object

### LINKED ISSUE #1578: [BUG] The file "/tmp/phpnWYRUf" does not exist
### Prerequisites

- [x] I have already [searched in existing issues](./issues?q=is%3Aissue+type%3ABug) and found no previous report of this bug.

When I replace a picture in the menu module of my template the shop crashes. The error indicates that it happens in a hooked module. When I add some debug code (Prestashop is horrible with errors in hooked code) I see that the causing module is autoupgrade and the hook is DisplayBackofficeHeader.

The filename is every time different but it always starts with /tmp/php

I upgraded autoupgrade to the latest version but it made no difference.

Image

### Expected behavior

No crash

### Steps to reproduce

I used pm_advancedtopmenu v1.15.2 - by Presta-Module. But as the error occurs in Autoupgrade it is likely to happen elsewhere too.

8.2.3

8.2

_No response_

### LINKED ISSUE #1586: Symfony Request::createFromGlobals() in Autoupgrade hook 'displayBackOfficeHeader' breaks other module file upload in PS 9 (tmp file deleted before hook execution)
### Prerequisites

### This is for the dev team

When saving a module configuration form that includes a file upload, PrestaShop 9.0.2 throws a RuntimeError stating that the uploaded temporary file does not exist. This happens even though the module correctly moves the uploaded file to a permanent location.

Image

After investigation, the root cause is not in the module performing the upload, but in the Autoupgrade module’s `displayBackOfficeHeader` hook.

During the hook execution `hookDisplayBackOfficeHeader`, Autoupgrade calls `DisplayBackOfficeHeader::renderUpdateNotification` which executes `$request = Request::createFromGlobals();`. This triggers Symfony’s internal `Request::initialize()` logic, which rebuilds the FileBag from the current global $_FILES. This can be seen in `Request::initialize` method in the statement `$this->files = new FileBag($files);`

At this point, $_FILES still contains the original tmp_name of the uploaded file, but the file has already been moved by the module and no longer exists in `/tmp`. Symfony performs strict validation and throws an exception because the temporary file referenced in $_FILES is gone.

I have tested this with the following setup:

- PrestaShop v9.0.2
- Update Assistant v7.4.4
- The module that has a file upload in the configuration page: https://addons.prestashop.com/en/product-visuals-media/88117-product-images-swiper-w-video-integration-lightbox.html

However, the issue can be reproduced with any module that:

- Provides a configuration form in the back office
- Accepts file uploads
- Moves the uploaded file using `move_uploaded_file()`
- Does not redirect after POST (to preserve success messages)

### Workaround
In the module that uploads a file, after moving the uploaded file, I had to unset the $_FILES[KEY] `unset($_FILES[KEY]);`.
This prevents Symfony from revalidating a non-existing tmp_name during the Autoupgrade hook (as the `$files` will be empty in `$this->files = new FileBag($files);`).
However, this is only a workaround. The potential underlying issue remains in the Autoupgrade module's use of `Request::createFromGlobals()` inside a hook to update the notification display of a module that is not currently edited by the user.

**I cannot confirm whether this behavior is directly caused by the Symfony upgrade, but I would like to raise a broader question: Why does PrestaShop execute hooks from other modules while saving the configuration page of a specific module? I guess this is the architecture of PrestaShop.
In my humble opinion, when I am editing the configuration of one module, I would expect only the hooks registered by that module to be executed. In this case, PrestaShop executes the `displayBackOfficeHeader` hook of the Autoupgrade module in order to render its update notification, even though the Autoupgrade configuration page is not open.
From a module‑developer perspective, this behavior is unexpected and leads to side effects such as the one described above.**

### Expected behavior

- Saving a module configuration form with a file upload should not trigger exceptions.
- Hook execution should not re-parse $_FILES after the module has already processed the upload.
- Or avoid running hooks from other modules. Only run the hooks of the module that has its configuration page open for modification.

### Steps to reproduce

1. Make sure you have Update Assistant v7.4.4 already installed.
2. Install a module with a configuration form containing a file upload field.
3. Upload an image and save the form.
4. PrestaShop processes the form and moves the uploaded file to a permanent directory.
5. Immediately after saving, PrestaShop executes BO hooks, including `displayBackOfficeHeader` (Autoupgrade), which calls `Request::createFromGlobals()`.
6. Symfony rebuilds the FileBag using $_FILES, which still contains the now‑invalid tmp_name.
7. Symfony throws: RuntimeError: The file /tmp/phpXXXXXX does not exist

9.0.2

_No response_

8.4

autoupgrade v7.4.4

_No response_

### LINKED ISSUE #39239: Module Autoupgrade cause FileNotFoundException on fileupload
### Prerequisites

When you try to upload a file in module Everblog you get an error :
PHP Fatal error: Uncaught Symfony\\Component\\HttpFoundation\\File\\Exception\\FileNotFoundException: The file "/tmp/phpPOc2rI" does not exist

### Expected behavior

File upload run ok without error.

### Steps to reproduce

Install Everpsblog module (github.com/TeamEver/everpsblog/releases) and Autoupgrade v 7.2.
Create a new blog article.
Edit blog article and add or change the image.

1.7.8.11

Github repo

7.4

everpsblog 5.6 or 6.0

Netenvie

Quetzacoalt91:
| Questions         | Answers
| Description?      | Update Assistant breaks the BO when a module is uploaded and handled in the request. This PR tends to use the existing request from Symfony when possible. There are some limitations though:<ul><li>It appears the service "request_stack" may be non-existing on PS 1.7 because of the limited integration of Symfony on these versions, especially on legacy pages</li><li>The request stack may have no requests in their list.</li></ul>In these cases, we still rely on the use of `Request::createFromGlobals()` with the risk of failures on pages where a file is uploaded. The issue is not completely fixed but mitigated.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/autoupgrade/issues/1586, Fixes https://github.com/PrestaShop/autoupgrade/issues/1578, Fixes https://github.com/PrestaShop/PrestaShop/issues/39239
| How to test?      | Use the module [test_module_40410.zip](https://github.com/user-attachments/files/24477569/test_module_40410.zip), configure it with Update Assistant installed. On the configuration of this test module, select a file and upload it, you'll get the exception about a non-existing file that is fixed by this PR.

### DIFF
--- a/autoupgrade.php
+++ b/autoupgrade.php
@@ -204,7 +204,10 @@ class Autoupgrade extends Module
             return '';
         }
 
-        return (new \PrestaShop\Module\AutoUpgrade\Hooks\DisplayBackOfficeHeader($this->getUpgradeContainer()))->renderUpdateNotification();
+        return (new \PrestaShop\Module\AutoUpgrade\Hooks\DisplayBackOfficeHeader(
+            $this->getUpgradeContainer(),
+            $this->getCurrentRequest()
+        ))->renderUpdateNotification();
     }
 
     /**
@@ -267,5 +270,27 @@ class Autoupgrade extends Module
         }
 
         return $this->container;
+    }
+
+    /**
+     * @return \Symfony\Component\HttpFoundation\Request
+     */
+    public function getCurrentRequest()
+    {
+        // When possible, retrieve an existing instance of Request to avoid issues on pages with uploaded files.
+        // We are compatible with PS 1.7.0 BUT:
+        //  - Module::get() exists since PS 1.7.3.
+        //  - Service "request_stack" is found from PS 8.
+        if (version_compare(_PS_VERSION_, '8.0.0', '>=')) {
+            /** @var \Symfony\Component\HttpFoundation\RequestStack $requestStack */
+            $requestStack = $this->get('request_stack');
+            $request = $requestStack->getCurrentRequest();
+
+            if ($request) {
+                return $request;
+            }
+        }
+
+        return \Symfony\Component\HttpFoundation\Request::createFromGlobals();
     }
 }
--- a/classes/Hooks/DisplayBackOfficeHeader.php
+++ b/classes/Hooks/DisplayBackOfficeHeader.php
@@ -56,6 +56,11 @@ class DisplayBackOfficeHeader
     private $container;
 
     /**
+     * @var Request
+     */
+    private $request;
+
+    /**
      * @var Upgrader
      */
     private $upgrader;
@@ -93,9 +98,10 @@ class DisplayBackOfficeHeader
     /**
      * @throws Exception
      */
-    public function __construct(UpgradeContainer $container)
+    public function __construct(UpgradeContainer $container, Request $request)
     {
         $this->container = $container;
+        $this->request = $request;
         $this->upgrader = $this->container->getUpgrader();
         $this->updateNotificationService = $this->container->getUpdateNotificationService();
         $this->updateNotificationConfiguration = $this->updateNotificationService->getUpdateNotificationConfiguration();
@@ -129,8 +135,7 @@ class DisplayBackOfficeHeader
 
         $this->addScriptsVariables();
 
-        $request = Request::createFromGlobals();
-        $this->addUIAssets($request);
+        $this->addUIAssets($this->request);
 
         if (!$this->isEmployeeDefaultController()) {
             return $this->content;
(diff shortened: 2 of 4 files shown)


ga-devfront (approved):

AureRita (approved):
Hi @

https://github.com/user-attachments/assets/11520972-8c7b-41ca-b2c6-3eb28f1a81da



Tested on : 
9.0.2

Thank you

ingridusta:
FYI following v7.6 beta release tests, it's not working on 8.2.3 neither (but works on 9.0.3 😉).

PR #1599, opened 2026-01-15
Speed up execution of files updates by moving files instead of copying them

Quetzacoalt91:
| Questions         | Answers
| Description?      | During the update of files, move them instead of copying them which is a really fast operation when the source and destination are on the same disk partition.
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Check the time needed to run the updates, especially until the end of the update of files. You should expect the time taken to be cut in half.

### DIFF
--- a/classes/Task/Update/UpdateFiles.php
+++ b/classes/Task/Update/UpdateFiles.php
@@ -182,7 +182,7 @@ class UpdateFiles extends AbstractTask
             // upgrade exception were above. This part now process all files that have to be upgraded (means to modify or to remove)
             // delete before updating (and this will also remove deprecated files)
             try {
-                $this->container->getFileSystem()->copy($orig, $dest, true);
+                $this->container->getFileSystem()->rename($orig, $dest, true);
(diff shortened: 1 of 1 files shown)


Hlavtox (approved):
Should work good, but watch out for permissions. Rename needs write+execute rights for both `orig` and `dest` directories. It should work just fine, but could be more sensitive.

Quetzacoalt91:
It should just work indeed as we need to change the `dest` anyway and Update Assistant is the one who brought the `orig` files as well by unzipping the release.

I had another concern which the probability to use the same file a second time, but that's not the case. :+1:

M0rgan01 (approved):

AureRita (approved):
Hi @

From 9.0.0 classic to 9.0.2 on localhost
Before PR : 1m30
After PR : 1m20 

8.1.4 to 9.0.2 on localhost
before PR : 1min58
After PR : 1min50 

8.2.3 to 9.0.2 on docker : 
before PR : 3min30
After PR : 2min02 

Thank you

PR #1602, opened 2026-01-20
Let the file_get_contents method throw warnings about the issues they encountered

Quetzacoalt91:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Some issues may be occurring while the project is running, but the root cause is not seen because of a silencing `@` symbol before the method call.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/autoupgrade/issues/1562
| Sponsor company   | @PrestaShopCorp
| How to test?      | Modify your PHP config to add this key: allow_url_fopen=0, then start the update configuration process. You'll see the warning appearing now.

NOTE: When https://github.com/PrestaShop/autoupgrade/pull/1600 will be merged, the warning will be visible in the error report modal to make self debugging easier.

PR #1608, opened 2026-01-21
Add fallback to curl when allow_url_open is disabled; error if curl not available.

### LINKED ISSUE #1562: HTTP 500 on update Assistant when allow_url_fopen is off
### Prerequisites

We get a 500 error after trying to update the Prestashop core code. Even with development mode on it doesn't show any other errors, not in log either.

Our initial issue why we want to upgrade the core code is that in the backend several links seem to redirect weirdly (e.g. performance just redirects us to the login page) due to "?"s in the URL being replaced by "&"s instead. We hope the core upgrade will fix this.

### Expected behavior

We want to update our Prestashop core to the latest version

### Steps to reproduce

Go to https://www.modestoffenonline.nl/xxxxxxxxxxxxx/index.php?controller=AdminSelfUpgrade&route=home-page 
Click on "Update shop" 
See 500 page 

8.1.1

_No response_

8.3.26

Update Assistant 7.4.3.

V-Web / Modestoffenonine

instancezero:
| Questions         | Answers
| Description?      | Fallback to curl when retrieving update list; detect when curl is also missing.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes #1562 
| Sponsor company   | Abivia Inc. @abivia
| How to test?      | Run update assistant when allow_url_fopen is off, verify update list is retrieved (I have tested this on a live system but haven't rebuilt a PHP with no curl, just handling the edge case here.)


Quetzacoalt91:
Thanks @instancezero, would it be possible to export this code in a dedicated method?
There is at least two occurrences from where we could enjoy the change: https://github.com/PrestaShop/autoupgrade/pull/1602/changes

The change could be in `classes/Services/DownloadService.php` next to the existing download() method.

instancezero:
> Thanks @instancezero, would it be possible to export this code in a dedicated method? There is at least two occurrences from where we could enjoy the change: https://github.com/PrestaShop/autoupgrade/pull/1602/changes
> 
> The change could be in `classes/Services/DownloadService.php` next to the existing download() method.

It is possible. Can you suggest a method name? 

Another option would be to make the `$destinationPath` argument nullable, change the return type to `?string` (and then `string|void` when minimum PHP is raised to 8.3) and have the `download()` method return the string when no destination is provided.

Quetzacoalt91:
> It is possible. Can you suggest a method name?
> 
> Another option would be to make the `$destinationPath` argument nullable, change the return type to `?string` (and then `string|void` when minimum PHP is raised to 8.3) and have the `download()` method return the string when no destination is provided.

We can use something like "fetchContents" or "getContents" to distinguish them from the method `download()` that stores on the filesystem.
I wouldn't modify the existing method, because it handle file contents as streams, alowing us to download the module to download huge amount of data without using too much memory and in consequence not reaching the memory limit exception.

instancezero:
I thought about this some more and came to the same conclusion about modifying `download()`.

I also neglected to mention that the latest commit implements `fetch()` as a static method. Instantiating the class would require passing a `Translator` and a `Logger`, and the translator requires a base path, all of which seems like unnecessary overhead to me.

classes/Services/DownloadService.php
```diff
@@ -77,6 +77,28 @@ public function download(string $downloadUrl, string $destinationPath): void
...
+
+    /**
+     * Return the contents of a URL as a string.
+     * @param string $downloadUrl
+     * @return string|null Null is returned if fopen() can't open a URL and CURL isn't present.
+     */
+    static public function fetch(string $downloadUrl): mixed
+    {
+        if (ini_get('allow_url_fopen')) {
+            $response = @file_get_contents($downloadUrl);
```
Quetzacoalt91, line 94:
```suggestion
            $response = file_get_contents($downloadUrl);
```

Let's make the warning clear for the user, as done in https://github.com/PrestaShop/autoupgrade/pull/1602

instancezero: > Let's make the warning clear for the user, as done in #1602

Actually, this goes back to why I was looking at this in the first place. In a production system, warnings will probably not be reported, and in the case of a service call, all the admin sees is a 500 error; diagnostic information (if any) can only be found by inspecting the JSON response with dev tools, hardly something we would expect from a store manager.

Also returning null is too old skool (some bad habits die hard). Also, the last patch was bad, these methods fail with `false`, not `null`.

I have revised this so that `fetch()` throws an `IOException` so that callers can at least generate a message with some semantic context. `DistributionApiService::getApiEndpoint()` catches the error and throws a `DistributionApiException`.

Lastly, I have suffered from "quick fix" disease, failed to follow my own rules and didn't write tests for the previous commit. There are now new PHPUnit tests that cover the changes.

Quetzacoalt91: Thank you, I appreciate the time taken for bringing the additional changes and the tests.

> diagnostic information (if any) can only be found by inspecting the JSON response with dev tools, hardly something we would expect from a store manager.

We are adding in Update Assistant 7.6 the warning in the JSON responses when needed, so they can be visible in the error report modal for easier debugging.

We can leave it for now and will see if we get reports about this method later.

classes/Services/DistributionApiService.php
```diff
@@ -64,16 +66,21 @@ public function __construct(Translator $translator)
      */
     public function getApiEndpoint(string $endPoint)
     {
-        $response = @file_get_contents(self::API_URL . '/' . $endPoint);
-
-        if (!$response) {
-            throw new DistributionApiException($this->translator->trans('Error when retrieving data from Distribution API'), DistributionApiException::API_NOT_CALLABLE_CODE);
+        try {
+            $downloader = new DownloadService($this->translator, new WebLogger());
```
Quetzacoalt91: Can you please move the `DownloadService` property as a constructor parameter that will be sent by the Container?

This requires another change in `UpgradeContainer::getDistributionApiService()` where the class is instanciated with

```php
new DistributionApiService($this->getTranslator(), $this->getgetDownloadService());
```

instancezero: > Can you please move the `DownloadService` property as a constructor parameter that will be sent by the Container?

Done.

Quetzacoalt91:
Excellent, if you can rebase your branch to solve the conflict, we should be good to go

instancezero:
That should do it...

Quetzacoalt91 (dismissed):
Thank you very much.

AureRita:
Hi @instancezero 

Thank you for your PR, unfortunately, this one is still in conflict. Even without this conflict, I was able to test the PR and we get a 500 error even when we have allow_url_fopen = On, as you can see : 


https://github.com/user-attachments/assets/bc70a2b6-29cb-4261-a5ec-50a683365750

You can see that all CI is red because of this issue 

Waiting for feedback

instancezero:
I'll take a look. The issue is the change to the constructor, which is breaking other parts of the code.

Quetzacoalt91 (dismissed):

instancezero:
Is there anything else required for this patch?

Quetzacoalt91 (approved):

instancezero:
Okay, that's it. I'm throwing in the towel on this. Suggest requiring fopen_url be enabled instead.

So long and thanks for all the fish.

Quetzacoalt91:
@instancezero Do you want me to have one last look?

instancezero:
If you can make use of it, please go ahead. It's failing on browser tests where there's not enough telemetry to figure out WTF is going wrong without duplicating the test environment. This has gone from a "hey guys I found the issue" (which I solved by changing my config) to a six week long process that started with a patch and grew, and now it has sucked up all the time I have available for it and more.

PR #1646, opened 2026-02-09
Add uninstall module task

ga-devfront:
| Questions         | Answers
| Description?      | Add uninstall module task to prevent error on update
| BC breaks?        | maybe
| Deprecations?     | no
| Fixed ticket?     | https://github.com/PrestaShop/autoupgrade/discussions/1631
| How to test?      | Check if uninstalled modules are really not compatible with destination version.

Results of update:check-modules command:
Results of update:check-modules command

### DIFF
--- a/classes/Models/Module/Marketplace/ModuleUpgradeCompatibility.php
+++ b/classes/Models/Module/Marketplace/ModuleUpgradeCompatibility.php
@@ -36,10 +36,10 @@ class ModuleUpgradeCompatibility
     private $latestRelease; // ou un objet Version
 
     public function __construct(
-        bool     $isCompatible,
-        bool     $hasUpdateAvailable,
+        bool $isCompatible,
+        bool $hasUpdateAvailable,
         ?Release $compatibleRelease,
-        Release  $latestRelease
+        Release $latestRelease
     ) {
         $this->isCompatible = $isCompatible;
         $this->hasUpdateAvailable = $hasUpdateAvailable;
--- a/classes/Models/Module/Marketplace/ModuleUpgradeCompatibility.php
+++ b/classes/Models/Module/Marketplace/ModuleUpgradeCompatibility.php
@@ -29,22 +29,22 @@ class ModuleUpgradeCompatibility
     /** @var bool */
     private $hasUpdateAvailable;
 
-    /** @var Release|null */
-    private $compatibleRelease;
-
     /** @var Release */
     private $latestRelease;
+
+    /** @var Release|null */
+    private $compatibleRelease;
 
     public function __construct(
         bool $isCompatible,
         bool $hasUpdateAvailable,
-        ?Release $compatibleRelease,
-        Release $latestRelease
+        Release $latestRelease,
+        ?Release $compatibleRelease = null
     ) {
         $this->isCompatible = $isCompatible;
         $this->hasUpdateAvailable = $hasUpdateAvailable;
-        $this->compatibleRelease = $compatibleRelease;
         $this->latestRelease = $latestRelease;
+        $this->compatibleRelease = $compatibleRelease;
     }
 
     public function isCompatible(): bool
--- a/classes/Models/Module/Marketplace/ModuleUpgradeCompatibility.php
+++ b/classes/Models/Module/Marketplace/ModuleUpgradeCompatibility.php
@@ -1,0 +1,69 @@
+<?php
+
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License version 3.0
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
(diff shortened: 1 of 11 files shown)


classes/Models/Module/Marketplace/ModuleUpgradeCompatibility.php
```diff
@@ -0,0 +1,69 @@
...
+    private $compatibleRelease;
+
+    /** @var Release */
+    private $latestRelease; // ou un objet Version
+
+    public function __construct(
+        bool $isCompatible,
+        bool $hasUpdateAvailable,
+        ?Release $compatibleRelease,
+        Release $latestRelease
```
Quetzacoalt91: Shall we move the optional argument at the end of the list?

classes/Services/MarketplaceService.php
```diff
@@ -59,4 +62,50 @@ public function getModuleDetail(string $module)
 
         return Module::fromArray($data);
     }
+
+    /**
+     * @description allows you to get compatibility information for a module based on the target version of PrestaShop.
```
Quetzacoalt91:
```suggestion
     * Allows you to get compatibility information for a module based on the target version of PrestaShop.
```

If you do that, does PHP-CS-Fixer tries to change the PHPdoc as a simple doc?

classes/Task/Update/UninstallModules.php
```diff
@@ -0,0 +1,180 @@
...
+                    continue;
+                }
+            }
+
+            $totalModuleToUninstall = count($moduleToUninstallList);
+
+            $this->container->getFileStorage()->save(
+                [
+                    'backlog' => $moduleToUninstallList,
+                    'initialTotal' => $totalModuleToUninstall,
```
Quetzacoalt91: Can we write these variables in plural form?


```suggestion
                    'backlog' => $modulesToUninstallList,
                    'initialTotal' => $totalModulesToUninstall,
```

classes/Models/Module/Marketplace/ModuleUpgradeCompatibility.php
```diff
@@ -0,0 +1,69 @@
...
+    private $isCompatible;
+
+    /** @var bool */
+    private $hasUpdateAvailable;
+
+    /** @var Release|null */
+    private $compatibleRelease;
+
+    /** @var Release */
+    private $latestRelease; // ou un objet Version
```
Hlavtox: 🥖

ga-devfront: thank you my bad, let my baguette comment ><

classes/Task/Update/UninstallModules.php
```diff
@@ -0,0 +1,180 @@
...
+                UpgradeFileNames::MODULES_TO_UNINSTALL_LIST
+            );
+        } catch (ProcessException $e) {
+            $this->handleException($e);
+
+            return ExitCode::FAIL;
+        }
+
+        if ($totalModuleToUninstall) {
+            $this->logger->info($this->translator->trans('%s installed modules will be uninstalled.', [$totalModuleToUninstall]));
```
Quetzacoalt91:
```suggestion
            $this->logger->info($this->translator->trans('%nbOfModules% incompatible modules with PrestaShop %version% will be uninstalled.', ['%nbOfModules%' => $totalModuleToUninstall, '%version%' => $targetVersion]));
```

Hlavtox:
Guys, what does it do exactly? Does it concern only marketplace modules, or all distribution api? Or all that dont have compatibility stated in the php file? Thanks!

ga-devfront:
@Hlavtox 

> Guys, what does it do exactly? Does it concern only marketplace modules, or all distribution api? Or all that dont have compatibility stated in the php file? Thanks!

This only applies to modules available on the marketplace. There are three possible scenarios:
- The module is found and has a version compatible with the target version: we keep it.
- The module is found but does not have a version compatible with the target version: we uninstall it.
- The module is not found: we keep it. 

We cannot rely on the source file (PHP) because everyone is using `_PS_VERSION_` as the maximum version, which is incorrect. We will implement this feature when the situation has changed.

I would add that modules that have an update downloaded during the previous step (regardless of its source marketplace/local/composer/distribution API) are excluded from this step because we already know that they will be compatible.

upgrade/sql/8.0.1.sql
M0rgan01, line 1: why remove this file ?

classes/Task/Update/UninstallModules.php
```diff
@@ -0,0 +1,177 @@
...
+                    );
+
+                    if (!$moduleCompatibility->isCompatible()) {
+                        $modulesToUninstallList[] = $module['name'];
+                    }
+                } catch (MarketplaceApiException $e) {
+                    $this->logger->warning($this->translator->trans('Unable to retrieve module %s information. Ignored.', [$module['name']]));
+                    $this->container->getUpdateState()->setWarningDetected(true);
+
+                    continue;
```
M0rgan01: useless

M0rgan01:
Please, be consistant @ga-devfront

upgrade/sql/8.0.1.sql
ga-devfront: did you check what's inside and the PR ?

Quetzacoalt91: This file has no more effect, because the module is now deleted with the introduction of UninstallModule

ga-devfront:
> Please, be consistant @ga-devfront

Thank you, you are amazing :heart:

Quetzacoalt91 (approved):

upgrade/sql/8.0.1.sql
M0rgan01: @ga-devfront Please take @Quetzacoalt91 as an example, instead of making remarks.

M0rgan01 (approved):

AureRita (approved):
Hi @ga-devfront 

https://github.com/user-attachments/assets/0801dca4-c245-4bb1-a579-fb8048028080

And if a module have some troble to uninstall, we have correctly an information about it as we can see : 

Image

Tested from : 
8.2.1 to 9.0.3
1.7.8.8 to 8.2.4
8.2.4 to 9.0.3 

Thank you

PR #1651, opened 2026-02-11
Add failing step to segment call

ga-devfront:
| Questions         | Answers
| Description?      | Add failing step then process is on error to know to get an idea of ​​which step is the most problematic in case of an error.
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | N/A
| How to test?      | Check internal ticket

### DIFF
--- a/classes/Analytics.php
+++ b/classes/Analytics.php
@@ -103,12 +103,10 @@ class Analytics
             $this->getProperties($propertiesType)
         );
 
-        if (count($extraProperties)) {
-            $dataToSend['properties'] = array_merge(
-                $dataToSend['properties'],
-                $extraProperties
-            );
-        }
+        $dataToSend['properties'] = array_merge(
+            $dataToSend['properties'],
+            $extraProperties
+        );
 
         \Segment::track($dataToSend);
         \Segment::flush();
--- a/classes/Analytics.php
+++ b/classes/Analytics.php
@@ -90,7 +90,7 @@ class Analytics
     /**
      * @param string $event
      * @param self::WITH_*_PROPERTIES $propertiesType
-     * @param array<string, mixed>|null $extraProperties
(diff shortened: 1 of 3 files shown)


classes/Analytics.php
```diff
@@ -90,17 +90,24 @@ public function __construct(
     /**
      * @param string $event
      * @param self::WITH_*_PROPERTIES $propertiesType
+     * @param string|null $failingStep
      */
-    public function track(string $event, $propertiesType = self::WITH_COMMON_PROPERTIES): void
+    public function track(string $event, $propertiesType = self::WITH_COMMON_PROPERTIES, ?string $failingStep = null): void
```
Quetzacoalt91: What do you think about a more common parameter? Something like "extraProperties" ?

classes/Analytics.php
```diff
@@ -90,17 +90,27 @@ public function __construct(
...
         }
 
-        \Segment::track(array_merge(
+        $dataToSend = array_merge(
             ['event' => '[SUE] ' . $event],
             $this->getProperties($propertiesType)
-        ));
+        );
+
+        if (count($extraProperties)) {
```
Quetzacoalt91: You can remove the `if` statement as you always have an array. the performance hit is fine even if you try to merge an empty array

Quetzacoalt91 (approved):

AureRita (approved):
Hi @ga-devfront 

https://github.com/user-attachments/assets/e2d462c4-c66e-4b1f-9b08-ae0f8544d41d

Image

Tested from : 
1.7.8.8 to 8.2.3

Thank you

AureRita (approved):
Hi @ga-devfront 

https://github.com/user-attachments/assets/68890856-be48-459f-b902-cac488b08b60

Tested from : 
1.7.8.8 to 8.2.3

Thank you

PR #1658, opened 2026-02-12
Introduce a module source provider for updates about modules included in release but not in composer.json

Quetzacoalt91:
| Questions         | Answers
| Description?      | Introduce a ModuleSourceProvider that prevents unexpected copy of modules during the extraction of the PrestaShop files.
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Modules included in classic releases (i.e ps_mbo) are not copied anymore during the step UpdateFiles, but taken properly in account during the DownloadModules step if the version included is higher than the marketplace. A custom release of PrestaShop must be built in order to test this PR.

## Files to use for tests

:link: https://drive.google.com/drive/folders/1d3EwknLnkNYhDA8fmxoufuEktOCMKv6Q?usp=sharing
This is the original release of PrestaShop 9.0.3 with a change of the version of ps_mbo: `5.2.2` -> `9.9.9`.

## Expected behavior

* The update of the module must be found in the archive `NOTICE - ModuleDownloader - Module ps_mbo update files (5.1.0 => 9.9.9) have been fetched from /var/www/html/**admin_folder**/autoupgrade/tmp/files/modules/ps_mbo.
`
* No file of the same module is copied in UpdateFiles
* After the update, the module manager must show the proper version, the one from the update we applied with the zip


classes/PrestashopConfiguration.php
```diff
@@ -59,13 +57,9 @@ public function getModuleVersion(): ?string
...
+        $path = _PS_ROOT_DIR_ . '/modules/autoupgrade';
 
-        if ($this->filesystem->exists($path)
-            && $xml_module_version = simplexml_load_file($path)
-        ) {
-            $this->moduleVersion = (string) $xml_module_version->version;
-        }
+        $this->moduleVersion = $this->moduleAdapter->getVersionFromConfigXmlInFolder($path);
 
         return $this->moduleVersion;
```
ga-devfront, line 64:
```suggestion
        return $this->moduleAdapter->getVersionFromConfigXmlInFolder($path);
```

ga-devfront (dismissed):
Thank you for this contribution

classes/PrestashopConfiguration.php
Quetzacoalt91: This is what I originally did, but note we store the result in a class property as a cache.

ga-devfront (dismissed):

Hlavtox:
@Quetzacoalt91 Hi Thomas, does it also fix the issue of all native modules copied from the zip? 🙏

Quetzacoalt91:
> @Quetzacoalt91 Hi Thomas, does it also fix the issue of all native modules copied from the zip? 🙏

It depends on what you call a native module. If you mean the modules bundled with the PrestaShop release, then yes.

This PR fixes the issue where the modules present in the `modules/` folder but not in `composer.json` were copied whatever the state of the current installation of PrestaShop. A module in this situation would be updated twice, the first one potentially bringing a lower version than the one being already installed. 
From now, modules are copied only in this case: They are part of the composer.json and they have not been manually updated or reinstalled by the merchant (checked with the absence of a `vendor/` folder in the module).

Hlavtox:
@Quetzacoalt91 Let's say that we speak of native modules that are in composer.json now.

I think that the expected behavior should be to `copy all modules currently present on the disk`.

So that:
- All installed modules are updated.
- All uninstalled modules are updated, but kept uninstalled.
- No extra modules are copied on the disk.

Because right now, I use let's say 10 native modules on my shops. After upgrade, I have 70 modules on disk, unused.

Quetzacoalt91:
>     * All installed modules are updated.
>     * No extra modules are copied on the disk.

These are expected changes with this PR

>     * All uninstalled modules are updated, but kept uninstalled.

I need to double check this one. I would expect the modules inside the `composer.json` (= native modules) to be copied even if they've been uninstalled + removed, because the module checks for each module in composer.json that they don't have a `vendor/` folder to decide if they must be copied with PrestaShop.

For instance, if you delete Faceted Search from your module before an update, it would be back.

AureRita:
Hi @Quetzacoalt91 

Thank you for your PR, currently I have some issue when I want to upgrade with ps_mbo as you can see : 


https://github.com/user-attachments/assets/5b4d72e9-b550-4721-afaf-b1ec88c31a18

Waiting for feedbacks

Quetzacoalt91:
After checking, it appears the module is unzipped in `modules/` while I am expecting it in `modules/ps_mbo/` in this case. At the moment, I don't know if this is a regression or an old issue we finally discover.

From the Symfony documentation (https://symfony.com/doc/current/components/filesystem.html#mirror): [mirror()](https://github.com/symfony/symfony/blob/8.0/src/Symfony/Component/Filesystem/Filesystem.php#:~:text=function%20mirror) copies **all the contents** of the source directory into the target one (use the [copy()](https://github.com/symfony/symfony/blob/8.0/src/Symfony/Component/Filesystem/Filesystem.php#:~:text=function%20copy) method to copy single files)

I'll check next week.

ga-devfront (approved):

AureRita (approved):
Hi @Quetzacoalt91 

On Classic : 
Image

Image


On OS : 
Image


Module after update : 
Image

Tested from : 
8.2.1 OS to 9.0.3 
8.2.1 classic to 9.0.3

Thank you

PR #1664, opened 2026-02-17
Fix task uninstall module not catch error properly

ga-devfront:
| Questions         | Answers
| Description?      |Fix issue with uninstall module
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | CI is :green_circle:

### DIFF
--- a/classes/Services/MarketplaceService.php
+++ b/classes/Services/MarketplaceService.php
@@ -42,11 +42,9 @@ class MarketplaceService
     }
 
     /**
-     * @return Module
-     *
      * @throws MarketplaceApiException
      */
-    public function getModuleDetail(string $module)
+    public function getModuleDetail(string $module): Module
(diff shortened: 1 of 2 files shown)


Quetzacoalt91 (dismissed):

classes/Services/MarketplaceService.php
Quetzacoalt91:
```suggestion
    * @throws MarketplaceApiException
    */
public function getModuleDetail(string $module): Module
```

Quetzacoalt91 (approved):

AureRita (approved):
Hi @ga-devfront 

Image

Tested from : 
8.2.1 classic to 9.0.3
8.2.1 os to 9.0.3

Thank you

PR #1689, opened 2026-03-03
Init Symfony kernel during modules uninstallation

Quetzacoalt91:
| Questions         | Answers
| Description?      | Some module try to load Symfony services while they're being uninstalled. This PR loads Symfony and the necessary components of the Core
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | /
| How to test?      | Update from 9.0.3 to PrestaShop 9.2.0, all modules will be uninstalled. Some of them will trigger an issue.

### DIFF
--- a/classes/Task/Update/UninstallModules.php
+++ b/classes/Task/Update/UninstallModules.php
@@ -99,6 +99,8 @@ class UninstallModules extends AbstractTask
     public function init(): void
     {
         $this->container->initPrestaShopCore();
+        // Container may be needed on uninstallation if the module loads services.
+        $this->container->getSymfonyAdapter()->initKernel();
     }
 
     /**
--- a/classes/UpgradeTools/SymfonyAdapter.php
+++ b/classes/UpgradeTools/SymfonyAdapter.php
@@ -59,6 +59,8 @@ class SymfonyAdapter
             }
 
             $kernel->boot();
+            // Starting from PrestaShop 9, some parts of the new context are defined by event listeners on kernel.request.
+            // We may have to trigger it with dummy data in the future.
         }
 
         return $kernel;


classes/UpgradeTools/SymfonyAdapter.php
```diff
@@ -59,6 +59,9 @@ public function initKernel()
             }
 
             $kernel->boot();
+            // Starting from PrestaShop 9, some parts of the new context are defined by event listeners on kernel.request.
+            // We trigger it with dummy data.
+            $kernel->handle(new Request(), HttpKernelInterface::MAIN_REQUEST, true);
```
Quetzacoalt91: Fixes

Image

classes/Task/Update/UninstallModules.php
```diff
@@ -99,6 +99,8 @@ public function run(): int
     public function init(): void
     {
         $this->container->initPrestaShopCore();
+        // Container may be needed on uninstallation if the module loads services.
+        $this->container->getSymfonyAdapter()->initKernel();
```
Quetzacoalt91, line 103: Fixes

Image

Fixes

Image

classes/Task/Update/UninstallModules.php
```diff
@@ -99,6 +99,8 @@ public function run(): int
     public function init(): void
     {
         $this->container->initPrestaShopCore();
+        // Container may be needed on uninstallation if the module loads services.
```
M0rgan01, line 102: Be careful to verify that modules with their own container load it correctly. Normally this is done in the constructor, as with ps_shipping, but it's worth checking.

classes/UpgradeTools/SymfonyAdapter.php
```diff
@@ -59,6 +60,9 @@ public function initKernel()
             }
 
             $kernel->boot();
+            // Starting from PrestaShop 9, some parts of the new context are defined by event listeners on kernel.request.
+            // We trigger it with dummy data.
+            $kernel->handle(new Request());
```
M0rgan01: https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Context/LegacyControllerContextBuilder.php#L71

I don't believe that an empty request has a significant impact.

Quetzacoalt91: Good, that is my objective.

I'm expecting this class to be called https://github.com/PrestaShop/PrestaShop/blob/aaaeabd4710313cfafcb5f0423407d63abeec141/src/PrestaShopBundle/EventListener/Admin/Context/LanguageContextSubscriber.php#L57 so the context is set. The only way is to trigger `KernelEvents::REQUEST`

classes/Task/Update/UninstallModules.php
Quetzacoalt91: I had a run with an update to PS 9.2.0, which does not exist and declares all the modules as "to be uninstalled". I had many issues but their own container is not one of them.

classes/UpgradeTools/SymfonyAdapter.php
```diff
@@ -22,6 +22,7 @@
 namespace PrestaShop\Module\AutoUpgrade\UpgradeTools;
 
 use ReflectionClass;
+use Symfony\Component\HttpFoundation\Request;
```
ga-devfront: please remove it

ga-devfront (changes requested):
please be consistent

ga-devfront (approved):

AureRita (approved):
Hi @Quetzacoalt91 

Image

Tested from : 
9.0.3 classic to 9.2.0 

Thank you

PR #1705, opened 2026-03-09
Show technical error details before generic error message

kseghair:
| Questions         | Answers
| Description?      | Show technical error details before generic error message to make logs easier to read (see images for before/after)
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | /
| Sponsor company   | @prestashopCorp
| How to test?      | Update from 1.7.8.9 to 8.2.4, PHP version 7.2.34
update-warning-error-before

update-warning-error-after

### DIFF
--- a/classes/Task/AbstractTask.php
+++ b/classes/Task/AbstractTask.php
@@ -165,6 +165,10 @@ abstract class AbstractTask
 
     protected function handleException(ProcessException $e): void
     {
+        foreach ($e->getQuickInfos() as $log) {
+            $this->logger->warning($log);
+        }
+
         if ($e->getSeverity() === ProcessException::SEVERITY_ERROR) {
             $this->next = TaskName::TASK_ERROR;
             $this->setErrorFlag();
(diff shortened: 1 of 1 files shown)


Quetzacoalt91 (approved):

Touxten (approved):

classes/Task/AbstractTask.php
```diff
@@ -174,10 +178,6 @@ protected function handleException(ProcessException $e): void
             $this->logger->warning($e->getMessage());
             $this->container->getUpdateState()->setWarningDetected(true);
         }
```
ga-devfront, line 180: What about this part ? If you move the next quick info on the top to get warning first we have to move this part too no ? What do you think ?

AureRita (approved):
Hi @kseghair 

Image

Image

Tested from : 
8.2.1 classic to 9.2.0
1.7.8.9 php 7.2 to 8.2.4

Thank you

PR #1711, opened 2026-03-13
Update header stamps

ga-devfront:
| Questions         | Answers
| Description?      | Update header stamps package, and use new header like Core.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| How to test?      | Ci is :green_circle:

### DIFF
--- a/tests/composer.json
+++ b/tests/composer.json
@@ -1,5 +1,5 @@
 {
-    "license": "AFL-3.0",
+    "license": "OSL-3.0",
     "authors": [
         {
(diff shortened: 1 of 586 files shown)


tests/composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "license": "AFL-3.0",
+    "license": "OSL-3.0",
```
Quetzacoalt91: Keep the old value

Quetzacoalt91 (approved):

PR #1735, opened 2026-03-20
Improve performance of getModulesRequiringAttention with parallel HTTP calls

kseghair:
| Questions         | Answers
| Description?      | When running a PS with lots of modules installed, the modules requiring attention modal takes a lot of time to load. I Improved performance by using parallel HTTP calls. Applied also in the Uninstall Module step
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | /
| Sponsor company   | @prestashopCorp
| How to test?      | Upgrade from 8.2.3 to 8.2.4. The modal should should load correctly and faster. Modules not compatible with target version should be correctly uninstalled

### DIFF
--- a/classes/Services/MarketplaceService.php
+++ b/classes/Services/MarketplaceService.php
@@ -31,6 +31,9 @@ class MarketplaceService
     /** @var Translator */
     private $translator;
 
+    /** @var array<string, Module|MarketplaceApiException> */
+    private array $cache = [];
+
     const ADDONS_API_URL = 'https://api.addons.prestashop.com';
 
     /**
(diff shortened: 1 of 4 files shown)


Quetzacoalt91:
We had another pull-request attempting to plug more HTTP calls to cURL: https://github.com/PrestaShop/autoupgrade/pull/1608

I'd like to reuse the idea of having a specific class handling the downloads, and fallback on another method if curl does not exist for some reasons (or make cURL mandatory)

kseghair:
I'm closing my PR as performances have greatly improved with latest version of the module

Issue #1670, opened 2026-02-24
Undefined array key "addScript" in AjaxResponseBuilder


hackerconsole:
### Prerequisites

- [x] I understand and accept the project's [code of conduct](https://github.com/PrestaShop/PrestaShop/blob/develop/CODE_OF_CONDUCT.md).
- [x] I have already [searched in existing issues](./issues?q=is%3Aissue+label%3ABug) and found no previous report of this bug.

### Describe the bug and add attachments

On a clean installation of PrestaShop 9.1.0, after installing the latest version of the AutoUpgrade (1-Click Upgrade) module from the official GitHub repository, accessing the upgrade dashboard in the Back Office triggers a PHP warning that results in a Symfony exception (HTTP 500).

### Steps to reproduce

1. Install PrestaShop 9.1 clean
2. Install AutoUpgrade module latest from GitHub
3. Go to Backoffice → Dashboard → 1-Click Upgrade
4. Observe the error

### Expected behavior

The AutoUpgrade dashboard should load correctly without any PHP warnings or exceptions.

### Actual Result

Symfony Exception / HTTP 500 Internal Server Error with warning:

Warning: Undefined array key "addScript"
in modules/autoupgrade/classes/AjaxResponseBuilder.php (line 43)

<img width="1140" height="912" alt="Image" src="https://github.com/user-attachments/assets/81478ca4-19cc-46b9-9a46-fd9a7f50cc99" />

### PrestaShop versions where the bug happens

9.0.1

### Module version where the bug happens

7.5.1

### PHP version(s) where the bug happened

8.4

AureRita:
Hi @hackerconsole  ,

Thank you for your report, I tried to reproduce it with php 8.1 as you can see : 

https://github.com/user-attachments/assets/3db400ca-92e3-418f-9ec4-ed2c64ac6f25

Currently, Autoupgrade don't support PHP 8.4, try to launch it with php <8.4. 

Can you try and comeback to say if it's works ? 

Waiting for feedback

hackerconsole:
I can confirm this issue happens on PHP 8.1.34 with a clean installation of PrestaShop 9.1.0 and the latest AutoUpgrade module.


Could you please confirm if this is a known issue or if any additional PHP extension is required for this module?

Currently installed common extensions include:
- php-mbstring
- php-xml
- php-curl
- php-intl
- php-zip
- php-gd
- php-mysqlnd
- php-opcache

hackerconsole:
[...]

PR #1767, opened 2026-04-02
fix: avoid undefined array key warnings in AjaxResponseBuilder

Quetzacoalt91:
| Questions         | Answers
| Description?      | Use method `empty()` to check an array key exists to avoid warnings on strict PHP environments.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/autoupgrade/issues/1670
| How to test?      | Make your PHP configuration strict with error reporting level as E_ALL and run the module on PrestaShop v9.1

### DIFF
--- a/classes/AjaxResponseBuilder.php
+++ b/classes/AjaxResponseBuilder.php
@@ -36,11 +36,11 @@ class AjaxResponseBuilder
             'new_content' => $newContent,
         ];
 
-        if ($options['newRoute']) {
+        if (!empty($options['newRoute'])) {
             $arrayToReturn['new_route'] = $options['newRoute'];
         }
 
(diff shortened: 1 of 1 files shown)


ga-devfront (approved):

ingridusta (approved):
Hi @Quetzacoalt91, 

No matter how much i tried to reproduce the initial bug, I couldn't. So instead of testing the fact that the bug is not reproduced anymore, I did NR on UI. 

No reg on detected on : 
- modal to update
- backup
- upgrade
- restore. 

So, I can say it's QA approved ✅

Issue #1773, opened 2026-04-03
AutoUpgrade module remove modules without my consent - tested from 9.0.3 to 9.1.0


Maofree:
### Prerequisites

- [x] I understand and accept the project's [code of conduct](https://github.com/PrestaShop/PrestaShop/blob/develop/CODE_OF_CONDUCT.md).
- [x] I have already [searched in existing issues](./issues?q=is%3Aissue+label%3ABug) and found no previous report of this bug.

### Describe the bug and add attachments

Hi @Quetzacoalt91 
I'm testing how to upgrade other sites to 9, as I've almost finished redoing all my modules. Now I see that the 1-Click module is uninstalling modules from sites because it deems them incompatible with PrestaShop 9.1. Even if the basic versions don't have this, I can fix them myself and get them to work. However, I'd like to be able to decide certain actions; an option needs to be added to allow users to choose whether or not to uninstall incompatible modules, as perhaps existed in the past. I tried upgrading that site again from 9.0.3 to 9.1.0, and it removed two without warning.

removed the menu and the cookiebanner, but they are basic versions that I modified myself and will stop using them once I have made my own versions. However, I would like to be the one to have control over what happens

### Steps to reproduce

Update from 9.0.3 to 9.1.0 with outdated modules purchased on prestashop addons in the past but fixed

### Actual Result

I would like a choice to be added to decide this type of behavior and to show the list of no longer compatible modules, I would like it to be my choice

### PrestaShop versions where the bug happens

9.1.0

### Module version where the bug happens

7.6.1

### PHP version(s) where the bug happened

8.4

Maofree:
I just upgraded a site from version 8.2.0 to 8.2.5 with the 1-click upgrade 7.6.1 and had no problems. Then I was able to disable the option that would have disabled the modules; none of the modules were removed.
Why can't a button like this be used for PrestaShop 9 to enable or disable the removal of incompatible modules?

<img width="1389" height="838" alt="Image" src="https://github.com/user-attachments/assets/b24690dd-8fbf-4f84-88f5-3bec4bca4d16" />

SerenaButler:
I had the same problem with 3 Shops upgrading from 8.2.3 to 8.2.5 uninstalling modules and removing modules settings from the database. I now will upgrade manually, wish there where upgrade packs.

Maofree:
Hi @Quetzacoalt91 
I'll list two errors I encountered while updating a very large site to help you improve this module.

[...]

PR #1794, opened 2026-04-24
Check module compatibility with both source and destination versions of PrestaShop

Quetzacoalt91:
| Questions         | Answers
| Description?      | When the Marketplace API returns a module with its product page offline and no compatible release exists for the target PS version, the module was always reported as **Incompatible**. This was inaccurate, because the module probably never received an actual zip, has been disabled for all versions of PrestaShop etc. We can't distinguish the `diabled` / `hidden` cases  when is_active: false and no compatible release is found for the target version. So from now: <ul><li>The module was compatible with the source version → Incompatible (it worked before, won't work after. So a genuine breaking change to flag)</li><li>The module was not compatible with  the source version  → Uncertain (it was already incompatible before the upgrade, we cannot say more)</li></ul>
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/autoupgrade/issues/1773
| How to test?      | The test cases can be added in `tests/integration/MarketplaceAndModuleCompatibilityTest.php`, where calls to the marketplace are actually made. Third-party modules that are disabled on the marketplace should get the uncertain state, unless they had a version compatible with the destination version the user updates to.

### DIFF
--- a/classes/Task/Update/UninstallModules.php
+++ b/classes/Task/Update/UninstallModules.php
@@ -137,7 +137,8 @@ class UninstallModules extends AbstractTask
 
             $checkResult = $this->container->getModuleCompatibilityChecker()->getModulesRequiringAttention(
                 $modulesList,
-                $targetVersion
+                $targetVersion,
+                $this->container->getUpdateState()->getCurrentVersion()
             );
 
             foreach ($checkResult['uncertain_modules'] as $moduleName) {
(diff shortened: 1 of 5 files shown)


ga-devfront:
Please run UI test

ga-devfront (approved):

ga-devfront (approved):

Quetzacoalt91:
To be tested on the beta

PR #1818, opened 2026-05-27
Fix feature_value position backfill during 9.0.0 upgrade

### LINKED ISSUE #41554: BIG BUG when upgrading to PrestaShop 9.x from previous versions, related to feature value positions.
### Prerequisites

- [x] I have already [searched in existing issues](https://github.com/PrestaShop/PrestaShop/issues?q=is%3Aissue%20type%3ABug) and found no previous report of this bug.

Hi
I upgraded 7 sites to PrestaShop 9: 5 from 8 with the 1-click module, 1 from 1.6, and 1 from 1.7 with a clean PrestaShop 9 data migration module.
All of these sites have a problem with the feature value positions; they're all set to 1, and it's not possible to change their positions.

I would like to update other sites to 9, but with these errors I have to stop, first I have to figure out how to solve this problem by putting the correct positions back in the database.

This is a critical issue, it should be a high priority, I believe all sites updated to 9 will have this problem

Image

looking in the database in the ps_feature_value table the values ​​all have position 0

Image

### Steps to reproduce

Take a site with prestashop 8 and upgrade it with 1-click to 9 or using a data migration module from a site with 8 to a clean 9

### Actual Result

_No response_

### PrestaShop version where the bug happens

9.1.3

### How have you installed PrestaShop

zip file

8.5

_No response_

Codencode:
| Questions         | Answers
| Description?      | This PR fixes the PrestaShop 9.0.0 upgrade path for shops that already have feature values.<br/><br/>When the new `position` column is added to `feature_value`, existing rows all receive the default value `0`, which can cause ordering issues. This change adds a dedicated upgrade step that initializes sequential positions per feature during the `9.0.0` migration, while avoiding overwriting already-initialized data.| Type?             | bug fix
| BC breaks?        | yes
| Deprecations?     | yes
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/41554
| Sponsor company   | Codencode snc
| How to test?      | 1. Install PS version 8.x.x <br/>2. Upgrade to PrestaShop 9.x.x <br/>3. Check that the position field in the feature_value table is correctly populated and try changing the position of the values in the admin panel.

### NOTE
The positions could also have been set using this query:
```sql
UPDATE ps_feature_value fv
INNER JOIN (
    SELECT
        id_feature_value,
        ROW_NUMBER() OVER (
            PARTITION BY id_feature
            ORDER BY id_feature_value ASC
        ) - 1 AS new_position
    FROM ps_feature_value
) x ON x.id_feature_value = fv.id_feature_value
SET fv.position = x.new_position;
```

However, I preferred to create a PHP file for compatibility reasons with different MySQL / MariaDB versions.

### DIFF
--- a/upgrade/php/ps_900_init_feature_value_positions.php
+++ b/upgrade/php/ps_900_init_feature_value_positions.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * For the full copyright and license information, please view the
+ * LICENSE.md file that was distributed with this source code.
+ */
+
+use PrestaShop\Module\AutoUpgrade\Database\DbWrapper;
+
+/**
+ * Initialize feature value positions after the position column is added in 9.0.0.
+ *
+ * Existing shops upgraded to 9 may already contain feature values, which all get
+ * the default position 0 when the column is introduced. We backfill sequential
+ * positions per feature to avoid duplicates and preserve a stable ordering.
+ *
+ * @return void
+ *
+ * @throws \PrestaShop\Module\AutoUpgrade\Exceptions\UpdateDatabaseException
+ */
+function ps_900_init_feature_value_positions()
+{
+    $hasInitializedPositions = (int) DbWrapper::getValue('
+        SELECT COUNT(*)
+        FROM `' . _DB_PREFIX_ . 'feature_value`
+        WHERE `position` <> 0
+    ');
+
+    if ($hasInitializedPositions > 0) {
+        return;
+    }
+
+    $featureValues = DbWrapper::executeS('
+        SELECT `id_feature_value`, `id_feature`
+        FROM `' . _DB_PREFIX_ . 'feature_value`
+        ORDER BY `id_feature` ASC, `id_feature_value` ASC
+    ');
+
+    if (empty($featureValues)) {
+        return;
+    }
+
+    $currentFeatureId = null;
+    $position = -1;
+
+    foreach ($featureValues as $featureValue) {
+        $featureId = (int) $featureValue['id_feature'];
+
+        if ($featureId !== $currentFeatureId) {
+            $currentFeatureId = $featureId;
+            $position = 0;
+        } else {
+            ++$position;
+        }
+
+        DbWrapper::update(
+            'feature_value',
+            ['position' => $position],
+            '`id_feature_value` = ' . (int) $featureValue['id_feature_value']
+        );
+    }
+}
--- a/upgrade/sql/9.0.0.sql
+++ b/upgrade/sql/9.0.0.sql
@@ -422,6 +422,7 @@ DELETE FROM `PREFIX_hook_module_exceptions` WHERE `id_hook` NOT IN (SELECT id_ho
 /* Feature value position */
 /* https://github.com/PrestaShop/PrestaShop/pull/37042 */
 /* PHP:add_column('feature_value', 'position', 'int(10) unsigned NOT NULL DEFAULT \'0\''); */;
+/* PHP:ps_900_init_feature_value_positions(); */;
 /* PHP:add_configuration_if_not_exists('PS_FEATURE_VALUES_ORDER', 'name'); */;
 
 /* Upgrade attachment names length */


upgrade/php/ps_900_init_feature_value_positions.php
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @return void
+ *
+ * @throws \PrestaShop\Module\AutoUpgrade\Exceptions\UpdateDatabaseException
+ */
+function ps_900_init_feature_value_positions()
+{
+    if (empty(DbWrapper::executeS('SHOW TABLES LIKE "' . _DB_PREFIX_ . 'feature_value"'))) {
+        return;
+    }
```
kpodemski: It seems not relevant to check for existence of this table, what do you think?

upgrade/php/ps_900_init_feature_value_positions.php
```diff
@@ -0,0 +1,69 @@
...
+
+    $hasInitializedPositions = (int) DbWrapper::getValue('
+        SELECT COUNT(*)
+        FROM `' . _DB_PREFIX_ . 'feature_value`
+        WHERE `position` <> 0
+    ');
+
+    if ($hasInitializedPositions > 0) {
+        return;
+    }
```
kpodemski, line 30: This probably is not needed, we know the positions aren't set, right?

upgrade/php/ps_900_init_feature_value_positions.php
Codencode: Yes, you're right, I'll correct it.

upgrade/php/ps_900_init_feature_value_positions.php
Codencode: @kpodemski 
And if for some reason the column had already been added and has the correct positions?
In the upgrade, the `add_column()` function checks whether the column already exists, so it could happen that the column already exists and is populated.

upgrade/php/ps_900_init_feature_value_positions.php
Codencode: @kpodemski
I removed the checks on both the table and the field because we are sure they exist.

Codencode:
@Quetzacoalt91 @kpodemski 
Some files have the header/license in the old format, for example:
https://github.com/PrestaShop/autoupgrade/blob/dev/tests/integration/MarketplaceAndModuleCompatibilityTest.php

Should I fix them in this PR?

Also, I’m not sure whether targeting the `dev `branch was the right choice. Maybe I should have targeted the `7.6.x` branch?

Also, there are errors (example: https://github.com/PrestaShop/autoupgrade/actions/runs/26496496929/job/78025905639?pr=1818) that do not depend on my change, perhaps precisely because I need to change branch?

upgrade/php/ps_900_init_feature_value_positions.php
Hlavtox: Good point - backported feature or failed upgrade mid-point.

Quetzacoalt91:
Yes, can you please target the `7.6.x` branch so your change can reach faster a stable release?

I've seen your how-to-test field, and can you tell me if that's solving an issue while using the shop?

Codencode:
>Yes, can you please target the 7.6.x branch so your change can reach faster a stable release?

@Quetzacoalt91, I performed the rebase, but I ran into some difficulties, so I had to do a reset and re-add the changes. That's why there's now only one commit, unlike before when there were two, but the final result is the same (the first commit, which was modified by the second, is missing).

> I've seen your how-to-test field, and can you tell me if that's solving an issue while using the shop?

Yes, you can see the issue that explains in detail the problem visible in the admin: https://github.com/PrestaShop/PrestaShop/issues/41554

Quetzacoalt91:
I've the updated the base branch so the CI does not fail for a missing library `canvas`. Can you please rebase?

Codencode:
@Quetzacoalt91 done

Quetzacoalt91 (approved):

ingridusta (approved):
Hi @Codencode, 

Thanks for the catch ! 

I reproduced the bug : 

Image

Image

<br>And with ur fix : 

Image

Image

Thanks ! It's QA approved ✅

PR #1819, opened 2026-05-27
Create and drop indexes, keys and foreign keys only when applicable

### LINKED ISSUE #1636: Upgrade from PS 8.1 to 9.0 fails: SQL error when dropping non‑existent index/column
### Prerequisites

When upgrading from PrestaShop 8.x to 9.0 using the autoupgrade module, the upgrade process fails during the database migration step.
The failure occurs because the migration attempts to DROP an index or column (image_type_name) that does not exist in the target database.

This results in a fatal SQL error that stops the upgrade.

**Error message**
`SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP 'image_type_name'; check that column/key exists`

**The error is thrown from:**
```
/classes/db/DbPDO.php line 149
/modules/autoupgrade/.../CoreUpgrader.php
```

**Context**
The upgrade log shows that the process reaches the migration file for version 9.0.0, and then fails when executing a query that attempts to drop image_type_name from ps_image_type.
However, in many real‑world installations, this index/column may have already been removed.

### Expected behavior
Prestashop upgrade module should
Check whether the index or column exists before attempting to drop it 
or use DROP INDEX IF EXISTS / DROP COLUMN IF EXISTS (MySQL 8+)
or wrap the operation in a conditional query using information_schema

### Steps to reproduce
From a PrestaShop 8.x installation where the table ps_image_type does not contain the index/column image_type_name.
Launch the upgrade to PrestaShop 9.0 using the autoupgrade module.
The upgrade stops during the SQL migration step with error 1091.

### PrestaShop version
8.1

Hlavtox:
| Questions         | Answers
| Description?      | Wraps several queries into conditional scripts to further eliminate any potential failures during execution of these queries.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes https://github.com/PrestaShop/autoupgrade/issues/1636
| How to test?      | Auto test should pass, it should verify if all the scripts work fine.

### DIFF
--- a/upgrade/php/add_index_if_not_exists.php
+++ b/upgrade/php/add_index_if_not_exists.php
@@ -27,7 +27,7 @@ use PrestaShop\Module\AutoUpgrade\Database\DbWrapper;
  *
  * @throws \PrestaShop\Module\AutoUpgrade\Exceptions\UpdateDatabaseException
  */
-function add_index_if_not_exists(string $table, string $index, string $parameters): bool
+function add_index_if_not_exists(string $table, string $index, string $parameters, string $indexType = 'INDEX'): bool
 {
     // Verify if we need to create unique key
     $keys = DbWrapper::executeS(
@@ -38,5 +38,5 @@ function add_index_if_not_exists(string $table, string $index, string $parameter
         return true;
     }
 
-    return DbWrapper::execute('ALTER TABLE `' . _DB_PREFIX_ . pSQL($table) . '` ADD INDEX `' . pSQL($index) . '` ' . pSQL($parameters));
+    return DbWrapper::execute('ALTER TABLE `' . _DB_PREFIX_ . pSQL($table) . '` ADD ' . pSQL($indexType) . ' `' . pSQL($index) . '` ' . pSQL($parameters));
 }
--- a/upgrade/php/drop_index_if_exists.php
+++ b/upgrade/php/drop_index_if_exists.php
@@ -20,7 +20,7 @@
 
 use PrestaShop\Module\AutoUpgrade\Database\DbWrapper;
 
-function drop_index_if_exists(string $table, string $index): bool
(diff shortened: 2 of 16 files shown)


upgrade/php/drop_index_if_exists.php
```diff
@@ -0,0 +1,35 @@
...
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License version 3.0
+ */
+
+use PrestaShop\Module\AutoUpgrade\Database\DbWrapper;
+
+function drop_index_if_exists(string $table, string $index, string $indexType = 'INDEX'): bool
```
Quetzacoalt91, line 23: Can we specify the allowed values to reduce the risk of SQL errors in the future?

```suggestion

/**
 * @param 'KEY'|'UNIQUE KEY'|'INDEX' $indexType
 */
function drop_index_if_exists(string $table, string $index, string $indexType = 'INDEX'): bool
```

upgrade/php/add_index_if_not_exists.php
```diff
@@ -13,9 +27,9 @@
  *
  * @throws \PrestaShop\Module\AutoUpgrade\Exceptions\UpdateDatabaseException
  */
-function add_index_if_not_exists(string $table, string $index, string $parameters): bool
+function add_index_if_not_exists(string $table, string $index, string $parameters, string $indexType = 'INDEX'): bool
```
Quetzacoalt91, line 30: Can you please add the detailed `@param 'KEY'|'UNIQUE KEY'|'INDEX' $indexType` in the phpdoc ?

Issue #1682, opened 2026-02-03
Issue with Upgrade from 9.0.2 to 9.0.3 - Failed to remove directory: Directory not empty


joecoolio66:
### Prerequisites

- [x] I understand and accept the project's [code of conduct](https://github.com/PrestaShop/PrestaShop/blob/develop/CODE_OF_CONDUCT.md).
- [x] I have already [searched in existing issues](https://github.com/PrestaShop/PrestaShop/issues?q=is%3Aissue+label%3ABug) and found no previous report of this bug.

### Describe the bug and add attachments

I reported this issue last September and it was referred to [autoupgrade#1523](https://github.com/PrestaShop/autoupgrade/pull/1523) as a potential fix, but I can confirm that I received this error today upgrading 9.0.2 to 9.0.3. Afterwards we had to apply the SQL on the update manually to the DB (luckily it was only 1 query) and also luckily we keep our site up to date so no modules needed to be updated.

Why is this error still happening? What can we do to avoid it in the future? It's like roulette updating. We updated our backup fine.. this error occurred only in our PROD. The same situation as last time.

### Steps to reproduce

Run the Update Assistant (7.5.1) and try updating 9.0.2 to 9.0.3.  It appears that this error is somewhat random

### PrestaShop version(s) where the bug happened

9.02

### How you installed PrestaShop

Update Assistant

### PHP version(s) where the bug happened

8.4

joecoolio66:
Reference to my other bug report which is now closed:  PrestaShop/PrestaShop#39509

Quetzacoalt91:
This issue appears when ajax requests are sent to the shop when the update is running. Among them, we find the check of notifications to display at the top of the page.

The tab running the update won't send these ajax requests, but if you have other pages opened at the same time, they will still send these ajax requests which can impact the process.

Can you close the other tabs opened on the store while the update is running?

Touxten:
Hello @Quetzacoalt91 
What do you think about adding this information somewhere?
I always have several tabs open in the BO.

Quetzacoalt91:
The best solution would be to have a maintenance mode for the BO as well, but yeah, you got a good idea.

joecoolio66:
I thought this issue was supposed to be addressed in https://github.com/PrestaShop/autoupgrade/pull/1523.  Will this be re-addressed for future releases?

Quetzacoalt91:
[...]

PR #1523, opened 2025-11-11
Block ajax requests to the shop while a process runs

Quetzacoalt91:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Updates are reported to fail randomly. This is caused by the ajax request that check regularly if new notifications should be displayed on the top of the back office. This PR introduces a guard that prevents all ajax requests to reach the shop if they are not related to Update Assistant. This fixes the parallel generation of the cache at a moment we try to empty it.
| Type?             | bug fix
| BC breaks?        | Nope
| Deprecations?     | Nope
| Fixed ticket?     | Fixes https://github.com/PrestaShop/PrestaShop/issues/39509
| Sponsor company   | @PrestaShopCorp
| How to test?      | As reported on the video below, the browser console will now display when the guard is toggled on and off. You can manually send ajax requests by clicking on the bell at the top of the back office page. On the `dev` branch, you can trigger it when you want and this makes the update fail if called at the end of the update database, while on this PR no ajax request will appear on the network tab.

**Note:** If requests are sent to the shop from other sources (API, probably front end as well...), Update Assistant won't be able to block them.

https://github.com/user-attachments/assets/1baae177-be50-4f75-b118-966350045fec

PR #1820, opened 2026-05-27
Catch IOException in CacheCleaner to prevent upgrade abort on locked cache files

rboadalabelgrup:
During an upgrade, if PHP-FPM processes have cache files open, clearDirectory() throws an IOException and the upgrade process aborts completely.

This change wraps the clearDirectory() call in a try/catch so that locked files produce a warning in the log instead of an unhandled exception, allowing the upgrade to continue.

| Questions         | Answers
| Description?      | During an upgrade, if PHP-FPM processes have cache files open, `clearDirectory()` throws an `IOException` and the upgrade process aborts completely. This change wraps the `clearDirectory()` call in a try/catch so that locked files produce a warning in the log instead of an unhandled exception, allowing the upgrade to continue.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | N/A
| Sponsor company   | N/A
| How to test?      | 1. Run a PrestaShop instance with PHP-FPM <br> 2. Start an upgrade while PHP-FPM processes have cache files open <br> 3. **Before:** the upgrade aborts with an unhandled IOException <br> 4. **After:** the upgrade continues and a warning appears in the log instead

## Possible impacts

- Cache clearing behavior during upgrades
- Upgrade log output (a new warning message may appear when cache files are locked)

### DIFF
--- a/classes/UpgradeTools/CacheCleaner.php
+++ b/classes/UpgradeTools/CacheCleaner.php
@@ -24,6 +24,7 @@ namespace PrestaShop\Module\AutoUpgrade\UpgradeTools;
 use Exception;
 use PrestaShop\Module\AutoUpgrade\Log\LoggerInterface;
 use PrestaShop\Module\AutoUpgrade\UpgradeContainer;
+use Symfony\Component\Filesystem\Exception\IOException;
 
 class CacheCleaner
 {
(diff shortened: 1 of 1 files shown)


ps-jarvis:
Hello @rboadalabelgrup!

Quetzacoalt91:
Can you please target the branch `7.6.x` and rebase from there?

About the changes, the changes look good to me but as this is not an actual fix and the cache partially remains, this will require some tests before being approved by the QA.

rboadalabelgrup:
> Can you please target the branch `7.6.x` and rebase from there?
> 
> About the changes, the changes look good to me but as this is not an actual fix and the cache partially remains, this will require some tests before being approved by the QA.

Agreed, this is a best-effort approach. The upgrade continues and the operator is warned via the log. If a stricter behavior is preferred (e.g. retrying or listing the locked files), I'm open to suggestions.

Quetzacoalt91:
We also lack actual solution on that. Our investigations showed that any parallel request to the store while the update is running will generate the cache. If the deletion of the cache happens at the same time, we will encounter the collision.
We have limited the risk with https://github.com/PrestaShop/autoupgrade/pull/1523, but the issue didn't disappear.

rboadalabelgrup:
@Quetzacoalt91 Thanks for the context! I had a look at #1523 — makes sense as a first line of defense for back office AJAX requests. My change would cover the cases that guard can't block, like requests coming from the front end or external APIs. The upgrade wouldn't abort in those situations, the locked files would just remain and the operator gets a warning in the log. Not a perfect solution, but it could complement what's already there. Happy to adjust if you have something else in mind.

Quetzacoalt91:
I've the updated the base branch so the CI does not fail for a missing library `canvas`. Can you please rebase?

rboadalabelgrup:
Done, rebased on 7.6.x.

Quetzacoalt91 (approved):

Quetzacoalt91:
The fix has been reported as working by @sassabi on https://github.com/PrestaShop/autoupgrade/issues/1682#issuecomment-5701741626 :+1:

PR #1868, opened 2026-07-10
Warm up the core cache after the final cache clean on update completion

### LINKED ISSUE #1867: Back office redirects to the front office after updating to 9.0+ (core cache wiped after warm-up)
### Describe the bug

After a CLI update to **PrestaShop 9.0+**, the freshly updated shop's **back office redirects to the front office**: opening `/<admin>/` (e.g. `/admin-dev/`) returns a `302` to `/` → `/<lang>/` instead of the login page. The shop stays in that state (it does not self-heal), until a Symfony cache rebuild is triggered.

In PrestaShop 9.0 the BO login is a Symfony route (`admin_login` → `/login`); when the Symfony `prod` cache is missing, the admin entry point can't boot and falls back to the front office.

### Root cause

During the update, `UpdateDatabase` warms up the core cache (`CoreUpgrader::warmupCoreCache()`, gated by `shouldWarmupCoreCache()` = destination ≥ 9.0.0 && CLI), which builds `var/cache/prod`. But the later `UpdateComplete` task calls `CacheCleaner::cleanFolders()` (which clears `var/cache/`) **after** that warm-up, wiping the just-built prod container. The update finishes with an empty `var/cache/prod` (only a leftover `<version><lang>.xml`), so the 9.0 Symfony back office cannot start.

### How I reproduced it (locally, mirrors the CI `Upgrade CLI` job)

1. `docker compose` PrestaShop `8.1.7-8.1`, then autoupgrade CLI local update to `9.0.0`.
2. Right after the update, `curl -s -o /dev/null -w '%{redirect_url}' http://localhost/admin-dev/` → `http://localhost/` (front office), **30/30 times**.
3. `var/cache/prod/` contains only `9.0.0en.xml` (no container).
4. `php bin/console cache:clear --env=prod` → BO immediately works again (`/admin-dev/` → `/admin-dev/login`).

### Impact

- Merchants updating to 9.0+ can land on a broken back office until the cache is rebuilt.
- CI: this is what makes the `Upgrade CLI` (and Upgrade UI) sanity `should login in BO` step flaky on 9.x destinations — the post-update sanity test lands on the front office and times out waiting for `#email`. Whether it passes depends on whether the first web hit manages to rebuild the container, hence the intermittent ~50% failures on the `dev` nightly and on PRs targeting `7.6.x`.

### Expected behavior

After the update completes, the shop should boot straight into the back office without any manual cache clear.

### Fix

Warm up the core cache again **after** the final `cleanFolders()` in `UpdateComplete` (PR to follow).

### Versions

- autoupgrade: `7.6.x` / `dev`
- Destination: PrestaShop `9.0.0` (also affects 9.x in general)

mattgoud:
| Questions       | Answers                                                    |
| --------------- | ---------------------------------------------------------- |
| Description?    | Warm up the core cache again after the final cache clean at the end of the update, so a shop updated to 9.0+ boots straight into the back office. |
| Type?           | bug fix                                                    |
| BC breaks?      | no                                                         |
| Deprecations?   | no                                                         |
| Fixed ticket?   | Fixes #1867                                                |
| Sponsor company | @PrestaShopCorp                                            |
| How to test?    | Update a shop to 9.0.0 (CLI, local channel), then open the admin URL (e.g. `/admin-dev/`). Before this PR it 302-redirects to the front office (empty `var/cache/prod`); with this PR it lands on the back office login. |

### What happens

During the update, `UpdateDatabase` warms up the core cache (`CoreUpgrader::warmupCoreCache()`, for destination ≥ 9.0.0 on CLI), building `var/cache/prod`. The later `UpdateComplete` task then calls `CacheCleaner::cleanFolders()`, which clears `var/cache/` **after** that warm-up and wipes the just-built prod container. The shop is left with an empty prod cache, and because the PrestaShop 9.0 back office login is a Symfony route, the admin entry point can't boot and falls back to the front office until a manual cache rebuild.

### Fix

- Move `getCoreUpgrader()` (and its cached instance) up from `UpdateDatabase` to the shared `AbstractTask` so it can be reused.
- In `UpdateComplete`, after the final `cleanFolders()`, warm the core cache again (same `shouldWarmupCoreCache()` gate), so the freshly updated shop keeps a valid Symfony `prod` cache.

### Validation

Reproduced locally with the `Upgrade CLI` docker setup (8.1.7 → 9.0.0, local channel):

- Before: `/admin-dev/` redirects to the front office 30/30, `var/cache/prod` only holds `9.0.0en.xml`; the sanity `should login in BO` step times out on `#email`.
- After: `/admin-dev/` → `/admin-dev/login` 8/8, `var/cache/prod` holds the built container; the full `01_ordersBO/01_filterOrders` sanity campaign passes 11/11 with the unmodified test.

This is also the root cause of the intermittent `should login in BO` failures on 9.x destinations in the Upgrade CLI / Upgrade UI jobs.

### DIFF
--- a/classes/Task/AbstractTask.php
+++ b/classes/Task/AbstractTask.php
@@ -28,6 +28,10 @@ use PrestaShop\Module\AutoUpgrade\Exceptions\ProcessException;
 use PrestaShop\Module\AutoUpgrade\Log\Logger;
 use PrestaShop\Module\AutoUpgrade\Task\Runner\ChainedTasks;
 use PrestaShop\Module\AutoUpgrade\UpgradeContainer;
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\CoreUpgrader\CoreUpgrader;
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\CoreUpgrader\CoreUpgrader17;
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\CoreUpgrader\CoreUpgrader80;
+use PrestaShop\Module\AutoUpgrade\UpgradeTools\CoreUpgrader\CoreUpgrader81;
 use PrestaShop\Module\AutoUpgrade\UpgradeTools\Translator;
 use ReflectionClass;
 
@@ -84,6 +88,9 @@ abstract class AbstractTask
      */
     protected $next;
 
+    /** @var CoreUpgrader */
+    private $coreUpgrader;
+
     /**
      * @throws Exception
      */
@@ -100,6 +107,23 @@ abstract class AbstractTask
                 $this->logger->updateLogsPath($logPath);
             }
         }
+    }
+
+    public function getCoreUpgrader(): CoreUpgrader
+    {
+        if ($this->coreUpgrader !== null) {
+            return $this->coreUpgrader;
+        }
+
+        if (version_compare($this->container->getUpdateState()->getDestinationVersion(), '8', '<')) {
+            $this->coreUpgrader = new CoreUpgrader17($this->container, $this->logger);
+        } elseif (version_compare($this->container->getUpdateState()->getDestinationVersion(), '8.1', '<')) {
+            $this->coreUpgrader = new CoreUpgrader80($this->container, $this->logger);
+        } else {
+            $this->coreUpgrader = new CoreUpgrader81($this->container, $this->logger);
+        }
+
+        return $this->coreUpgrader;
     }
 
     /**
--- a/classes/Task/Update/UpdateComplete.php
+++ b/classes/Task/Update/UpdateComplete.php
@@ -94,7 +94,13 @@ class UpdateComplete extends AbstractTask
         // earlier during the database update. On PrestaShop 9.0+ the back office relies on that
         // (Symfony) cache; leaving it empty makes the admin entry point fall back to the front
         // office. Warm it up again so the updated shop boots straight into the back office.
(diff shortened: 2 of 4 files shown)


mattgoud:
Pushed d9403377 to fix the red UI jobs.

Root cause: `UpdateComplete::run()` called `getCoreUpgrader()->shouldWarmupCoreCache()`, and instantiating `CoreUpgrader` runs `Db::getInstance()` in its constructor. In the web (AJAX) completion request the legacy `Db` class isn't autoloaded at that late stage, so it fataled with `Class "Db" not found` -> HTTP 500 -> the UI never reached the success screen and every UI upgrade hung at 'update complete' until the test timed out (1.4h). CLI was fine because `Db` is loaded there, which is why only the UI jobs were red.

Fix: the warmup only ever runs from the CLI path (`shouldWarmupCoreCache()` already requires `php_sapi_name() === 'cli'`), so I short-circuit on the SAPI before touching `getCoreUpgrader()`, meaning the upgrader is never instantiated during the web completion request.

Reproduced locally with the UI docker setup (8.0.3 -> 8.1.0 local channel): before = timeout at 'wait until the end of the update'; after = 26 passed, the step completes in ~50s and the autoupgrade log runs UpdateComplete through cleanly with no CRITICAL.

Quetzacoalt91:
At the moment running an update with the proper user is enough, let's keep the changes simple.

Replaced by #1872

PR #1889, opened 2026-07-29
Migrate combination image associations to carry their shop

boo-code:
Migration for PrestaShop/PrestaShop#42190, which gives `product_attribute_image` the shop its association was made for so that choosing combination images in one shop stops discarding what the other shops chose (PrestaShop/PrestaShop#12638, PrestaShop/PrestaShop#39226).

The new shape mirrors `product_carrier`: `PRIMARY KEY (id_product_attribute, id_image, id_shop)`.

The order matters, because one existing row becomes one row per shop and the old two column key cannot hold them:

1. add the column with a temporary default so the existing rows stay valid,
2. widen the primary key, otherwise the extra rows collide,
3. give every existing association to each shop its image belongs to, read from `image_shop`,
4. drop the placeholder rows and the temporary default.

Existing installs therefore keep exactly the selection they have today: an association becomes visible in the shops that already had the image, and in nothing else.

Verified against a copy of a real 9.2 database with a second shop added and half the images associated to it:

```
before:  39 rows
after:   57 rows   (39 for shop 1, 18 for shop 2)
leftover id_shop = 0 rows: 0
PRIMARY KEY: id_product_attribute, id_image, id_shop
```

The `INSERT ... SELECT` reads through a derived table so MySQL is not selecting from the table it inserts into.

### DIFF
--- a/upgrade/sql/9.2.0.sql
+++ b/upgrade/sql/9.2.0.sql
@@ -232,3 +232,11 @@ UPDATE `PREFIX_feature_flag` SET `stability` = 'stable' WHERE `name` = 'merchand
 
 -- https://github.com/PrestaShop/PrestaShop/pull/42024
 DELETE FROM `PREFIX_feature_flag`  WHERE `name` = 'state';
+
+-- https://github.com/PrestaShop/PrestaShop/pull/42190
+/* Combination image associations become per shop: add the column, widen the primary key, then give every existing association to each shop its image belongs to so current installs keep the selection they have */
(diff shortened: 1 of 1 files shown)


ps-jarvis:
Hello @boo-code!

boo-code:
Putting this on hold: the core change it migrates, PrestaShop/PrestaShop#42190, has been held for v10 - `9.2.x` is under feature freeze and the change is a wide one. Drafting this so it cannot land ahead of the column it migrates to.

The migration itself is unchanged and verified; it will need moving to whichever upgrade file is current when the core side is picked up.

PR #1895, opened 2026-08-27
Add new 9.2.0 hooks to the SQL upgrade script

mattgoud:
## Description

Adds the new 9.2.0 hooks to `upgrade/sql/9.2.0.sql`, as part of the 9.2.0-rc.1 release process.

Generated with:

```
php bin/console prestashop:update:sql-upgrade-file-hooks-listing 9.1.5 <autoupgrade>
```

`9.1.5` is the comparison base: it is the last 9.1.x release, and there is no `9.2.0-beta.1.sql`, so `9.2.0.sql` has to cover the whole 9.1.5 → 9.2.0 delta.

The command reported **45 new hooks** and no obsolete ones, so no `DELETE` statement is needed.

## One manual adjustment

The command appends its block without looking at what the file already contains, so `actionCheckoutBuildProcess` ended up listed twice: once in the generated block, once in the insert added earlier by #41047.

Both inserts have been merged into the single generated block, which now holds all 45 hooks, and a comment records where that one row comes from. The file is left with a single `INSERT INTO PREFIX_hook` statement.

## How to test

- Executing the block against a 9.2.0 database raises no SQL error.
- All 45 hooks already exist with identical `title` and `description` in a freshly installed 9.2.0 database, so the block is a no-op there and stays consistent with the `install-dev` fixtures. On an upgraded shop it inserts the missing rows, and `ON DUPLICATE KEY UPDATE` refreshes titles and descriptions.
- `SELECT name, COUNT(*) FROM ps_hook GROUP BY name HAVING COUNT(*) > 1` returns nothing — `ps_hook.name` is a `UNIQUE KEY` and no hook is listed twice in the file.

### DIFF
--- a/upgrade/sql/9.2.0.sql
+++ b/upgrade/sql/9.2.0.sql
@@ -153,11 +153,6 @@ CREATE TABLE IF NOT EXISTS `PREFIX_b2b_role_authorization_role` (
 
 -- End of B2B feature
 
--- https://github.com/PrestaShop/PrestaShop/pull/41047
-INSERT INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `position`) VALUES
-  (NULL, 'actionCheckoutBuildProcess', 'Build checkout process', 'This hook is triggered before the checkout is rendered. Modules may return a checkout process provider. The provider is used only when exactly one enabled and valid provider is available.', '1')
-ON DUPLICATE KEY UPDATE `title` = VALUES(`title`), `description` = VALUES(`description`);
-
 -- https://github.com/PrestaShop/PrestaShop/pull/41092
 -- Create the extra property definition registry table.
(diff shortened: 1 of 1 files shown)


mattgoud:
Closing: this was already done. `7.6.x` has carried the exact same 45-hook block since fea50bf4 ("Update hooks for 9.2.0 upgrade", 2026-07-20), including the same decision to keep `actionCheckoutBuildProcess` in a single INSERT — the two hook name sets are identical, so this PR adds nothing.

It also targeted the wrong branch: `9.2.0.sql` is maintained on `7.6.x`, not `dev`, which is still missing that block along with #41594, #41777 and #41977.

One genuine gap came out of the exercise and is being opened separately against `7.6.x`: `actionNotFound` is never inserted by any upgrade script.

PR #1900, opened 2026-09-01
Modify wording of unknown modules

Hlavtox:
| Questions         | Answers
| Description?      | The current message suggests that modules whose compatibility cannot be verified may be "homemade" or unlisted from the Marketplace. This can be misleading. A module may be professionally developed and distributed directly by its developer without ever being listed on the PrestaShop Marketplace. In particular, this is common for local payment, shipping and other integrations in some markets. The term "homemade" may also imply lower quality to non-technical merchants, even though PrestaShop only knows that the module's compatibility could not be verified. This PR makes the message neutral and directs merchants to review the affected modules and check their compatibility with their developers.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      |

### DIFF
--- a/views/templates/dialogs/dialog-modules-report-content.html.twig
+++ b/views/templates/dialogs/dialog-modules-report-content.html.twig
@@ -28,7 +28,7 @@
 {% if uncertain_modules is defined and uncertain_modules|length > 0 %}
   {% include "@ModuleAutoUpgrade/components/scrollable-list.html.twig" with {
     'title': 'Uncertain modules'|trans({}),
-    'description': 'The compatibility of the following modules with the destination version of PrestaShop cannot be checked. It could be because they are homemade or have been unlisted from the Marketplace. Please review them via the [1]Module Manager[/1].'|trans({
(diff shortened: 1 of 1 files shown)


Quetzacoalt91:
Hi Daniel, let's go to change the wording, however I'd like to keep a mention of the Marketplace in the text because it helps people to know were we get the data from.

kpodemski (dismissed):
See @Quetzacoalt91 request.

Hlavtox:
@Quetzacoalt91 @kpodemski Updated the wording to mention the PrestaShop Marketplace as the source of compatibility data and direct merchants to confirm compatibility with the module developers.

Quetzacoalt91 (approved):

mattgoud (approved):

Quetzacoalt91:
So far so good for me, at least for the web version. For the CLI we can add the same change there: https://github.com/PrestaShop/autoupgrade/blob/855d436020d62acd88f6337d93e5965a45342bc5/classes/Commands/CheckModulesCommand.php#L198

Quetzacoalt91:
image

:heavy_check_mark: Looks good to me, I'll open another PR for the CLI.
The two failing workflow can't pass at the moment, there is a certificate issue while updating dependencies on Debian Bullseye.

PR #1917, opened 2026-09-17
Update Assistant CLI: add an option to skip the modules step

LaBisquerie:
| Questions         | Answers
| Description?      | Adds a `--skip-modules-step` CLI flag (and `skip_modules_step` config key) to `update:start`, to update only the core. When set, modules are left untouched: the `DownloadModules`, `UninstallModules` and `UpdateModules` steps are skipped, the `/modules` folder is excluded from the files update, and non-native modules are not disabled. Module-related options are then ignored.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | https://github.com/PrestaShop/autoupgrade/discussions/1916
| Sponsor company   | creabilis.com
| How to test?      | 1. Run `php modules/autoupgrade/bin/console update:start <admin-dir> --channel=online_recommended --skip-modules-step`.<br>2. Check the logs: after Unzip, the process goes UpdateFiles → UpdateDatabase → CleanDatabase → UpdateComplete (no DownloadModules, UninstallModules nor UpdateModules).<br>3. Check the store is updated, the `/modules` folder is unchanged, modules keep their enabled state and no module name in `ps_module` starts with `UA_`.<br>4. Run without the flag: modules are handled as before.


classes/UpgradeTools/FileFilter.php
```diff
@@ -145,6 +145,13 @@ public function getFilesToIgnoreOnUpgrade(): array
             '/override/modules',
         ];
 
+        // Leave the whole modules folder untouched, including native modules and new modules shipped with the release
+        if ($this->updateConfiguration->shouldSkipModulesStep()) {
```
Codencode, line 149: I may be missing something here, but is excluding only `/modules` enough to guarantee that the modules directory remains completely untouched?

In `UpdateFiles::warmUp()`, files marked as `deleted` are also added from the version diff. Those paths do not seem to go through the recursive directory filtering, so a path such as `/modules/foo/old-file.php` could still end up in the upgrade list.

Since the goal of `--skip-modules-step` is to leave the whole `/modules` directory unchanged, shouldn't we also explicitly exclude `/modules/*` from the deleted files list?

LaBisquerie: Good point, but deleted files are already filtered. ChecksumCompare::compareReleases(), which builds the deleted files list, calls isFileSkipped($file, $path, 'upgrade', '') on each entry before recursing. So /modules is skipped as a whole and none of its children are added.
I checked it with compareReleases() and a v1 containing /modules/foo/old-file.php and /modules/removed_module, both missing in v2:
with --skip-modules-step: deleted = ['/old.php']
without: deleted = ['/old.php', '/modules/foo/old-file.php', '/modules/removed_module']

Codencode: Thanks for checking it 👍 You're right, I was looking at how the deleted files are consumed in UpdateFiles::warmUp() without following the whole path back to how that list is built.

Since compareReleases() applies isFileSkipped() before recursing into the directory, excluding /modules there also prevents its children from being added to the deleted files list.

The additional test confirms it. Thanks for the clarification!

jf-viguier (approved):

